PR #1, opened 2023-04-04
PoC of Synapse - request for review

merkata:
This is a PR requesting feedback on the Synapse PoC. The code is far from ideal and complete, though please don't spare any comment and feedback, from apparent things like the duplication of code in running processes inside the container to decisions on how to handle config changes to what to use as ingress (nginx vs. traefik vs...).


src/charm.py
```diff
@@ -0,0 +1,323 @@
...
+        return_code = result.return_code
+        logger.debug(
+                  "Run command: %s, return code %s\nstdout: %s\nstderr:%s",
+                  cmd,
+                  return_code,
+                  result.stdout,
+                  result.stderr,
+         )
+
+    def _run_generate_synapse(self):
```
amandahla, line 188: When you generate the config with the env vars set, will change the config file to use something like [this](https://github.com/matrix-org/synapse/blob/develop/docker/conf/homeserver.yaml#L53) instead of sqlite3?

src/charm.py
```diff
@@ -0,0 +1,323 @@
...
+    """Charm the service."""
+
+    _CONTAINER_NAME = "synapse"
+    _SYNAPSE_CONFIG_PATH = "/data/homeserver.yaml"
+
+    _stored = StoredState()
+
+    def __init__(self, *args):
+        super().__init__(*args)
+        self.framework.observe(self.on.install, self._on_install)
```
amandahla, line 45: While observing on.install, I got an error in line `process = self._container().exec(` because the container was not ready yet. After removing it, the deploy was successful. Is there a reason for observing this event?

.gitignore
```diff
@@ -0,0 +1,13 @@
+venv/
```
jdkandersson, line 1: Should we copy the `.gitignore` from, for example, here: https://github.com/canonical/upload-charm-docs/blob/main/.gitignore, it has a bunch of things relating to Python in it already

CONTRIBUTING.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Build the charm
+
+Build the charm in this git repository using:
+
+```shell
+charmcraft pack
+```
+
+<!-- You may want to include any contribution/style guidelines in this document>
```
jdkandersson, line 33: Do we want to keep this comment?

README.md
```diff
@@ -0,0 +1,90 @@
...
+
+Use links instead.
+-->
+
+# matrix-operator
+
+Charmhub package name: matrix-operator
+More information: https://charmhub.io/matrix-operator
+
+Describe your charm in one or two sentences.
```
jdkandersson, line 16: Looks like the default readme?

actions.yaml
```diff
@@ -0,0 +1,20 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+#
+
+register-user:
+  description: Registers a user for the Synapse server.
+  properties:
+    username:
+      description: User name
```
jdkandersson, line 9: Would it be worth it to make these descriptions be a bit more informative? They are essentially re-stating the name at the moment

jdkandersson: also for `password`

config.yaml
```diff
@@ -0,0 +1,30 @@
...
+  log-level:
+    description: |
+      Configures the log level of synapse.
+
+      Acceptable values are: "info", "debug", "warning", "error" and "critical"
+    default: "info"
+    type: string
+  server_name:
+    description: |
+      Synapse server name.
```
jdkandersson, line 14: Similar comment as earlier, should the descriptions in this file also be updated to be a bit more informative? Is this supposed to be the name of the DNS record? Or something else?

homeserver.yaml
```diff
@@ -0,0 +1,37 @@
...
+registration_shared_secret: "n5M1pehEj:fr+i.+4o1P0My7*~V,iKfQ8tawjy*_0RGuxXMNSu"
+report_stats: true
+macaroon_secret_key: "wboT^:0l6YL;0:XtDezWsWp5pY0=3QdBbhY_nsSV2Je0ffNQ;q"
+form_secret: "4rTUH6S3,E6AO2rcb36^s,mu64ROITAb3W8U0,@s2bA:NF2*6L"
+signing_key_path: "/data/my.matrix.host.signing.key"
+trusted_key_servers:
+  - server_name: "matrix.org"
+
+
+# vim:ft=yaml
```
jdkandersson, line 37: Final new line?

homeserver.yaml
```diff
@@ -0,0 +1,37 @@
...
+# This is a YAML file: see [1] for a quick introduction. Note in particular
+# that *indentation is important*: all the elements of a list or dictionary
+# should have the same indentation.
+#
+# [1] https://docs.ansible.com/ansible/latest/reference_appendices/YAMLSyntax.html
+#
+# For more information on how to configure Synapse, including a complete accounting of
+# each option, go to docs/usage/configuration/config_documentation.md or
+# https://matrix-org.github.io/synapse/latest/usage/configuration/config_documentation.html
+server_name: "my.matrix.host"
```
jdkandersson, line 12: How does this interact with the charm configuration option? Should this perhaps be a template in a `files` directory of the charm?

homeserver.yaml
```diff
@@ -0,0 +1,37 @@
...
+pid_file: /data/homeserver.pid
+listeners:
+  - port: 8008
+    tls: false
```
jdkandersson, line 16: Who needs encryption? 😉

homeserver.yaml
```diff
@@ -0,0 +1,37 @@
...
+    x_forwarded: true
+    resources:
+      - names: [client, federation]
+        compress: false
+database:
+  name: sqlite3
+  args:
+    database: /data/homeserver.db
+log_config: "/data/my.matrix.host.log.config"
+media_store_path: /data/media_store
```
jdkandersson, line 27: Similar question to above, how does this interact with the charm configuration? Also applicable to any other cases where this file and the charm config overlaps

homeserver.yaml
```diff
@@ -0,0 +1,37 @@
...
+registration_shared_secret: "n5M1pehEj:fr+i.+4o1P0My7*~V,iKfQ8tawjy*_0RGuxXMNSu"
```
jdkandersson, line 28: Oops? Should this secret be here?

homeserver.yaml
```diff
@@ -0,0 +1,37 @@
...
+report_stats: true
+macaroon_secret_key: "wboT^:0l6YL;0:XtDezWsWp5pY0=3QdBbhY_nsSV2Je0ffNQ;q"
```
jdkandersson, line 30: Oops? Should this secret be here?

homeserver.yaml
```diff
@@ -0,0 +1,37 @@
...
+form_secret: "4rTUH6S3,E6AO2rcb36^s,mu64ROITAb3W8U0,@s2bA:NF2*6L"
```
jdkandersson, line 31: Oops? Should this secret be here?

pyproject.toml
```diff
@@ -0,0 +1,39 @@
...
+minversion = "6.0"
+log_cli_level = "INFO"
+
+# Formatting tools configuration
+[tool.black]
+line-length = 99
+target-version = ["py38"]
+
+# Linting tools configuration
+[tool.ruff]
```
jdkandersson, line 18: Most of our charms still use `flake8` due to certain plugins, I guess this will be reverted to `flake8` in the future?

src/charm.py
```diff
@@ -0,0 +1,323 @@
+#!/usr/bin/env python3
+# Copyright 2023 Mariyan Dimitrov
+# See LICENSE file for licensing details.
+#
+# Learn more at: https://juju.is/docs/sdk
+
+"""Charm the service.
```
jdkandersson, line 7: Is it worth updating this? It seems to be the default description

src/charm.py
```diff
@@ -0,0 +1,323 @@
...
+VALID_LOG_LEVELS = ["info", "debug", "warning", "error", "critical"]
+
+
+class MatrixOperatorCharm(CharmBase):
+    """Charm the service."""
+
+    _CONTAINER_NAME = "synapse"
+    _SYNAPSE_CONFIG_PATH = "/data/homeserver.yaml"
+
+    _stored = StoredState()
```
jdkandersson, line 41: Is this necessary? The recommendation is currently against using this and instead using a file on the application container

src/charm.py
```diff
@@ -0,0 +1,323 @@
...
+        )
+        self._stored.set_default(
+            db_name=None,
+            db_user=None,
+            db_password=None,
+            db_host=None,
+            db_port=None,
+        )
+
+        self.db = pgsql.PostgreSQLClient(self, "db")  # pylint: disable=C0103
```
jdkandersson, line 67: We have guidance on how to disable linting issues, is it applicable here: https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#static-code-analysis?

jdkandersson: Also applies for any other linting rule disables

src/charm.py
```diff
@@ -0,0 +1,323 @@
...
+          )
+        self.framework.observe(
+              self.db.on.master_changed,
+              self._on_master_changed
+          )
+
+    def _on_database_relation_joined(
+         self, event: pgsql.DatabaseRelationJoinedEvent  # type: ignore
+    ) -> None:
+        """Handle db-relation-joined.
```
jdkandersson, line 80: Perhaps describe this in more detail? When is it triggered, what should it do, etc?

src/charm.py
```diff
@@ -0,0 +1,323 @@
...
+
+        Args:
+            event: Event triggering the database relation joined handler.
+        """
+        if self.model.unit.is_leader():
+            # Provide requirements to the PostgreSQL server.
+            event.database = DATABASE_NAME
+            event.extensions = ["hstore:public", "pg_trgm:public"]
```
jdkandersson, line 88: Could this be a tuple instead of a list since it seems constant?

src/charm.py
```diff
@@ -0,0 +1,323 @@
...
+            event.defer()
+
+    # pgsql.MasterChangedEvent is actually defined
+    def _on_master_changed(self, event: pgsql.MasterChangedEvent) -> None:
+        """Handle changes in the primary database unit.
+
+        Args:
+            event: Event triggering the database master changed handler.
+        """
+        if event.master is None:
```
jdkandersson, line 101: The `woke` tool probably won't like this

src/charm.py
```diff
@@ -0,0 +1,323 @@
...
+
+        Returns:
+            The pebble instance of the Synapse container.
+        """
+        return self.unit.get_container(self._CONTAINER_NAME)
+
+    def _populate_synapse_env_settings(self):
+        """Populate env settings to pass around."""
+        if os.getenv("SYNAPSE_SERVER_NAME") is None:
+            os.environ["SYNAPSE_SERVER_NAME"] = self.config["server_name"]
```
jdkandersson, line 139: Is this intending to manipulate the environment of the charm container?

src/charm.py
```diff
@@ -0,0 +1,323 @@
...
+                "POSTGRES_HOST": self._stored.db_host,
+                "POSTGRES_PORT": self._stored.db_port,
+                "POSTGRES_USER": self._stored.db_user,
+                "POSTGRES_PASSWORD": self._stored.db_password,
+                }
+        return pod_config
+
+    def _register_user(self, event):
+        """Registers a user for usage with Synapse."""
+        Result = collections.namedtuple(
```
jdkandersson, line 156: Should this be defined in a separate class rather than on each function invocation?

src/charm.py
```diff
@@ -0,0 +1,323 @@
...
+                "CommandExecResult", "return_code stdout stderr")
+
+        user = event.params["username"]
+        password = event.params["password"]
+        admin = event.params["admin"]
+        admin_switch = "--admin" if admin == "yes" else "--no-admin"
+
+        cmd = ["register_new_matrix_user",
```
jdkandersson, line 164: Would it be possible to use the SDK instead of commands? https://matrix.org/docs/projects/sdk/matrix-org-python-sdk

src/charm.py
```diff
@@ -0,0 +1,323 @@
...
+                  "Run command: %s, return code %s\nstdout: %s\nstderr:%s",
+                  cmd,
+                  return_code,
+                  result.stdout,
+                  result.stderr,
+         )
+
+    def _run_generate_synapse(self):
+        """Runs the generate command for synapse"""
+        Result = collections.namedtuple(
```
jdkandersson, line 190: Similar comment as above, this seems to also be a duplicate definition

src/charm.py
```diff
@@ -0,0 +1,323 @@
...
+                "CommandExecResult", "return_code stdout stderr")
+
+        cmd = ["/start.py", "generate"]
```
jdkandersson, line 193: Similar question regarding SDK as above

src/charm.py
```diff
@@ -0,0 +1,323 @@
...
+                  "Run command: %s, return code %s\nstdout: %s\nstderr:%s",
+                  cmd,
+                  return_code,
+                  result.stdout,
+                  result.stderr,
+         )
+
+    def _run_migrate_synapse(self):
+        """Runs the migrate command for synapse"""
+        Result = collections.namedtuple(
```
jdkandersson, line 216: Similar question as above

src/charm.py
```diff
@@ -0,0 +1,323 @@
...
+                "CommandExecResult", "return_code stdout stderr")
+        cmd = ["/start.py", "migrate_config"]
```
jdkandersson, line 218: Similar question as above regarding SDK

src/charm.py
```diff
@@ -0,0 +1,323 @@
...
+        process = self._container().exec(
+            cmd,
+            working_dir="/data",
+            environment=self._populate_synapse_env_settings(),
+        )
+        try:
```
jdkandersson, line 224: This also seems to be done in a similar way a few times, potentially worth separating out into a function? Perhaps we can use/ write a shared library?

src/charm.py
```diff
@@ -0,0 +1,323 @@
...
+        return_code = result.return_code
+        logger.debug(
+                  "Run command: %s, return code %s\nstdout: %s\nstderr:%s",
+                  cmd,
+                  return_code,
+                  result.stdout,
+                  result.stderr,
+         )
+
+    def _current_synapse_config(self):
```
jdkandersson, line 239: type hints? Or for the future?

src/charm.py
```diff
@@ -0,0 +1,323 @@
...
+                # Push an updated layer with the new config
+                container.add_layer(
+                        self._CONTAINER_NAME, self._pebble_layer, combine=True)
+                container.replan()
+
+                logger.debug(
+                        "Log level for synapse changed to '%s'", log_level
+                        )
+                self.unit.status = ActiveStatus()
+            else:
```
jdkandersson, line 283: Looks like a bit of nesting here, should we reduce that, e.g., using a sub function?

tests/integration/test_charm.py
```diff
@@ -0,0 +1,35 @@
...
+import logging
+from pathlib import Path
+
+import pytest
+import yaml
+from pytest_operator.plugin import OpsTest
+
+logger = logging.getLogger(__name__)
+
+METADATA = yaml.safe_load(Path("./metadata.yaml").read_text())
```
jdkandersson, line 15: Should this have an encoding as per the new standard? https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#file-encoding

tests/unit/test_charm.py
```diff
@@ -0,0 +1,74 @@
...
+
+import unittest
+
+import ops.testing
+from charm import MatrixOperatorCharm
+from ops.model import ActiveStatus, BlockedStatus, WaitingStatus
+from ops.testing import Harness
+
+
+class TestCharm(unittest.TestCase):
```
jdkandersson, line 14: Should this be written in pytest rather than unittest?

src/charm.py
merkata: Yes, exactly, I added a clarifying paragraph in the docstring of the function, thanks!

src/charm.py
merkata: My intent is to run a generate for the template once and create a `homeserver.yaml` whenever a config changes, my understanding was that I could use `on.install` for a one-time initial setup like for the generate function. There might be a better way, as I see that `on.install` relates to unit setup and not to charm setup, I might've been lucky with the ordering as I haven't faced the error. I think it's safe to remove it and rely on the `on.start` or consider moving the logic to `on.config_changed` altogether.

config.yaml
nrobinaubertin: This description could link to https://matrix-org.github.io/synapse/latest/setup/installation.html#choosing-your-server-name

.gitignore
merkata: Thanks, that's a better `.gitignore`, I adjusted it. I used the one generated by `charmcraft` and only added terraform specific ignores.

CONTRIBUTING.md
merkata: Not really, deleted it - thanks!

README.md
merkata: Thanks, changed it and cleaned up the README a bit.

src/charm.py
nrobinaubertin: It actually won't like `pgsql.MasterChangedEvent` either

actions.yaml
merkata: Added some more information for the fields.

config.yaml
merkata: Makes sense, I update the description and added an example of a valid server name.

homeserver.yaml
merkata: Yes, though that file is an old artifact so I deleted it altogether, it's not needed. Thanks!

homeserver.yaml
merkata: It doesn't, I deleted the file, it creeped into version control. The config lifecycle is via a `generate` command in the container that places a template file and then a `migrate` command that based on the template and environment variables set creates the static `homeserver.yaml` file.

homeserver.yaml
merkata: I know, right? :) The path taken here is to have synapse listen locally and terminate TLS on the ingress. So that port shouldn't be exposed and clients shouldn't be able to talk to synapse directly.

homeserver.yaml
merkata: see above, file deleted

homeserver.yaml
merkata: no (although it is created during `migrate` and is a one-time password, so there is no future reuse)

homeserver.yaml
merkata: same as above

homeserver.yaml
merkata: same as above

metadata.yaml
```diff
@@ -0,0 +1,39 @@
...
+storage:
+  data:
+    type: filesystem
+    location: /data
+
+requires:
+  nginx-route:
+    interface: nginx-route
+  db:
+    interface: pgsql
```
nrobinaubertin, line 38: We probably want to use the new `postgresql_client` interface (https://github.com/canonical/charm-relation-interfaces/blob/main/interfaces/postgresql_client/v0/README.md)

This should also use the lib from https://github.com/canonical/data-platform-libs.

But this represents some work and could be for another PR.

pyproject.toml
merkata: Yes, this is just the skeleton brought to life by `charmcraft`.

src/charm.py
merkata: It's reusing code from our charms, I'm fine to change it. Can you post a link to the recommendation?

config.yaml
nrobinaubertin: I feel like this config option does not make justice of the fact that the server name cannot be changed later. It should probably be retrieved from a configuration file instead.

src/charm.py
merkata: Thanks, that's from reusing code from one of our charms, I intend to address these when having CI in place, I'm currently not running the code through a linter.

src/charm.py
merkata: It could be and I'm fine with it being a tuple, again it is taken directly from our charms and is consistent with how it is implemented there.

src/charm.py
merkata: Updated the description.

src/charm.py
merkata: It won't like at least two Metallica albums either...

src/charm.py
merkata: Yes, same approach as in wordpress charm, this is keeping a list of env variables that are fed to pebble.

src/charm.py
merkata: Could be, I'd defer it to later though.

src/charm.py
merkata: I haven't looked into it, for the PoC I went with what is used and described in the synapse docs for the dockerized version. If there are more commands coming up, it makes sense to use an SDK. For now, I think it is just this one and it is tied to the PoC - I don't expect to create users via that action except maybe for admins?

src/charm.py
merkata: I've generalized the run command in a separate function to at least not have duplicate code.

src/charm.py
merkata: I'm not sure whether `generate` and `migrate` are covered in the SDK, I need to check. They seem auxiliary functions to help with config management.

src/charm.py
merkata: Answered above.

src/charm.py
merkata: Answered above.

src/charm.py
merkata: I've separated it out, I like the idea of a shared library - I've seen such functionality in other charms we have and it makes sense to reuse it via a library.

src/charm.py
merkata: I'm tempted to say for the future, although it should be the very near future while the code base is still small. I expect a lot of things to change with the received feedback, so I suggest to pause for until the PoC is done.

tests/integration/test_charm.py
merkata: It should, however these are tests generated by `charmcraft`, there are no tests written yet as part of the PoC.

tests/unit/test_charm.py
merkata: It should, `charmcraft` still generates tests in `unittest`, I haven't changed it yet.

metadata.yaml
merkata: I agree, I even have it in the README, as I noticed as well that the old interface is getting deprecated. The path I took is to reuse what we have and I know it is working (and how), swapping it out should be easy, it is some additional work like you've said. I had feedback on using `traefik` instead of `nginx-ingress-integrator` and that would be fine as well, still in terms of architecture there is no difference, it's just a design decision and a component we're swapping.

config.yaml
merkata: the charm uses this as a last resort, SYNAPSE_SERVER_NAME env takes precedence

src/charm.py
amandahla: When I checked the Postgresql database, there were no tables and in the pod there was a homeserver.db file . Am I missing something? 🤔

Here is my Juju status:
```
 juju status --relations
Model      Controller       Cloud/Region        Version  SLA          Timestamp
synapsium  github-pr-a25ab  microk8s/localhost  2.9.42   unsupported  10:44:07-03:00

App      Version                       Status  Scale  Charm                     Channel        Rev  Address         Exposed  Message
db       res:postgresql-image@8a72e11  active      1  postgresql-k8s            latest/stable   20  10.152.183.119  no       Pod configured
ingress  25.3.0                        active      1  nginx-ingress-integrator  latest/stable   59  10.152.183.187  no       
synapse                                active      1  synapse                                    3  10.152.183.34   no       

Unit        Workload  Agent  Address      Ports     Message
db/0*       active    idle   10.1.36.234  5432/TCP  Pod configured
ingress/0*  active    idle   10.1.36.228            
synapse/0*  active    idle   10.1.36.241            

Relation provider  Requirer    Interface  Type     Message
db:db              synapse:db  pgsql      regular  
db:peer            db:peer     peer       peer     
```

src/charm.py
```diff
@@ -0,0 +1,298 @@
...
+                    "startup": "enabled",
+                    "environment": self._populate_synapse_env_settings(),
+                }
+            },
+            "checks": {
+                "synapse-ready": {
+                    "override": "replace",
+                    "level": "alive",
+                    "http": {
+                        "url": "http://localhost:8008/health",
```
amandahla, line 290: Minor (and I don't know if it was already mentioned): Could "8008" be a constant?

src/charm.py
```diff
@@ -0,0 +1,298 @@
...
+        return self.unit.get_container(self._CONTAINER_NAME)
+
+    def _populate_synapse_env_settings(self):
+        """Populate env settings to pass around."""
+        if os.getenv("SYNAPSE_SERVER_NAME") is None:
+            os.environ["SYNAPSE_SERVER_NAME"] = self.config["server_name"]
+        if os.getenv("SYNAPSE_REPORT_STATS") is None:
+            os.environ["SYNAPSE_REPORT_STATS"] = self.config["report_stats"]
+        pod_config = {
+                "SYNAPSE_SERVER_NAME": os.getenv("SYNAPSE_SERVER_NAME"),
```
weiiwang01, line 153: I am a little curious why not directly retrieve the server name from configuration, instead of saving the variable in the charm environment variables then put into this env setting dict?

```python
        pod_config = {
                 "SYNAPSE_SERVER_NAME": self.config["server_name"],
        ...
        }
```

src/charm.py
merkata: makes sense, I'll change it

src/charm.py
merkata: Well (for now) the server name has a default of "synapse-local" which should be a default fallback in case you want to spin up synapse locally (and have synapse-local resolve to 127.0.0.1). The docker version of synapse relies on you exposing these two env variables before starting the service and here I'm replicating that behavior - if you haven't set these variables, you would use the config fallback. That is not the best thing, changing the server name very much creates a new synapse server under a new name, so some thought needs to be put as to how to manage that.

weiiwang01: The charm environment variables is not persisted across events. For example, when charm finished handling the `config-changed` event, charm will actual terminate itself. And for the next event, for example, `pebble-ready`, a new charm process will spawn and the charm environment variables set during the last event handling will be gone. Is this acceptable?

merkata: I think I get your point - if I understand correctly, that way I will always set the SYNAPSE_SERVER_NAME to self.config["server_name"], as on any new event, it does not get populated (and does not persist from a previous event). Then the proper way perhaps is to have a default on "server_name" to be "" and a check in the src/charm.py that blocks the charm if it is "", so until we set it to something that we intend to use. Thoughts?

gregory-schiano: See my comment in the config.yaml concerning this option

src/charm.py
gregory-schiano: I'd also start with if "not can_connect" to exit the non happy cases first same for if "db_password is None"

gregory-schiano (commented):
Made a quick first pass with obvious first comments
As global comments I'd add:
- Apply our standard following the IS Charm guidelines (see what Weii did for Spring Boot and Flask for example)
- Try to follow the ISD014 charm architectures rules (mainly implement a State and separate business logic in services)
- Add a dockerfile or rockcraft to provide our own OCI image

PR #5, opened 2023-06-12
chore(ISD-782): Add Synapse rock

amandahla:
Add Synapse rockfile based on https://github.com/matrix-org/synapse/blob/develop/docker/Dockerfile

obs.: Branch with wrong name, the right one would be ISD-782

Bash package is needed otherwise SQLite throws an error "sqlite3.OperationalError: disk I/O error"

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -15,3 +15,11 @@ maintainers:
 source: https://github.com/canonical/synapse-operator
 assumes:
   - k8s-api
+containers:
+  synapse:
+    resource: synapse-image
+resources:
+  synapse-image:
+    type: oci-image
+    description: Docker image for Synapse
+    auto-fetch: true
--- a/synapse_rock/rockcraft.yaml
+++ b/synapse_rock/rockcraft.yaml
@@ -1,0 +1,80 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+name: synapse
+summary: Synapse rock
+description: Synapse OCI image for the Synapse charm
+version: "1.0"
+base: bare
+build-base: ubuntu:22.04
+license: Apache-2.0
+platforms:
+  amd64:
+parts:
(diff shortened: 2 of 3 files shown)


synapse_rock/rockcraft.yaml
```diff
@@ -0,0 +1,80 @@
...
+            - libxslt1-dev
+            - openssl
+            - zlib1g-dev
+            - libicu-dev
+            - python3-dev
+            - python3-pip
+            - python3-setuptools
+        stage-packages:
+            - bash
+            - build-essential
```
arturo-seijas, line 35: Do you need to stage `build-essential` ?

gregory-schiano: Same comment for the `-dev` deps, they're usually only needed when you build

gregory-schiano: The final stage in the dockerfile doesn't install all these:
https://github.com/matrix-org/synapse/blob/develop/docker/Dockerfile#L168-L181

amandahla: Removed, thanks

arturo-seijas (approved):

metadata.yaml
```diff
@@ -15,3 +15,11 @@ maintainers:
...
 assumes:
   - k8s-api
+containers:
+  synapse:
+    resource: synapse-image
+resources:
+  synapse-image:
+    type: oci-image
+    description: Docker image for Synapse
+    auto-fetch: true
```
mthaddon, line 25: I believe this is obsolete?

merkata (approved):

metadata.yaml
amandahla: Thanks! Since this PR has 2 approvals already, I'll merge it and then fix this on https://github.com/canonical/synapse-operator/pull/6 ok?

PR #6, opened 2023-06-13
feat(ISD-782): Create initial version of the Synapse charm

amandahla:
This PR creates a small and initial version of Synapse charm. There are two configurations available server_name and report_stats. After the user configures server_name, the charm runs the command "migrate_config" and then starts.

After deploying the charm, you can check the result with "curl http://<< unit IP address >>:8008"


metadata.yaml
```diff
@@ -15,3 +15,21 @@ maintainers:
...
+    resource: synapse-image
+    mounts:
+      - storage: data
+        location: /data
+
+resources:
+  synapse-image:
+    type: oci-image
+    description: OCI image for Synapse
+    auto-fetch: true
```
arturo-seijas, line 30: I don't think we need this `auto-fetch`

src/synapse.py
```diff
@@ -0,0 +1,135 @@
...
+    """
+    if not is_server_name_valid(state):
+        raise CharmConfigInvalidError(
+            "The server_name is empty, please review your charm configuration"
+        )
+    if not is_configuration_valid(state):
+        raise CharmConfigInvalidError(
+            "Configuration is not valid, please review your charm configuration"
+        )
+    migrate_config_command = [COMMAND_PATH, "migrate_config"]
```
merkata, line 100: I would put a TODO comment here about looking into migrating to SDK usage instead of invoking the command

metadata.yaml
amandahla: Removed, thanks

arturo-seijas (dismissed):
LGTM

src/synapse.py
amandahla: I'm not sure if it's possible to use the SDK to generate the configuration file. Maybe using a template?

src/synapse.py
```diff
@@ -0,0 +1,135 @@
...
+    Args:
+        state: The state of the charm.
+
+    Returns:
+        A dictionary representing the Synapse environment variables.
+    """
+    return {
+        "SYNAPSE_SERVER_NAME": f"{state.server_name}",
+        "SYNAPSE_REPORT_STATS": f"{state.report_stats}",
+        "SYNAPSE_NO_TLS": "True",
```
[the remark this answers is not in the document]
amandahla: good point, thanks

tests/integration/conftest.py
```diff
@@ -0,0 +1,80 @@
...
+
+@pytest_asyncio.fixture(scope="module", name="build_charm")
+async def build_charm_fixture(ops_test) -> str:
+    """Build the charm"""
+    charm = await ops_test.build_charm(".")
+    return charm
+
+
+@pytest_asyncio.fixture(scope="module", name="synapse_app")
+async def synapse_app_fixture(
```
jdkandersson, line 36: Usually the `fixture_` is a prefix?

tests/integration/conftest.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+@pytest_asyncio.fixture(scope="module", name="model")
+async def fixture_model(ops_test: OpsTest) -> Model:
+    """Return the current testing juju model."""
+    assert ops_test.model
+    return ops_test.model
+
+
+@pytest_asyncio.fixture(scope="module", name="build_charm")
```
jdkandersson, line 28: Should the fixture name be what is returned by the fixture rather than what it does? E.g., just call it `charm`?

tests/integration/conftest.py
```diff
@@ -0,0 +1,80 @@
...
+async def build_charm_fixture(ops_test) -> str:
+    """Build the charm"""
+    charm = await ops_test.build_charm(".")
+    return charm
+
+
+@pytest_asyncio.fixture(scope="module", name="synapse_app")
+async def synapse_app_fixture(
+    build_charm: str,
```
jdkandersson, line 37: Is this type hint right?

tests/integration/conftest.py
```diff
@@ -0,0 +1,80 @@
...
+    model: Model,
+    server_name: str,
+    pytestconfig: Config,
+):
+    """Build and deploy the Synapse charm."""
+    app_name = "synapse-k8s"
+
+    resources = {
+        "synapse-image": pytestconfig.getoption("--synapse-image"),
```
jdkandersson, line 46: Should this be a fixture that checks that the argument has been provided and provides a nice message when it isn't?

tests/unit/test_charm.py
```diff
@@ -0,0 +1,112 @@
...
+def test_synapse_pebble_layer(harness: Harness) -> None:
+    """
+    arrange: none
+    act: start the Synapse charm, set Synapse container to be ready and set server_name.
+    assert: Synapse charm should submit the correct Synapse pebble layer to pebble.
+    """
+    harness.begin_with_initial_hooks()
+    harness.set_can_connect(harness.model.unit.containers[SYNAPSE_CONTAINER_NAME], True)
+    harness.disable_hooks()
+    server_name = token_hex(16)
```
jdkandersson, line 25: Perhaps set it to something that identifies the test?

config.yaml
```diff
@@ -1,7 +1,15 @@
...
-# This file defines charm config options, and populates the Configure tab on Charmhub.
-# If your charm does not require configuration options, delete this file entirely.
-#
-# See https://juju.is/docs/config for guidance.
+options:
+  server_name:
+    type: string
+    description: Synapse server name. Must be set to deploy the charm.
+    default: ""
+  report_stats:
```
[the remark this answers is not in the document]
amandahla: report_stats configuration option is a boolean but SYNAPSE_REPORT_STATS used by start.pt migrate_config is an environment var that expects "yes" or "no".
Reference: https://github.com/matrix-org/synapse/tree/develop/docker

src/charm_state.py
```diff
@@ -0,0 +1,45 @@
...
+        return self._charm.config["server_name"]
+
+    @property
+    def report_stats(self) -> str:
+        """Return report_stats config.
+
+        Returns:
+            str: report_stats config.
+        """
+        return self._charm.config["report_stats"]
```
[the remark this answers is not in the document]
amandahla: Good point but this is already checked by the script that generates the configuration file, checking here might be redundant.

src/synapse.py
```diff
@@ -0,0 +1,134 @@
...
+            "migrate config failed, stdout: %s, stderr: %s",
+            migrate_config_result.stdout,
+            migrate_config_result.stderr,
+        )
+        raise CommandMigrateConfigError(
+            "Migrate config failed, please review your charm configuration"
+        )
+
+
+def _exec(
```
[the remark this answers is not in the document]
amandahla: Yes, it can be used for the action to register user to be addressed in a future PR/story. If not, then it can be refactored.

src/synapse.py
```diff
@@ -0,0 +1,134 @@
...
+
+    Args:
+        container: Container of the charm.
+        state: State of the charm.
+
+    Raises:
+        CommandMigrateConfigError: something went wrong running migrate_config.
+        CharmConfigInvalidError: charm configuration is not valid.
+    """
+    if not is_server_name_valid(state):
```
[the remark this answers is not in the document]
amandahla: Thanks, changed :)

src/synapse.py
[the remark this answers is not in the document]
amandahla: Sure, changed.

config.yaml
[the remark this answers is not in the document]
jdkandersson: So we can map the boolean to that? It will help reduce the possibility of mistakes since operators can't accidentally misspell "yes" or "no"

src/charm_state.py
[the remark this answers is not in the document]
jdkandersson: Perhaps the script generating the config doesn't need to check and pydantic should be used instead?

config.yaml
```diff
@@ -1,7 +1,15 @@
...
 # See LICENSE file for licensing details.
 
-# This file defines charm config options, and populates the Configure tab on Charmhub.
-# If your charm does not require configuration options, delete this file entirely.
-#
-# See https://juju.is/docs/config for guidance.
+options:
+  server_name:
+    type: string
+    description: Synapse server name. Must be set to deploy the charm.
```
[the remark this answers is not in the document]
amandahla: Yes, I added a note about how this corresponds to server_name on Synapse configuration file.

config.yaml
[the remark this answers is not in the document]
amandahla: Changed.

src/charm_state.py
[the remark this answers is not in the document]
amandahla: Solved by converting to boolean.

src/charm.py
```diff
@@ -3,4 +3,76 @@
...
+        self.unit.status = ops.ActiveStatus()
+
+    @property
+    def _pebble_layer(self) -> Dict:
+        """Return a dictionary representing a Pebble layer."""
+        return {
+            "summary": "Synapse layer",
+            "description": "pebble config layer for Synapse",
+            "services": {
+                "synapse": {
```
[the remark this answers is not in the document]
amandahla: Nice, changed for constants file.

tests/unit/test_charm.py
amandahla: Hum...I'm confused, David. Isn't that the same case that you suggested [here](https://github.com/canonical/indico-operator/pull/248#discussion_r1139815154) to use token_hex instead of a value?

tests/integration/conftest.py
amandahla: It is? Here it's like a suffix: https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#test-fixture
and here in [Flask](https://github.com/canonical/flask-k8s-operator/blob/main/tests/integration/conftest.py#L96) (my model 😊 ) too

tests/integration/conftest.py
amandahla: I assumed that since in [Flask](https://github.com/canonical/flask-k8s-operator/blob/main/tests/integration/conftest.py#L57) was written like that then it would be ok. But I agree on that too. I'm changing to "synase_charm".

tests/integration/conftest.py
amandahla: Well, ops_test.build actually return a pathlib.Path but deploy expects a str so I'm leaving like that.

tests/integration/conftest.py
amandahla: This will be kind of different from other charms but an additional check cant hurt :) Changed.

src/charm.py
```diff
@@ -3,4 +3,82 @@
...
+        if not container.can_connect():
+            event.defer()
+            self.unit.status = ops.WaitingStatus("Waiting for pebble")
+            return
+        self.model.unit.status = ops.MaintenanceStatus("Configuring Synapse")
+        try:
+            self._synapse.execute_migrate_config(container)
+        except CommandMigrateConfigError as exc:
+            self.model.unit.status = ops.BlockedStatus(exc.msg)
+            event.defer()
```
weiiwang01, line 56: I don't think the defer here is necessary since the action of user correcting the configuration will trigger another config-changed event.

src/charm_state.py
```diff
@@ -0,0 +1,145 @@
...
+
+        Attrs:
+            extra: extra configuration.
+        """
+
+        extra = Extra.allow
+
+    @validator("server_name")
+    # no-self-argument disabled due to cls usage as expected by validator
+    def check_server_name_empty(cls, value: str) -> str:  # pylint: disable=no-self-argument
```
weiiwang01, line 53: Isn't the `server_name: str | None = Field(None, min_length=0)` already validate that server name can't be empty?

src/charm_state.py
```diff
@@ -0,0 +1,145 @@
...
+        Returns:
+            server_name value.
+        """
+        if not value:
+            raise ValueError("The server_name is empty, please review your charm configuration")
+        return value
+
+    @validator("report_stats")
+    @classmethod
+    def to_yes_or_no(cls, value: str) -> str:
```
weiiwang01, line 71: I think if the `report_stats` is defined as `report_stats: typing.Literal["yes", "no"]`, then the validation function here is not needed.

config.yaml
```diff
@@ -1,7 +1,16 @@
...
-#
-# See https://juju.is/docs/config for guidance.
+options:
+  server_name:
+    type: string
+    description: |
+      Synapse server name. Must be set to deploy the charm. Corresponds to the
+      server_name option on Synapse configuration file and sets the
+      public-facing domain of the server.
+    default: ""
```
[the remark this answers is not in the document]
amandahla: Good point, removed!

src/charm_state.py
amandahla: Yes, I changed the min_length, thanks

src/synapse.py
```diff
@@ -0,0 +1,107 @@
...
+
+from charm_state import CharmState
+from charm_types import ExecResult
+from constants import CHECK_READY_NAME, COMMAND_MIGRATE_CONFIG, COMMAND_PATH, SYNAPSE_PORT
+from exceptions import CommandMigrateConfigError
+
+logger = logging.getLogger(__name__)
+
+
+class Synapse:  # pylint: disable=too-few-public-methods
```
gregory-schiano, line 22: Do you really need to disable the too-few-public-methods ? There are 3 public methods and the default in pylint is 2

src/synapse.py
```diff
@@ -0,0 +1,107 @@
...
+        """Generate a environment dictionary from the charm configurations.
+
+        Returns:
+            A dictionary representing the Synapse environment variables.
+        """
+        return {
+            "SYNAPSE_SERVER_NAME": f"{self._charm_state.server_name}",
+            "SYNAPSE_REPORT_STATS": f"{self._charm_state.report_stats}",
+            # TLS disabled so the listener is HTTP. HTTPS will be handled by Traefik.
+            "SYNAPSE_NO_TLS": str(True),
```
gregory-schiano, line 56: So we don't support E2E TLS ?
Probably not a big deal for now, but since it's a chat app supposed to be pretty secured, E2E TLS will be probably interesting to support.
I don't even know if our ingresses supports an HTTPS backend

src/synapse.py
```diff
@@ -0,0 +1,107 @@
...
+            logger.error(
+                "migrate config failed, stdout: %s, stderr: %s",
+                migrate_config_result.stdout,
+                migrate_config_result.stderr,
+            )
+            raise CommandMigrateConfigError(
+                "Migrate config failed, please review your charm configuration"
+            )
+
+    def _exec(
```
gregory-schiano, line 84: Is it worth it to make this a method only used once ?

src/charm.py
amandahla: Changed. :)

src/synapse.py
amandahla: Just in case, I added the comment, thanks!

src/synapse.py
amandahla: Yes, it will be used for future stories.

src/charm.py
```diff
@@ -3,4 +3,82 @@
...
+        """Return a dictionary representing a Pebble layer."""
+        return {
+            "summary": "Synapse layer",
+            "description": "pebble config layer for Synapse",
+            "services": {
+                SYNAPSE_SERVICE_NAME: {
+                    "override": "replace",
+                    "summary": "Synapse application service",
+                    "startup": "enabled",
+                    "command": COMMAND_PATH,
```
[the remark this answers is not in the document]
amandahla: Good point, changed.

src/synapse.py
amandahla: Good catch :) Removed.

src/synapse.py
amandahla: Not for now. I added a comment so we can check this in the future.

src/charm_state.py
weiiwang01: Gotcha, thanks.

jdkandersson (approved):

PR #8, opened 2023-06-21
feat(ISD-847): Prevent server_name to be changed

amandahla:
Prevent server_name to be change by comparing with the one in the configuration file if exists.


src/constants.py
```diff
@@ -5,6 +5,8 @@
 
 CHECK_READY_NAME = "synapse-ready"
 COMMAND_MIGRATE_CONFIG = "migrate_config"
+SYNAPSE_CONFIG_DIR = "/data"
+SYNAPSE_CONFIG_PATH = SYNAPSE_CONFIG_DIR + "/homeserver.yaml"
```
arturo-seijas, line 9: Why not an f-string?

src/synapse.py
```diff
@@ -83,6 +91,46 @@ def execute_migrate_config(self, container: ops.Container) -> None:
...
+            "reset-instance": False,
+        }
+        logging.debug("Erasing directory %s", SYNAPSE_CONFIG_DIR)
+        try:
+            container.remove_path(SYNAPSE_CONFIG_DIR, recursive=True)
+        except PathError as path_error:
+            logger.debug(
+                "exception while erasing directory %s: %s", SYNAPSE_CONFIG_DIR, path_error
+            )
+        results["reset-instance"] = True
```
arturo-seijas, line 131: Isn't this supposed to be inside the try block? The result would otherwise be always true

src/constants.py
amandahla: Golang bad habit 🤦Thanks for pointing that out!

src/synapse.py
amandahla: Thanks, you made me realize that ignoring this error is not a good approach. I changed the action and now the charm will be blocked and the event will fail if something goes wrong while deleting the directory.

actions.yaml
```diff
@@ -0,0 +1,5 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+reset-instance:
+  description: Set a new server_name before running this action. This will erase all existing data.
```
jdkandersson, line 5: Should this describe why this action might be required?

requirements.txt
```diff
@@ -1,2 +1,3 @@
+jsonschema
```
jdkandersson, line 1: Should this version be pinned at least to a major version?

src/charm.py
```diff
@@ -52,18 +52,39 @@ def __init__(self, *args: Any) -> None:
             host=f"{self.app.name}-endpoints.{self.model.name}.svc.cluster.local",
             strip_prefix=True,
         )
+        self.framework.observe(self.on.config_changed, self._on_config_changed)
+        self.framework.observe(self.on.reset_instance_action, self._reset_instance_action)
```
jdkandersson, line 56: Should this be `self._on_reset_instance_action`?

src/charm.py
```diff
@@ -52,18 +52,39 @@ def __init__(self, *args: Any) -> None:
...
+            self._charm_state = CharmState.from_charm(charm=self)
+        except CharmConfigInvalidError as exc:
+            self.model.unit.status = ops.BlockedStatus(exc.msg)
+            return
         container = self.unit.get_container(SYNAPSE_CONTAINER_NAME)
         if not container.can_connect():
             event.defer()
             self.unit.status = ops.WaitingStatus("Waiting for pebble")
             return
+        server_name_configured = self._synapse.server_name_configured(container)
```
jdkandersson, line 74: perhaps `configured_server_name`? At the moment it reads like a boolean value of whether the server name is configured, I'm guessing this is either the configured server name or None?

src/charm.py
```diff
@@ -52,18 +52,39 @@ def __init__(self, *args: Any) -> None:
...
+            server_name_configured is not None
+            and server_name_configured != self._charm_state.server_name
+        ):
+            msg = (
+                f"server_name {self._charm_state.server_name} is different from the existing one"
+                f" {server_name_configured}."
+                " Please revert the config or"
+                " run the action reset-instance if you to erase the existing instance and start a"
+                " new one."
+            )
```
jdkandersson, line 85: Perhaps this would be better in the synapse module? Like `check server name` or something like that which takes the state as an argument?

src/charm.py
```diff
@@ -94,6 +115,35 @@ def _pebble_layer(self) -> Dict:
...
+        """Reset instance and report action result.
+
+        Args:
+            event: Event triggering the reset instance action.
+        """
+        results = {
+            "reset-instance": False,
+        }
+        if not self.model.unit.is_leader():
+            event.defer()
```
jdkandersson, line 128: @gregory-schiano I'm not sure on this, should this just do nothing unless the unit is the leader? I'm not sure if all the units get the action or just a specific unit?

jdkandersson: I guess:
* If all units get the action -> do nothing if the unit isn't the leader
* If only a specific unit gets the action -> fail the action if it isn't the leader

Not sure what deferring would do to an operator generated action

src/charm.py
```diff
@@ -94,6 +115,35 @@ def _pebble_layer(self) -> Dict:
...
+            return
+        container = self.unit.get_container(SYNAPSE_CONTAINER_NAME)
+        if not container.can_connect():
+            event.fail("Failed to connect to container")
+            return
+        self.model.unit.status = ops.MaintenanceStatus("Resetting Synapse instance")
+        try:
+            self._synapse.reset_instance_action(container)
+            self._synapse.execute_migrate_config(container)
```
jdkandersson, line 137: Should this be a single function on `synapse`? Perhaps a function that just calls these two methods?

src/synapse.py
```diff
@@ -83,6 +91,47 @@ def execute_migrate_config(self, container: ops.Container) -> None:
...
             )
 
+    def server_name_configured(self, container: ops.Container) -> str | None:
+        """Get server_name from configuration file.
+
+        Args:
+            container: Container of the charm.
+
+        Returns:
+            str | None: server_name or None if configuration file is not found.
```
jdkandersson, line 101: Under what conditions would the server name not be defined?

src/synapse.py
```diff
@@ -83,6 +91,47 @@ def execute_migrate_config(self, container: ops.Container) -> None:
...
+        try:
+            configuration_content = str(
+                container.pull(SYNAPSE_CONFIG_PATH, encoding="utf-8").read()
+            )
+        except PathError:
+            logger.error("configuration file %s does not exist", SYNAPSE_CONFIG_PATH)
+            return None
+        return yaml.safe_load(configuration_content)["server_name"]
+
+    def reset_instance_action(self, container: ops.Container) -> None:
```
jdkandersson, line 112: What is the reason for including the `_action` postfix in this function please? Just thinking that this is coupled with the charm.py in terms of the name?

src/synapse.py
```diff
@@ -83,6 +91,47 @@ def execute_migrate_config(self, container: ops.Container) -> None:
...
+            container: Container of the charm.
+
+        Raises:
+            PathError: if somethings goes wrong while erasing the Synapse directory.
+        """
+        logging.debug("Erasing directory %s", SYNAPSE_CONFIG_DIR)
+        try:
+            container.remove_path(SYNAPSE_CONFIG_DIR, recursive=True)
+        except PathError as path_error:
+            # The error "unlinkat //data: device or resource busy" is expected
```
jdkandersson, line 125: Would it be worth elaborating on why it is expected?

src/charm.py
amandahla: In the case of Synapse, for now, this check is just in case...Scaling will not be allowed and will be addressed in a different story/PR.

src/charm.py
amandahla: I see your point but since execute_migrate_config is executed here and also on config-changed, I believe would be better for now to let them explicitly separated.

src/synapse.py
amandahla: I followed the example here: https://juju.is/docs/sdk/actions "_on_grant_admin_role_action"
But I don't mind changing it. Is there a different standard/guideline for this?

src/charm.py
amandahla: Good catch, completely forgot. Changed.

src/charm.py
```diff
@@ -52,18 +52,39 @@ def __init__(self, *args: Any) -> None:
...
+            and server_name_configured != self._charm_state.server_name
+        ):
+            msg = (
+                f"server_name {self._charm_state.server_name} is different from the existing one"
+                f" {server_name_configured}."
+                " Please revert the config or"
+                " run the action reset-instance if you to erase the existing instance and start a"
+                " new one."
+            )
+            self.model.unit.status = ops.BlockedStatus(msg)
```
yanksyoon, line 86: Minor nitpick, I'm wondering if this message is too lengthy to be set as status message. Would logging be better and the status message direct the users to the logs?

tests/integration/test_charm.py
```diff
@@ -61,3 +62,47 @@ async def test_with_ingress(
...
+    assert "is different from the existing" in unit.workload_status_message
+
+
+async def test_reset_instance_action(
+    model: Model, synapse_app: Application, different_server_name: str
+):
+    """
+    arrange: build and deploy the Synapse charm.
+    act: change server_name via juju config.
+    assert: the old instance is deleted and the new one configured.
```
yanksyoon, line 90: Minor nitpick, how about using a fixture here? Another synapse charm fixture with a "function" scope might be useful :D 
I.e.
```python3
@pytest_asyncio.fixture(scope="function", name="name_configured_app")
async def name_configured_app_fixture(synapse_app: Application):
    await synapse_app.set_config({"server_name":different_server_name})
    
    yield synapse_app
    
    await synaps_app.reset_config(["server_name"])
```

This will make the tests flow a bit more focused as:
```python3
"""
arrange: given a configured synapse application.
act: when reset instance action is triggered.
assert: the old instance is deleted and the new one configured.
"""
```

tests/unit/conftest.py
```diff
@@ -156,6 +158,44 @@ def fixture_harness_server_name_configured(harness: Harness) -> Harness:
     harness.update_config({"server_name": TEST_SERVER_NAME})
     harness.enable_hooks()
     harness.begin_with_initial_hooks()
+    container: ops.Container = harness.model.unit.get_container(SYNAPSE_CONTAINER_NAME)
+    container.push(f"{SYNAPSE_CONFIG_PATH}", f'server_name: "{TEST_SERVER_NAME}"', make_dirs=True)
```
yanksyoon, line 162: I think the fstring on the first path can be removed :D
```suggestion
    container.push(SYNAPSE_CONFIG_PATH, f'server_name: "{TEST_SERVER_NAME}"', make_dirs=True)
```

tests/unit/test_charm.py
```diff
@@ -107,3 +108,112 @@ def test_traefik_integration(harness_server_name_configured: Harness) -> None:
...
+    """
+    harness = harness_server_name_configured
+    harness.set_leader(True)
+    server_name_changed = "pebble-layer-1.synapse.com"
+    harness.update_config({"server_name": server_name_changed})
+    assert isinstance(harness.model.unit.status, ops.BlockedStatus)
+    assert "is different from the existing" in str(harness.model.unit.status)
+    event = unittest.mock.Mock()
+    # Calling to test the action since is not possible calling via harness
+    harness.charm._reset_instance_action(event)  # pylint: disable=protected-access
```
yanksyoon, line 163: I've seen in other charms that `# pylint: disable=protected-access` is at the top of the module as a module scoped disable, would this be helpful here?

yanksyoon (commented):
Very nice!

actions.yaml
amandahla: Sure, changed.

src/charm.py
amandahla: Thanks for the suggestion! Changed.

src/charm.py
amandahla: Agree, moved :)

src/synapse.py
amandahla: Sure, added a note about that.

src/charm.py
amandahla: Good catch, changed :)

tests/integration/test_charm.py
amandahla: Nice suggestion, thanks!

tests/unit/conftest.py
amandahla: Oops...Fixed ;)

tests/unit/conftest.py
```diff
@@ -156,6 +158,44 @@ def fixture_harness_server_name_configured(harness: Harness) -> Harness:
...
     harness.set_can_connect(harness.model.unit.containers[SYNAPSE_CONTAINER_NAME], True)
     harness.framework.reemit()
     return harness
+
+
+@pytest.fixture(name="container_mocked")
+def fixture_container_mocked(monkeypatch: pytest.MonkeyPatch) -> unittest.mock.MagicMock:
```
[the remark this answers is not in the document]
amandahla: Fixed, thanks!

tests/unit/test_charm.py
amandahla: Nice, better than repeating, changed!

src/synapse.py
jdkandersson: That standard is more applicable to `charm.py`, within the synapse module it isn't applicable. The applicable standard is the guidance in the managing charm complexity doc around that modules shouldn't be charm aware. Does that make sense?

src/charm.py
```diff
@@ -53,43 +53,32 @@ def __init__(self, *args: Any) -> None:
...
-                f" {server_name_configured}."
-                " Please revert the config or"
-                " run the action reset-instance if you to erase the existing instance and start a"
-                " new one."
-            )
-            self.model.unit.status = ops.BlockedStatus(msg)
-            return
-        self.model.unit.status = ops.MaintenanceStatus("Configuring Synapse")
         try:
+            self._charm_state = CharmState.from_charm(charm=self)
```
jdkandersson, line 70: In most cases we do this in the `__init__` function, would that work here?

src/charm.py
```diff
@@ -53,43 +53,32 @@ def __init__(self, *args: Any) -> None:
...
+            self._synapse = Synapse(charm_state=self._charm_state)
+            self._synapse.check_server_name(container)
+            self.model.unit.status = ops.MaintenanceStatus("Configuring Synapse")
             self._synapse.execute_migrate_config(container)
```
jdkandersson, line 74: DO you think that this function should check for the server name as a pre-condition? That reduces the coupling between `charm.py` and the `synapse` module

src/charm.py
amandahla: The event will fail with the message "Only the juju leader unit can run reset instance action" now.

src/charm.py
amandahla: I'm glad that you asked because I wanted to ask you and @gregory-schiano about that and I forgot to do it yesterday.
While comparing the server_name I noticed that the state isn't updated when the config changes. From what is described [here](https://juju.is/docs/sdk/create-a-minimal-kubernetes-charm) I was expecting that the init would be called on config-changed event so I don't know what I'm missing here.

actions.yaml
gregory-schiano, line 1: I'm wondering if we shouldn't then recommend the user to deploy a new instance of the charm instead of providing a such action ?

So keep the check for server_name change to avoid it, but just instructing the user to deploy a new instance (and remove the current one if the don't need it anymore)

src/charm.py
gregory-schiano: Well this is already done in the `__init__`see L41
So I think this is not needed

src/charm.py
```diff
@@ -64,11 +66,19 @@ def _on_config_changed(self, event: ops.HookEvent) -> None:
             event.defer()
             self.unit.status = ops.WaitingStatus("Waiting for pebble")
             return
-        self.model.unit.status = ops.MaintenanceStatus("Configuring Synapse")
         try:
+            self._charm_state = CharmState.from_charm(charm=self)
+            self._synapse = Synapse(charm_state=self._charm_state)
```
gregory-schiano, line 71: Similarly to the CharmState, this service is already instanciated in the constructor, so I think it is not needed here

src/synapse.py
```diff
@@ -83,6 +91,76 @@ def execute_migrate_config(self, container: ops.Container) -> None:
...
+            configuration_content = str(
+                container.pull(SYNAPSE_CONFIG_PATH, encoding="utf-8").read()
+            )
+            configured_server_name = yaml.safe_load(configuration_content)["server_name"]
+            if (
+                configured_server_name is not None
+                and configured_server_name != self._charm_state.server_name
+            ):
+                msg = (
+                    f"server_name {self._charm_state.server_name} is different from the existing "
```
gregory-schiano, line 117: I think it is worthwhile adding a sentence explaining that the server_name can't be changed and pointing to some Synapse documentation

src/synapse.py
```diff
@@ -83,6 +91,76 @@ def execute_migrate_config(self, container: ops.Container) -> None:
...
+            # The files will be removed but SYNAPSE_CONFIG_DIR directory will
+            # remain.
+            if "device or resource busy" in str(path_error):
+                pass
+            else:
+                logger.error(
+                    "exception while erasing directory %s: %s", SYNAPSE_CONFIG_DIR, path_error
+                )
+                raise
+
```
gregory-schiano, line 163: What about the database ? It'll still be filled with the previous data.

src/charm.py
```diff
@@ -94,6 +104,35 @@ def _pebble_layer(self) -> Dict:
             },
         }
 
+    def _on_reset_instance_action(self, event: ActionEvent) -> None:
```
gregory-schiano, line 107: This action is pretty dangerous, shouldn't we add a parameter like "config" that needs to be set to "IUNDERSTANDITWILLDELETEALLMYDATA" (or something shorter 😄

tests/integration/test_charm.py
```diff
@@ -61,3 +62,41 @@ async def test_with_ingress(
...
+    assert unit.workload_status == "blocked"
+    assert "server_name modification is not allowed" in unit.workload_status_message
+
+
+async def test_reset_instance_action(
+    model: Model, another_synapse_app: Application, another_server_name: str
+):
+    """
+    arrange: build and deploy the Synapse charm.
+    act: change server_name via juju config.
```
gregory-schiano, line 85: I don't think the arrange and act are properly set
It should be something like
`arrange: a deployed Synapse charm in a blocked state due to a server_name change`
`act: call the reset_instance action

tests/unit/test_charm.py
gregory-schiano, line 1: I'm wondering if we shouldn't also unit test the Synapse module, these tests are almost integration test
So instead of calling the synapse service we could mock synapse service methods so we really unit test both modules and their contract

src/synapse.py
amandahla: Oh sorry, I didn't notice. Changed!

src/charm.py
amandahla: Makes sense, changed.

src/charm.py
amandahla: If I dont call it again, the state is not updated with the new config :|

actions.yaml
amandahla: This action will make more sense in the future with all the integrations (saml, smtp, PostgreSQL,S3). It will erase only the database and recreate the configuration file. I think it would be easier than redeploy and recreate the integrations.

src/synapse.py
amandahla: There is no database integration yet. This will be addressed in a future story/PR.

src/charm.py
amandahla: I'll close this one in favor of https://github.com/canonical/synapse-operator/pull/8#discussion_r1245192606 so we can discuss there.

tests/integration/test_charm.py
amandahla: Oops, you're right, thanks. Changed.

src/charm.py
gregory-schiano: This is very very strange.
Normally each hook is executed in a new instance BUT I read that this is not guaranteed (but never good clear answers on when this could happen)
Is it systematic ?

src/charm.py
gregory-schiano: Well if you add the confirm parameter, that warns you, and you didn't want to execute this we can't do much for you.
But I still think it's a safeguard that avoids launching the action by mistake, but I agree that ideally it should ask for confirmation before execution but that's no possible in juju

src/charm.py
amandahla: Hmmm the deployed charm works as expected (juju config -> init -> config-changed) but the unit tests using "harness.update_config({"server_name": server_name_changed})" dont call init again.

amandahla: https://github.com/canonical/operator/issues/736

gregory-schiano: Ah indeed, that's something we realised with Weii too while working on Flask charm

amandahla: I added a new fixture called harness_server_name_changed_fixture that will force the charm to be initialized before the config-chaged event :)

gtrkiller (approved):
LGTM

yanksyoon (approved):
LGTM!

weiiwang01 (approved):

tests/unit/test_charm.py
amandahla: I agree but I'll add that as a note on the stories regarding actions since there Synapse module will have more features and this will be more clear (isd-786 and 787). This will be addressed there. Thanks.

PR #13, opened 2023-07-04
feat(ISD-843): Add Postgresql integration

amandahla:
Add Postgresql integration


src/charm.py
```diff
@@ -140,6 +150,24 @@ def _on_reset_instance_action(self, event: ActionEvent) -> None:
...
+        if not self.unit.is_leader():
+            event.defer()
+            return
+
+        self.model.unit.status = ops.MaintenanceStatus("Preparing the database")
+        try:
+            self._database.prepare_database()
+            self._on_config_changed(event)
+        except psycopg2.Error as exc:
+            self.model.unit.status = ops.BlockedStatus(str(exc))
```
gregory-schiano, line 168: Can the user act in this situation ? If so can we put a more explicit message of what we expect them to do, if not I think we should put it in ErrorStatus

src/charm.py
```diff
@@ -140,6 +150,24 @@ def _on_reset_instance_action(self, event: ActionEvent) -> None:
...
+            event: Event triggering the database created handler.
+        """
+        if not self.unit.is_leader():
+            event.defer()
+            return
+
+        self.model.unit.status = ops.MaintenanceStatus("Preparing the database")
+        try:
+            self._database.prepare_database()
+            self._on_config_changed(event)
```
gregory-schiano, line 166: I think the business logic should be moved to a dedicated method that both hook method will call instead of calling the hook method directly

src/charm.py
```diff
@@ -140,6 +150,24 @@ def _on_reset_instance_action(self, event: ActionEvent) -> None:
...
         event.set_results(results)  # type: ignore[arg-type]
 
+    def _on_database_created(self, event: DatabaseCreatedEvent) -> None:
+        """Handle database created.
+
+        Args:
+            event: Event triggering the database created handler.
+        """
+        if not self.unit.is_leader():
+            event.defer()
```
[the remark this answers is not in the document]
amandahla: Good question. Now I think it's not relevant. Removed.

src/charm.py
amandahla: Depending on the error, it might have some action to do on the database side, that's why I added it as blocked status

src/charm.py
```diff
@@ -43,7 +46,13 @@ def __init__(self, *args: Any) -> None:
         except CharmConfigInvalidError as exc:
             self.model.unit.status = ops.BlockedStatus(exc.msg)
             return
-        self._synapse = Synapse(charm_state=self._charm_state)
+        self._database = DatabaseObserver(self)
+        self.framework.observe(
+            self._database.database.on.database_created, self._on_database_created
+        )
```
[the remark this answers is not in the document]
amandahla: Nice, added, thanks.

src/charm.py
amandahla: Hm...When there is a database relation, the configuration file should be updated considering the database information like a configuration change would do. Unless I change to something like:

```
def update_config...

def _config_changed:
    change_config()

def _on_database_created:
    ...
    update_config()
   ...
```
Is that what you mean?

requirements.txt
```diff
@@ -1,4 +1,4 @@
-jsonschema >= 4.17
+jsonschema >=4.10.0,<4.18.0
```
jdkandersson, line 1: If we're pinning all the way to patch, do you think we should just pick a specific version?

src/charm.py
```diff
@@ -43,7 +46,16 @@ def __init__(self, *args: Any) -> None:
...
+        self.framework.observe(
+            self._database.database.on.endpoints_changed, self._on_config_changed
```
jdkandersson, line 54: The guidance is to have one event handler per event with a name that matches the event. Events can share functionality from within those event handlers by calling underlying functions

src/charm.py
```diff
@@ -43,7 +46,16 @@ def __init__(self, *args: Any) -> None:
...
+        )
+        self._synapse = Synapse(
+            charm_state=self._charm_state, database_data=self._database.get_relation_data()
```
jdkandersson, line 57: Should the database relation data be passed via state?

src/charm.py
```diff
@@ -43,7 +46,16 @@ def __init__(self, *args: Any) -> None:
         except CharmConfigInvalidError as exc:
             self.model.unit.status = ops.BlockedStatus(exc.msg)
             return
-        self._synapse = Synapse(charm_state=self._charm_state)
+        self._database = DatabaseObserver(self)
+        self.framework.observe(
+            self._database.database.on.database_created, self._on_database_created
+        )
+        self.framework.observe(
```
jdkandersson, line 53: Any thoughts on the observer pattern for the database?

src/charm.py
```diff
@@ -127,19 +150,48 @@ def _on_reset_instance_action(self, event: ActionEvent) -> None:
         if not container.can_connect():
             event.fail("Failed to connect to container")
             return
-        self.model.unit.status = ops.MaintenanceStatus("Resetting Synapse instance")
+        logger.info("Replan service to not restart")
```
jdkandersson, line 153: Perhaps it is worth explaining why the restart is not needed?

src/charm.py
```diff
@@ -127,19 +150,48 @@ def _on_reset_instance_action(self, event: ActionEvent) -> None:
...
+            SYNAPSE_CONTAINER_NAME, self._pebble_layer_without_restart, combine=True
+        )
+        container.replan()
         try:
+            self.model.unit.status = ops.MaintenanceStatus("Stop Synapse instance")
+            logger.info("Stop Synapse instance")
+            container.stop(SYNAPSE_SERVICE_NAME)
+            self.model.unit.status = ops.MaintenanceStatus("Erase Synapse data")
             self._synapse.reset_instance(container)
+            if self._database.get_relation_data() is not None:
```
jdkandersson, line 164: Should this be from the charm state?

src/charm.py
jdkandersson: Yes I suspect that is what Gregory means. Event handler function should only be used to handle events and the code suggested above is a way of sharing the business logic between events

src/charm.py
jdkandersson: This is quite an interesting case actually - it isn't clear whether the charm should be in blocked or error state. It also doesn't seem to be reasonable for the charm to have logic to figure out whats wrong since this is more on the database side.

Perhaps it would be good to have something in the database library from the data platform team where you can send an exception and set the charm state based on that? Any thoughts @benhoyt

src/database.py
```diff
@@ -0,0 +1,144 @@
...
+        self._charm = charm
+        # SUPERUSER is required to update pg_database
+        self.database = DatabaseRequires(
+            self._charm,
+            relation_name=self._RELATION_NAME,
+            database_name=self._charm.app.name,
+            extra_user_roles="SUPERUSER",
+        )
+
+    def get_relation_data(self) -> typing.Optional[typing.Dict]:
```
jdkandersson, line 39: Similar to earlier comment, should this be done via the charm state?

src/database.py
```diff
@@ -0,0 +1,144 @@
...
+                logger.error("Failed to prepare database: %s", str(exc))
+                raise
+
+    def erase_database(self) -> None:
+        """Erase database.
+
+        Raises:
+           Error: something went wrong while erasing the database.
+        """
+        conn = self.get_conn("template1")
```
jdkandersson, line 88: What is `template1` here please?

src/database.py
```diff
@@ -0,0 +1,144 @@
...
+        database_name = self.get_database_name()
+        if conn is not None:
```
jdkandersson, line 90: What if it is `None`?

src/database.py
```diff
@@ -0,0 +1,144 @@
...
+        """
+        conn = self.get_conn()
+        database_name = self.get_database_name()
+        if conn is not None:
+            try:
+                with conn.cursor() as curs:
+                    curs.execute(
+                        "UPDATE pg_database "
+                        "SET datcollate='C', datctype='C' "
+                        "WHERE datname=%s",
```
jdkandersson, line 75: Is this vulnerable to a SQL injection attack? I think usually these kind of things are handled as variables in SQL rather than injecting directly into the SQL statement

jdkandersson: Also, What is the reason for picking `%s` compared to f-strings?

src/database.py
```diff
@@ -0,0 +1,144 @@
...
+        """
+        conn = self.get_conn("template1")
+        database_name = self.get_database_name()
+        if conn is not None:
+            try:
+                with conn.cursor() as curs:
+                    curs.execute(sql.SQL("DROP DATABASE {}").format(sql.Identifier(database_name)))
+                    curs.execute(
+                        sql.SQL(
+                            "CREATE DATABASE {} "
```
jdkandersson, line 96: Similar to above, is this vulnerable to SQL injection?

src/database.py
```diff
@@ -0,0 +1,144 @@
...
+    def get_database_name(self) -> str:
+        """Get database name from relation.
+
+        Returns:
+            database name
+        """
+        relation_data = self.get_relation_data()
+        if relation_data is not None:
+            return relation_data.get("POSTGRES_DB", "")
+        return ""
```
jdkandersson, line 144: Would it be confusing to get an empty string if there is no database configured? Why not `None`?

src/synapse.py
```diff
@@ -30,13 +30,15 @@
 class Synapse:
     """A class representing the Synapse application."""
 
-    def __init__(self, charm_state: CharmState):
+    def __init__(self, charm_state: CharmState, database_data: typing.Optional[typing.Dict]):
```
jdkandersson, line 33: At this stage should the database data be in the state?

src/synapse.py
```diff
@@ -57,13 +59,20 @@ def synapse_environment(self) -> typing.Dict[str, str]:
...
-        return {
+        environment = {
             "SYNAPSE_SERVER_NAME": f"{self._charm_state.server_name}",
             "SYNAPSE_REPORT_STATS": f"{self._charm_state.report_stats}",
             # TLS disabled so the listener is HTTP. HTTPS will be handled by Traefik.
             # TODO verify support to HTTPS backend before changing this  # pylint: disable=fixme
             "SYNAPSE_NO_TLS": str(True),
         }
+        if self._database_data is not None:
+            environment["POSTGRES_DB"] = self._database_data.get("POSTGRES_DB", "")
```
jdkandersson, line 70: Would this work? https://github.com/canonical/synapse-operator/pull/13/files#r1253285244

src/charm.py
benhoyt: Yeah, I think Gregory's intuition is correct -- this should probably be error status. The error is basically "some problem with your database". However, instead of explicitly setting an error status, I'd just remove the try/except and let it fail. Juju will automatically put it in error status with a generic message (and the stack track will be in `juju debug-log`), and Juju will automatically retry.

It seems like we're catching any database error here, so we'd have no idea what message to put in a `BlockedStatus` anyway. It's not blocked (waiting for a config value or other user action), it's error (we have no idea what went wrong).

@amandahla What kind of errors were you seeing that you put a try/except here? Or was it just a theoretical concern?

amandahla: I didn't test but I was considering the scenario when a backup is being created and would block new connections. So the action would be to check the backup.
Or if for some reason the user has no permission to update the pg_database table then the action would be to check the table permissions or user roles.
Or even if for some reason the relation doesn't exist or it was removed.

I didn't think about ErrorStatus because of this "Error state that is never set from Charm code. " in [Constructs](https://juju.is/docs/sdk/constructs).

Anyway, IMHO it's a matter of interpretation about what is a problem that might need an action so I won't mind changing this for ErrorStatus.

src/charm.py
amandahla: It was my first try but then I couldn't solve how to call the config-changed method from the Observer class. That's why I moved back to the charm.py. Any ideas on how could I do that from the Observer class?

src/database.py
amandahla: Now that you asked...Since this is an action that needs a relation to proceed, I believe it should put the charm in blocked status if is None?
Per what is written "or the charm requires a relation with another application which has not yet been defined" in [Constructs](https://juju.is/docs/sdk/constructs).

src/database.py
amandahla: Because this is the way recommended by the library documentation in "[Passing parameters to SQL queries](https://www.psycopg.org/docs/usage.html#query-parameters)".

I found out this because I tried to use f-string and this was the result:
```>> Issue: [B608:hardcoded_sql_expressions] Possible SQL injection vector through string-based query construction.
  Severity: Medium   Confidence: Medium
  CWE: CWE-89 (https://cwe.mitre.org/data/definitions/89.html)
  More Info: https://bandit.readthedocs.io/en/1.7.5/plugins/b608_hardcoded_sql_expressions.html
  Location: /home/ubuntu/github-runner/_work/synapse-operator/synapse-operator/src/database.py:78:24
77                        curs.execute(
78                            "UPDATE pg_database SET datcollate='C', datctype='C' "
79                            f"WHERE datname='{database_name}' AND datcollate='en_US.UTF-8' "
80                            "AND datctype='en_US.UTF-8';"
81                        )
```

src/database.py
amandahla: This is another way of doing the same thing as the UPDATE query, not vulnerable to SQL injection. I'll change the previous one to follow the same pattern.

src/database.py
amandahla: I'll change this in favour of sql.SQL and sql.Identifier.

src/database.py
amandahla: I think this is the same case that you asked before regarding the connection. Instead of returning None/empty, would be better to block the charm since a relation is being required and is not available?

src/synapse.py
amandahla: It seems that we have many cases where the point is what to do if something from a relation is required but is not available. Should block the charm?

src/charm.py
amandahla: Thanks, changed.

requirements.txt
amandahla: Agree, changed to use specific version.

src/charm.py
amandahla: My bad, changed to use same name as the event.

src/charm.py
amandahla: Sure, changed.

src/charm.py
amandahla: Good point, I added a comment about it.

src/charm.py
amandahla: Yes, changed.

src/database.py
amandahla: Yes, changed.

src/database.py
amandahla: template1 is a database provided by Postgresql used as template to create new database. The charm will connect on this database instead of the one used by the relation to be able to drop and recreate it since is not possible to do that while connected to the database. I added a comment explaining that, thanks for pointing that out.

src/database.py
amandahla: Changed.

src/database.py
amandahla: Now all queries are being constructed via sql.SQL

src/synapse.py
amandahla: Yes, changed.

src/database.py
amandahla: I'll mark this one as resolved so we can discuss this at the last comment, ok? To make it easier to track.

src/database.py
amandahla: I'll mark this one as resolved so we can discuss this at the last comment, ok? To make it easier to track.

src/charm.py
benhoyt: Yeah, you wouldn't set `ErrorStatus` explicitly, you'd just remove the `try/except` altogether and let the error bubble up to Juju, which would set an error status.

src/database.py
```diff
@@ -0,0 +1,147 @@
...
+                    database_name = self.get_database_name()
+                conn = psycopg2.connect(
+                    f"dbname='{database_name}' user='{user}' host='{host}'"
+                    f"password='{password}' connect_timeout=1"
+                )
+                conn.autocommit = True
+                return conn
+            except psycopg2.Error as exc:
+                logger.error("Failed to connect to database: %s", str(exc))
+                raise
```
weiiwang01, line 135: I think `logger.exception` is better in this scenario, since it will print the trackback and full exception information.

amandahla: Nice, thanks!

src/charm.py
amandahla: Ok, try/except removed, thanks.

src/charm.py
jdkandersson: @yanksyoon?

src/synapse.py
jdkandersson: yeh probably blocking the charm or error state since the operator probably can't do much at that point?

src/charm_state.py
```diff
@@ -102,12 +103,24 @@ def report_stats(self) -> typing.Union[str, bool, None]:
...
         return self._synapse_config.report_stats
 
+    @property
+    def database_data(self) -> typing.Optional[typing.Dict]:
+        """Return database data from database relation.
+
+        Returns:
+            Dict: Information needed for setting connection with database.
+        """
+        return self._database_data
```
jdkandersson, line 113: Should this be exposed in a more manageable form? E.g., as another dataclass or perhaps a typed dictionary? That way we get the advantages of type hints etc? What about pydantic validation of the database data?

src/charm.py
amandahla: Considering the discussion [here](https://github.com/canonical/flask-k8s-operator/pull/14#discussion_r1203524175) in Discourse and the example in [Flask](https://github.com/canonical/flask-k8s-operator/blob/main/src/databases.py#L59), the solution would replan the charm in the Observer but in this case, there is a call to migrate_config (a command to regenerate the command) so I'm afraid that I would need to replicate the config-change code on the Observer.

IMHO I don't see much gain in doing that. Couldn't we keep the observer in the charm and apply the Observer pattern in cases where there is no need for layer replan?

jdkandersson: @gregory-schiano probably one for you to take a look at

src/charm_state.py
```diff
@@ -16,6 +16,7 @@
     validator,
 )
 
+from database import ConnectionParams
```
jdkandersson, line 19: Perhaps this class should be defined here? Otherwise the state module depends on the database module and the database module depends on the state module?

src/charm.py
jdkandersson: And what would the circular import be please?

amandahla: I thought that would be in my previous commits but it wasn't so I created this branch to show what I tried:
https://github.com/canonical/synapse-operator/blob/postgresql-observer-pattern/src/database.py#L62

src/synapse.py
amandahla: Exactly. I'm considering leaving it as ErrorStatus because the relation with Postgresql is not required and maybe the BlockedStatus would give this impression.

src/charm_state.py
amandahla: Working on that :) thanks

src/charm.py
amandahla: @yanksyoon Thanks for the suggestion. I applied the idea of PebbleService here then I could remove the observer from the charm.

src/charm_state.py
amandahla: After the suggestion to use PebbleService, I changed the code a lot. Please take a look and feel free to reopen this comment or add new ones :)

jdkandersson (dismissed):

src/charm.py
```diff
@@ -145,18 +121,23 @@ def _on_reset_instance_action(self, event: ActionEvent) -> None:
...
             results["reset-instance"] = True
-        except (ops.pebble.PathError, CommandMigrateConfigError) as exc:
+        except (psycopg2.Error, ops.pebble.PathError, CommandMigrateConfigError) as exc:
             self.model.unit.status = ops.BlockedStatus(str(exc))
             event.fail(str(exc))
             return
-        self.model.unit.status = ops.ActiveStatus()
         # results is a dict and set_results expects _SerializedData
         event.set_results(results)  # type: ignore[arg-type]
+        self.model.unit.status = ops.ActiveStatus()
```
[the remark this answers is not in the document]
amandahla: As we discussed on MM, I moved erase_database and prepare_database to database_client (erase and prepare now). So now we have something like this:
```
        datasource = self.get_relation_as_datasource()
        db_client = DatabaseClient(datasource=datasource)
        db_client.prepare()
```

src/database_observer.py
```diff
@@ -0,0 +1,137 @@
...
+        self.framework.observe(self.database.on.endpoints_changed, self._on_endpoints_changed)
+
+    @property
+    def _pebble_service(self) -> typing.Any:
+        """Return instance of pebble service.
+
+        Returns:
+            instance of pebble service or none.
+        """
+        return getattr(self._charm, "pebble_service", None)
```
yanksyoon, line 59: Out of curiosity, would it be correct that this cannot be injected due to circular dependencies?

tests/integration/conftest.py
```diff
@@ -73,7 +79,10 @@ async def synapse_app_fixture(
         series="jammy",
         config={"server_name": server_name},
     )
-    await model.wait_for_idle(raise_on_blocked=True)
+    async with ops_test.fast_forward():
```
yanksyoon, line 82: Nice!

yanksyoon (approved):
Minor questions, LGTM, learned a few things here and there!

src/charm.py
```diff
@@ -66,20 +66,20 @@ def __init__(self, *args: typing.Any) -> None:
...
+    def change_config(self, event: ops.HookEvent) -> None:
+        """Change configuration.
 
         Args:
-            event: Event triggering the need of changing the configuration.
+            event: Event triggering after config needs to be changed.
         """
         container = self.unit.get_container(SYNAPSE_CONTAINER_NAME)
         if not container.can_connect():
             event.defer()
```
arturo-seijas, line 77: I've header we should avoid deferring events. DK if there's any alternative here

src/database_observer.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def _change_config(self, event: typing.Any) -> None:
+        """Change the configuration.
+
+        Args:
+            event: Event triggering the database created or endpoints changed handler.
+        """
+        container = self._charm.unit.get_container(SYNAPSE_CONTAINER_NAME)
+        if not container.can_connect() or self._pebble_service is None:
+            event.defer()
```
arturo-seijas, line 69: Same here

src/database_observer.py
```diff
@@ -0,0 +1,137 @@
...
+        return getattr(self._charm, "pebble_service", None)
+
+    def _change_config(self, event: typing.Any) -> None:
+        """Change the configuration.
+
+        Args:
+            event: Event triggering the database created or endpoints changed handler.
+        """
+        container = self._charm.unit.get_container(SYNAPSE_CONTAINER_NAME)
+        if not container.can_connect() or self._pebble_service is None:
```
arturo-seijas, line 68: Do you need `not container.can_connect()`? This condition is already in `self._pebble_service.change_config`

arturo-seijas, line 68: From what I can see, `self._pebble_service is None` is never None

arturo-seijas: This will also mean getting read of `_pebble_service` in this class

tests/integration/conftest.py
arturo-seijas: Pardon my ignorance but, what's this for?

src/charm_types.py
```diff
@@ -20,3 +20,21 @@ class ExecResult(typing.NamedTuple):
     exit_code: int
     stdout: str
     stderr: str
+
+
+class DatasourcePostgreSQL(typing.TypedDict):
```
arturo-seijas, line 25: There's only one addtional class here that doesn't seem to be used anymore. If that's the case it might make sense to remove it an rename this file to datasource.py or something similar

arturo-seijas, line 25: I'd move the `ExecResult` class to synapse.py and rename this file to datasource.py or something similar. I'm not too keen on having a file for generic stuff

src/database_client.py
```diff
@@ -0,0 +1,117 @@
...
+
+        Args:
+            datasource: datasource to use to connect.
+            alternative_database: database to connect to.
+                The default is to use the one provided by datasource.
+
+        Raises:
+            CharmDatabaseRelationNotFoundError: if there is no relation.
+        """
+        if datasource is None:
```
arturo-seijas, line 34: I don't see any flow where this is None

src/exceptions.py
```diff
@@ -38,6 +38,22 @@ def __init__(self, msg: str):
         self.msg = msg
 
 
+class CharmDatabaseRelationNotFoundError(Exception):
```
arturo-seijas, line 41: In line with the comment above, I think `CommandMigrateConfigError` and `ServerNameModifiedError` belong in synapse.py as they are part of the bussiness logic this file encapsultes.
`CharmConfigInvalidError` would probably belong in charm_state.py

src/charm.py
amandahla: You're right. I found the [discussion](https://github.com/canonical/operator/issues/935#issue-1722066396) that @gregory-schiano shared on MM and it looks like event.defer() might generate a situation where you can have a queue with events waiting to be addressed and then, when the condition that the charm needs to continue is true, suddenly this could cause a lot of restarts, for example.
Unfortunately, I didn't find many examples of how to prevent that except for this one on [Alertmanager](https://github.com/canonical/alertmanager-k8s-operator/blob/main/src/charm.py#L380). It looks like a nice way of handling this so I'll do the same here.

src/database_client.py
```diff
@@ -0,0 +1,117 @@
...
+            self._connect()
+            with self._conn.cursor() as curs:
+                curs.execute(
+                    sql.SQL("DROP DATABASE {}").format(sql.Identifier(self._database_name))
+                )
+                curs.execute(
+                    sql.SQL(
+                        "CREATE DATABASE {} "
+                        "WITH LC_CTYPE = 'C' LC_COLLATE='C' TEMPLATE='template0';"
+                    ).format(sql.Identifier(self._database_name))
```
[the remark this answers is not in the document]
amandahla: Do you mean to use the format "cur.execute(sql_query, (user_input,))" instead?

This can be used too but I prefer to use the [module one](https://www.psycopg.org/docs/sql.html#module-usage) as is it used by the Data team [here](https://github.com/canonical/postgresql-k8s-operator/blob/main/lib/charms/postgresql_k8s/v0/postgresql.py#L114) too.

src/exceptions.py
amandahla: Nice, changed.

tests/integration/conftest.py
yanksyoon: This speeds up the `update-status` interval to 10s :D
```
Temporarily speed up update-status firing rate for the current model.
```

src/database_client.py
arturo-seijas: Oh, I see

src/database_client.py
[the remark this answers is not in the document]
arturo-seijas: Ok! Thanks for the explanation

arturo-seijas (approved):
LGTM

yanksyoon (approved):

PR #15, opened 2023-07-18
Add action to register user

amandahla:
Add action to register user.

At first, I thought that would need to create an admin user via command line and then use it to create via API but then I found out this documentation [Register API](https://matrix-org.github.io/synapse/v1.41/admin_api/register_api.html) that make it possible to create user using a shared secret (defined in the configuration file) and a nonce.
This makes things much easier now.


actions.yaml
```diff
@@ -6,3 +6,26 @@ reset-instance:
...
+    Registers a user for the Synapse server.
+    You need to supply a user name and user password
+    and whether that user should be an admin or not.
+  properties:
+    username:
+      description: |
+        When not using SSO, a user name is needed
+        for the creation of a matrix account.
+      type: string
+    password:
```
jdkandersson, line 20: Should we auto-generate the password?

metadata.yaml
```diff
@@ -46,3 +46,7 @@ requires:
         interface: nginx-route
         limit: 1
         optional: true
+
+peers:
```
jdkandersson, line 50: What is the reason for needing this please?

src/charm.py
```diff
@@ -41,8 +42,9 @@ def __init__(self, *args: typing.Any) -> None:
         except CharmConfigInvalidError as exc:
             self.model.unit.status = ops.BlockedStatus(exc.msg)
             return
-        self._synapse = Synapse(charm_state=self._charm_state)
-        self.pebble_service = PebbleService(synapse=self._synapse)
+        self.synapse = Synapse(charm_state=self._charm_state)
```
jdkandersson, line 45: What is the reason for making this public please?

src/charm.py
```diff
@@ -138,6 +141,27 @@ def _on_reset_instance_action(self, event: ActionEvent) -> None:
...
+
+        Args:
+            event: Event triggering the reset instance action.
+        """
+        results = {
+            "register-user": False,
+        }
+        username = event.params["username"]
+        password = event.params["password"]
+        admin = bool(event.params["admin"] == "yes")
```
jdkandersson, line 155: Should this be done using pydantic?

src/charm.py
```diff
@@ -138,6 +141,27 @@ def _on_reset_instance_action(self, event: ActionEvent) -> None:
...
+        try:
+            self._synapse_api.register_user(username=username, password=password, admin=admin)
+            results["register-user"] = True
```
jdkandersson, line 158: Should this be together with `event.set_results`

src/charm.py
```diff
@@ -138,6 +141,27 @@ def _on_reset_instance_action(self, event: ActionEvent) -> None:
...
+        except (RegisterUserError, NetworkError) as exc:
+            event.fail(str(exc))
+            return
+        # results is a dict and set_results expects _SerializedData
```
jdkandersson, line 162: If it expects something else, why are we passing `dict`?

src/synapse.py
```diff
@@ -146,51 +146,70 @@ def execute_migrate_config(self, container: ops.Container) -> None:
                 "Migrate config failed, please review your charm configuration"
             )
 
-    def check_server_name(self, container: ops.Container) -> None:
-        """Check server_name.
-
-        Check if server_name of the state has been modified in relation to the configuration file.
+    def get_configuration_field(
```
jdkandersson, line 149: Could this be a private function?

src/synapse_api.py
```diff
@@ -0,0 +1,178 @@
...
+import ops
+import requests
+from ops.charm import CharmBase
+
+from constants import SYNAPSE_CONTAINER_NAME
+
+logger = logging.getLogger(__name__)
+
+SYNAPSE_URL = "http://localhost:8008"
+URL_REGISTER = f"{SYNAPSE_URL}/_synapse/admin/v1/register"
```
jdkandersson, line 24: Should this be called `REGISTER_URL`?

src/synapse_api.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+class RegisterUserError(Exception):
```
jdkandersson, line 27: Should this inherit from a base exception within this module?

src/synapse_api.py
```diff
@@ -0,0 +1,178 @@
...
+        Args:
+            msg (str): Explanation of the error.
+        """
+        self.msg = msg
+
+
+class SynapseAPI:
+    """The Synapse API handler."""
+
+    def __init__(self, charm: CharmBase):
```
jdkandersson, line 62: What is the reason for needing a class here please?

jdkandersson, line 62: What is the reason for taking the charm as an input here please?

src/synapse_api.py
```diff
@@ -0,0 +1,178 @@
...
+        Args:
+            shared_secret: registration_shared_secret from configuration file.
+            nonce: nonce generated.
+            user: username.
+            password: password.
+            admin: if is admin. Default False.
+            user_type: user type. Defaults to None.
+
+        Returns:
+            _type_: _description_
```
jdkandersson, line 111: This doesn't look right?

src/synapse_api.py
```diff
@@ -0,0 +1,178 @@
...
+        mac.update(user.encode("utf8"))
+        mac.update(b"\x00")
+        mac.update(password.encode("utf8"))
+        mac.update(b"\x00")
+        mac.update(b"admin" if admin else b"notadmin")
+        if user_type:
+            mac.update(b"\x00")
+            mac.update(user_type.encode("utf8"))
+
+        return mac.hexdigest()
```
jdkandersson, line 125: Is there a particular reason for the order and adding `b"\x00"` in between?

src/synapse_api.py
```diff
@@ -0,0 +1,178 @@
...
+            username: name to be registered.
+            password: user's password.
+            admin: if the user is admin or not.
+
+        Raises:
+            RegisterUserError: when registering user via API fails.
+            NetworkError: if there was an error registering the user.
+        """
+        # get registration_shared_secret from config file
+        container = self._charm.unit.get_container(SYNAPSE_CONTAINER_NAME)
```
jdkandersson, line 140: Couldn't this just take the container as an argument?

src/synapse_api.py
```diff
@@ -0,0 +1,178 @@
...
+        if not container.can_connect():
```
jdkandersson, line 141: Should this check be done on the charm?

src/synapse_api.py
```diff
@@ -0,0 +1,178 @@
...
+            self._charm.unit.status = ops.MaintenanceStatus("Waiting for pebble")
```
jdkandersson, line 142: This definitely should not be done here, these kind of modules should not be interacting with the charm outside of taking the container as an argument to (1) interact with the file system and (2) execute commands if there isn't an API available

src/synapse_api.py
```diff
@@ -0,0 +1,178 @@
...
+            return
+        # Synapse is defined in the charm
+        synapse = self._charm.synapse  # type: ignore[attr-defined]
+        registration_shared_secret = synapse.get_configuration_field(
```
jdkandersson, line 146: Should this be taken as an argument as well?

jdkandersson (commented):
Have not reviewed the tests, the `synapse_api` module will require architectural changes

actions.yaml
```diff
@@ -6,3 +6,26 @@ reset-instance:
...
     This actions will erase all data and create a instance with the new server_name.
+register-user:
+  description: |
+    Registers a user for the Synapse server.
+    You need to supply a user name and user password
+    and whether that user should be an admin or not.
+  properties:
+    username:
+      description: |
+        When not using SSO, a user name is needed
```
arturo-seijas, line 17: What happens when SSO is enabled?

actions.yaml
```diff
@@ -6,3 +6,26 @@ reset-instance:
...
+        for the creation of a matrix account.
+      type: string
+    password:
+      description: |
+        Specifies a user password that relates to a user name
```
arturo-seijas, line 22: "Password for the user account" seems enough description to me :)

actions.yaml
amandahla: Good point. I don't see why not, thanks. I'll change this.

metadata.yaml
amandahla: My bad, this should be removed.

src/charm.py
amandahla: I changed it to public so synapse_api could use it through the charm instance. It is needed to get the field registration_shared_secret from the configuration file.

src/charm.py
amandahla: No problem, changed.

src/charm.py
amandahla: Could you give me an example, please?
Do you mean something like this?
```
event.set_results({register-user": True, "user-password": password})
```

src/charm.py
amandahla: @yhaliaw can answer that better than me :)
See: https://github.com/canonical/github-runner-operator/pull/26#discussion_r1049138401
I could do as he did and change it to a Dict[str, "JsonObject"] but in this case, dict worked fine.

src/synapse.py
amandahla: Actually no because Synapse API needs it to get the registration_shared_secret

src/synapse_api.py
amandahla: This way register_user and future methods wouldn't need charm or synapse as parameters. Also makes it easier to define some methods related only to the API private. I believe there is a similar behavior when another charms as Prometheus and Grafana.

amandahla: To register a user, the charm needs a synapse instance to read the registration_shared_secret from the configuration file.
For this, there is a check to guarantee that is possible to connect to the container and then get this field.

src/synapse_api.py
amandahla: This is what the [documentation](https://matrix-org.github.io/synapse/v1.41/admin_api/register_api.html) suggests to do. The same happens in the [script](https://github.com/matrix-org/synapse/blob/develop/synapse/_scripts/register_new_matrix_user.py#L68).

src/synapse_api.py
amandahla: My bad, changed.

src/charm.py
amandahla: Reverted to private after changing Synapse API to a module.

src/synapse_api.py
amandahla: I moved this to the charm so the API will receive the registration shared secret as a parameter.

src/synapse_api.py
amandahla: Agree!

src/synapse_api.py
amandahla: Moved to the charm.

src/synapse_api.py
amandahla: Yes, now synapse_api gets the registration_shared_secret as a parameter.

src/charm.py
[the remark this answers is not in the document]
amandahla: This will be useful for testing before we have the integration with SAML.

actions.yaml
amandahla: There is an option where you can tell Synapse to allow both authentications : [localdb_enabled](https://matrix-org.github.io/synapse/latest/usage/configuration/config_documentation.html#password_config). 
When adding SSO/SAML integration we can discuss if it must be a config for changing that.

actions.yaml
amandahla: Agree :laughing: David suggested that the action could create the password so this field was removed.

src/charm.py
jdkandersson: As in, this line is within the `try ... except` but `event.set_result` isn't, shouldn't both be either in the `try ... except` or both out?

src/charm.py
jdkandersson: @benhoyt any comment please?

jdkandersson: in general, using `type: ignore` isn't great so it would be better not to use it

src/synapse.py
jdkandersson: ok, then should this method be on the `Synapse` class? It could just take the container as an argument and then we have less coupling between Synapse and Synapse API

src/synapse_api.py
jdkandersson: ok, do you think it should be documented with links here? I'm guessing a future dev would be as confused as I am 🙂

src/charm.py
```diff
@@ -141,22 +143,46 @@ def _on_reset_instance_action(self, event: ActionEvent) -> None:
         event.set_results(results)  # type: ignore[arg-type]
         self.model.unit.status = ops.ActiveStatus()
 
+    def _get_random_password(self) -> str:
```
jdkandersson, line 146: What is the reason for putting this on the charm class please? Perhaps a separate module?

src/charm.py
```diff
@@ -141,22 +143,46 @@ def _on_reset_instance_action(self, event: ActionEvent) -> None:
...
+        results = {"register-user": False, "user-password": ""}
+        container = self.unit.get_container(SYNAPSE_CONTAINER_NAME)
+        if not container.can_connect():
+            self.unit.status = ops.MaintenanceStatus("Waiting for pebble")
+            return
+        registration_shared_secret = self._synapse.get_configuration_field(
+            container=container, fieldname="registration_shared_secret"
+        )
+        if registration_shared_secret is None:
+            event.fail("registration_shared_secret was not found")
```
jdkandersson, line 171: Should this include some advice to the operator on what to do in that case?

src/charm.py
```diff
@@ -141,22 +143,46 @@ def _on_reset_instance_action(self, event: ActionEvent) -> None:
...
+            return
+        user_data = {
```
jdkandersson, line 173: This doesn't feel like it is in the right place, @gregory-schiano? Perhaps it should go into a separate user management module or the Synapse API can take care of it including the password generation?

src/charm.py
```diff
@@ -141,22 +143,46 @@ def _on_reset_instance_action(self, event: ActionEvent) -> None:
...
+            "username": event.params["username"],
+            "admin": event.params["admin"],
+            "password": self._get_random_password(),
         }
-        username = event.params["username"]
-        password = event.params["password"]
-        admin = bool(event.params["admin"] == "yes")
+        user = User(**user_data)
```
jdkandersson, line 178: What if there is a validation error? I'm guessing we're using pydantic here for validation

src/synapse_api.py
```diff
@@ -40,139 +38,109 @@ def __init__(self, msg: str):
...
+    admin: bool = False,
+    user_type: typing.Optional[str] = None,
+) -> str:
+    """Generate mac to register user.
+
+    Args:
+        shared_secret: registration_shared_secret from configuration file.
+        nonce: nonce generated.
+        user: username.
+        password: password.
```
jdkandersson, line 85: Perhaps expand on this a bit? Describing the argument using the name of the argument doesn't add much value

src/user.py
jdkandersson, line 1: What about putting the password generation in here?

jdkandersson (commented):
Getting closer

src/charm.py
amandahla: Oh, I get it, you're right, it doesn't make sense to be inside the try. I'll move it out.

src/synapse.py
amandahla: I have an idea: I created a public method called get_registration_shared_secret that will be called in the charm only. Then _get_configuration_field is private now.

src/synapse_api.py
amandahla: Sure!! Nice suggestion :)

src/charm.py
amandahla: I would suggest to check the logs. I'll add this to the message.

src/charm.py
amandahla: Update: I moved the password generation to User class.

src/user.py
amandahla: Perfect, changed.

amandahla: (I marked as resolved, but I'll push the commit with this change in a few minutes)

src/charm.py
amandahla: Ops, sure, changed!

src/charm.py
amandahla: Moved to User class.

src/charm.py
amandahla: Update 1: all interactions with Synapse were moved to synapse module as suggested by @gregory-schiano thanks :)

src/charm.py
benhoyt: I think you just need to update your version of ops. I improve a bunch of type annotations in 2.4.0, so the type of `set_results` is now `set_results(self, results: Dict[str, Any])` instead of using the (private) `_SerializedData`. So you should be able to update ops (I'd use `ops >= 2.4.1`) and then remove the `type: ignore` and it'll work fine.

src/charm.py
```diff
@@ -81,9 +80,9 @@ def change_config(self, _: ops.HookEvent) -> None:
             self.pebble_service.change_config(container)
         except (
             CharmConfigInvalidError,
```
jdkandersson, line 82: Perhaps these could be grouped by a parent exception? Exception granularity is useful if we do something different for each one, a long list like this becomes more difficult to maintain and reason about (e.g., which one is and isn't covered)

src/charm.py
```diff
@@ -131,60 +130,43 @@ def _on_reset_instance_action(self, event: ActionEvent) -> None:
                 # Otherwise PostgreSQL will prevent it if there are open connections.
                 db_client = DatabaseClient(datasource=datasource, alternative_database="template1")
                 db_client.erase()
-            self._synapse.execute_migrate_config(container)
+            synapse.execute_migrate_config(container=container, charm_state=self._charm_state)
```
jdkandersson, line 133: This is a good change 👍

src/charm.py
```diff
@@ -131,60 +130,43 @@ def _on_reset_instance_action(self, event: ActionEvent) -> None:
...
         }
-        user = User(**user_data)
         try:
-            synapse_api.register_user(
-                registration_shared_secret=registration_shared_secret, user=user
-            )
-            results["register-user"] = True
-            results["user-password"] = user.password
-        except synapse_api.SynapseAPIError as exc:
+            user = User(**user_data)
```
jdkandersson, line 164: rather than dict unpacking, can't we just pass the event data directly? E.g., `User(username=event.params["username"], "admin"=event.params["admin"])`, one fewer level of indirection plus we get linting checks on the arguments

src/pebble.py
```diff
@@ -10,27 +10,28 @@
...
-from synapse import Synapse
 
 logger = logging.getLogger(__name__)
 
 
 class PebbleService:
     """The charm pebble service manager."""
 
-    def __init__(self, synapse: Synapse):
+    def __init__(self, charm_state: CharmState):
```
jdkandersson, line 28: 👍

src/synapse/api.py
```diff
@@ -73,16 +92,21 @@ def _generate_mac(
     nonce: str,
     user: str,
     password: str,
-    admin: bool = False,
+    admin: str | bool = False,
     user_type: typing.Optional[str] = None,
 ) -> str:
     """Generate mac to register user.
 
+    "The MAC is the hex digest output of the HMAC-SHA1 algorithm, with the key being the shared
```
jdkandersson, line 100: are we sure we should use `SHA1`? Its quite short

src/synapse/api.py
```diff
@@ -73,16 +92,21 @@ def _generate_mac(
...
+    secret and the content being the nonce, user, password, either the string "admin" or "notadmin"
+    , and optionally the user_type each separated by NULs.".
```
jdkandersson, line 102: comma on previous line?

src/synapse/workload.py
```diff
@@ -0,0 +1,264 @@
...
+    SYNAPSE_COMMAND_PATH,
+    SYNAPSE_CONFIG_DIR,
+    SYNAPSE_CONFIG_PATH,
+    SYNAPSE_PORT,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class CommandMigrateConfigError(Exception):
```
jdkandersson, line 28: Should these have a common parent?

src/user.py
```diff
@@ -16,8 +20,17 @@ class User(BaseModel):
...
+    password: str = Field(False)
+
+    def __init__(self, **data: dict[str, typing.Any]) -> None:
+        """Initialize a new User instance.
+
+        Parameters:
+            data: A dictionary containing the user data.
+        """
+        super().__init__(**data)
+        self._set_password()
```
jdkandersson, line 33: It would be better to do something like `self.password = _generate_password()` where `_generate_password` can just be a function at the module level since it probably doesn't need access to the user data. This makes it clearer what is happening rather than initialising instance attributes in methods which should generally be avoided

jdkandersson (commented):
Good progress, will now review tests as well

tests/integration/test_charm.py
```diff
@@ -126,3 +126,20 @@ async def test_reset_instance_action(
...
+    """
+    unit = model.applications[synapse_app.name].units[0]
+    action_register_user: Action = await synapse_app.units[0].run_action(  # type: ignore
+        "register-user", username="operator", admin="yes"
+    )
+    await action_register_user.wait()
+    assert action_register_user.status == "completed"
+    assert action_register_user.results["register-user"]
+    assert action_register_user.results["user-password"]
+    assert unit.workload_status == "active"
```
jdkandersson, line 145: There is a standard here which states that further checks should be done to ensure the action worked, is that applicable in this case? https://juju.is/docs/sdk/styleguide#heading--integration-tests

tests/unit/test_register_user_action.py
```diff
@@ -0,0 +1,151 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Register user action unit tests."""
+
+# pylint: disable=protected-access
```
jdkandersson, line 6: Should this include an explanation why this is appropriate in this case?

tests/unit/test_register_user_action.py
```diff
@@ -0,0 +1,151 @@
...
+    monkeypatch.setattr("synapse.get_registration_shared_secret", get_registration_mock)
+    register_user_mock = unittest.mock.MagicMock()
+    monkeypatch.setattr("synapse.register_user", register_user_mock)
+    user = "username"
+    event = unittest.mock.MagicMock(spec=ActionEvent)
+    event.params = {
+        "username": user,
+        "admin": "no",
+    }
+    # Calling to test the action since is not possible calling via harness
```
jdkandersson, line 41: @benhoyt is that right? The harness can't trigger actions?

tests/unit/test_register_user_action.py
```diff
@@ -0,0 +1,151 @@
...
+def test_register_user_registration_none(
+    harness_server_name_configured: Harness, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    """
+    arrange: start the Synapse charm, set Synapse container to be ready and set server_name.
+    act: run register-user action.
+    assert: event fails if registration shared secret is not found.
+    """
+    harness = harness_server_name_configured
+    harness.set_leader(True)
```
jdkandersson, line 60: Would it be worth doing this in the `harness_server_name_configured` fixture?

tests/unit/test_register_user_action.py
```diff
@@ -0,0 +1,151 @@
...
+        "username": user,
+        "admin": "no",
+    }
+    # Calling to test the action since is not possible calling via harness
+    harness.charm._on_register_user_action(event)
+    assert "registration_shared_secret was not found" in event.fail_message
+    assert isinstance(harness.model.unit.status, ops.ActiveStatus)
+
+
+@pytest.mark.parametrize("harness", [0], indirect=True)
```
jdkandersson, line 87: What is `0` here? Would it be worth documenting that a bit, e.g., using `pytest.param` with an `id`?

tests/unit/test_register_user_action.py
```diff
@@ -0,0 +1,151 @@
...
+    harness = harness_server_name_configured
+    harness.set_leader(True)
+    get_registration_mock = unittest.mock.Mock(return_value="shared_secret")
+    monkeypatch.setattr("synapse.get_registration_shared_secret", get_registration_mock)
+    fail_message = "Some fail message"
+    synapse_api_error = synapse.SynapseAPIError(fail_message)
+    register_user_mock = unittest.mock.MagicMock(side_effect=synapse_api_error)
+    monkeypatch.setattr("synapse.register_user", register_user_mock)
+    user = "username"
+    event = unittest.mock.MagicMock(spec=ActionEvent)
```
jdkandersson, line 105: @benhoyt do you think something like this could be simplified? Is it necessary?

tests/unit/test_register_user_action.py
```diff
@@ -0,0 +1,151 @@
...
+    assert isinstance(harness.model.unit.status, ops.ActiveStatus)
+
+
+def test_username_empty():
+    """
+    arrange: create a user.
+    act: set username as empty.
+    assert: ValueError is raised.
+    """
+    user_data: dict[str, typing.Any] = {
```
jdkandersson, line 132: What is the reason for defining this as a dict rather than just passing it as arguments to the constructor please?

tests/unit/test_register_user_action.py
```diff
@@ -0,0 +1,151 @@
...
+        "username": "",
+        "admin": "yes",
+    }
+    with pytest.raises(ValueError, match="Username must not be empty"):
+        _ = User(**user_data)
```
jdkandersson, line 137: I think the `_ = ` is unnecessary?

tests/unit/test_synapse_api.py
```diff
@@ -0,0 +1,119 @@
...
+    get_nonce_return = "nonce"
+    get_nonce_mock = mock.MagicMock(return_value=get_nonce_return)
+    monkeypatch.setattr("synapse.api._get_nonce", get_nonce_mock)
+    generate_mac_mock = mock.MagicMock(return_value="mac")
+    monkeypatch.setattr("synapse.api._generate_mac", generate_mac_mock)
+    mock_response = mock.Mock()
+    mock_response.raise_for_status.return_value = None
+    monkeypatch.setattr("synapse.api.requests.post", mock_response)
+    shared_secret = token_hex(16)
+    synapse.register_user(shared_secret, user)
```
jdkandersson, line 40: Would it be worth separating the arrange, act and assert blocks using a new line? To make it clearer when the setup, execution and checks are being done?

tests/unit/test_register_user_action.py
benhoyt: I believe you can trigger an action more directly with `emit`, but unfortunately the Harness currently stubs out the testing `action_set` (and similar functions) used by `event.set_results`, hence why this needs to be mocked.

Yeah, we should provide Harness support for this. Feel free to open an issue to track.

tests/unit/test_register_user_action.py
benhoyt: Yeah, similar to above -- we should add proper support for actions and related hook tools to Harness.

tests/unit/test_register_user_action.py
jdkandersson: @amandahla would you mind opening that issue please?

actions.yaml
```diff
@@ -18,7 +18,7 @@ register-user:
       type: string
     admin:
       description: Whether to create an admin user ("yes" or "no")
-      type: boolean
+      type: string
```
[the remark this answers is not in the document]
merkata: That's an interesting one. Supplying "yes" or "no" maps to the cli option you can pass when creating users, so it might be good to expose the same option via a juju action. In yaml terms, "yes" and "no" are boolean, like 0 and 1, true and false. I guess that breaks when entering juju land, though I'm not sure? If I change this to type boolean, can I still use "yes" and "no", but then the other options as well?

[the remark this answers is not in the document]
amandahla: I was thinking about this one and it might not have a reason anymore for being a string since the user is registered via API now and there the parameter is a boolean. You can see a POST request example [here](https://matrix-org.github.io/synapse/v1.41/admin_api/register_api.html). Actually, being a bool would make things easier, not needing to validate anything. I'll change this.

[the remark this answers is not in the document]
amandahla: I didnt test if is possible to use yes/1/true while passing parameters to an action, its a good question.

src/charm.py
amandahla: Your comment gave me this idea and I think is kind of related to something that @gregory-schiano suggested. I created an actions module (not with observers, only a module to extract the interactions from the charm) so this made it possible to create specific exceptions like actions.ResetInstanceError for example. Please take a look and tell me what you think. I'll close this comment since is outdated compared to the change that I made.

src/charm.py
amandahla: I hope that you like the change that I made here creating a module for actions 🙈

src/charm.py
amandahla: No problem, changed.

src/synapse/api.py
amandahla: Yes, I didnt find any discussion regarding that but here is the [code](https://github.com/matrix-org/synapse/blob/develop/synapse/rest/admin/users.py#L587).

src/synapse/api.py
amandahla: If I change will raise line too long.

tests/unit/test_register_user_action.py
amandahla: Changed :)

src/user.py
amandahla: Agree, changed.

src/charm.py
amandahla: I'll mark this as resolved since the code has changed a lot, feel free to reopen it.

amandahla:
I'll continue working on the reviews of the tests on Monday.

actions.yaml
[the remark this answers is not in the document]
jdkandersson: In general - we don't need to make the charm be consistent with manually configuring synapse. It would be good to think about what an operator would expect - especially one that isn't that familiar with the application

src/synapse/api.py
jdkandersson: what if `"notadmin"` is moved to the next line then?

src/actions/reset_instance.py
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env python3
+
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Class to interact with Reset Instance action."""
```
jdkandersson, line 6: Is it a class?

src/charm.py
```diff
@@ -152,21 +137,14 @@ def _on_register_user_action(self, event: ActionEvent) -> None:
...
-            return
-        user_data = {
-            "username": event.params["username"],
-            "admin": event.params["admin"],
-        }
         try:
-            user = User(**user_data)
-            synapse.register_user(registration_shared_secret=registration_shared_secret, user=user)
-        except (ValidationError, synapse.SynapseAPIError) as exc:
+            password = actions.register_user(
```
jdkandersson, line 141: Do you think it makes more sense to return the user object? If I had to guess what the `register_user` method returns it would be an object representing the user

tests/integration/test_charm.py
amandahla: Thanks, I added a login check.

tests/unit/test_register_user_action.py
amandahla: I found one already open:
https://github.com/canonical/operator/issues/762

tests/unit/test_register_user_action.py
amandahla: Yes, changed.

tests/unit/test_register_user_action.py
amandahla: You made me realize that this is not even needed, I removed and change the other ones to use pytest.param with an id, thanks.

tests/unit/test_synapse_api.py
amandahla: Sure, I added comments now.

src/actions/reset_instance.py
amandahla: Oops, definitely not.

src/charm.py
amandahla: I agree. Changed.

src/charm.py
amandahla: Thanks!

actions.yaml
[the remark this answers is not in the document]
amandahla: I'm closing this one since the parameter is a boolean now but feel free to reopen it or add new reviews.

nrobinaubertin (approved):
Very nice and easy to read

merkata (approved):
Great work!

PR #16, opened 2023-07-24
Add COS - Prometheus integration

amandahla:
Add Prometheus integration.

The configuration generated by migrate-config does not enable metrics so this PR adds the integration and also change the configuration to enable the metrics endpoint.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -200,3 +200,26 @@ async def postgresql_app_fixture(
     async with ops_test.fast_forward():
         await model.deploy(postgresql_app_name, channel="14/stable", trust=True)
         await model.wait_for_idle()
+
+
+@pytest.fixture(scope="module", name="prometheus_app_name")
+def prometheus_app_name_fixture() -> str:
+    """Return the name of the prometheus application deployed for tests."""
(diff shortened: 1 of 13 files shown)


tests/integration/conftest.py
```diff
@@ -200,3 +200,26 @@ async def postgresql_app_fixture(
...
+@pytest_asyncio.fixture(scope="module", name="prometheus_app")
+async def deploy_prometheus_fixture(
+    model: Model,
+    prometheus_app_name: str,
+):
+    """Deploy prometheus."""
+    app = await model.deploy(
+        "prometheus-k8s",
+        application_name=prometheus_app_name,
+        channel="latest/edge",
```
nrobinaubertin, line 220: Why do we need latest/edge ? (I'm maybe out of the loop on that one)

nrobinaubertin (approved):

tests/integration/conftest.py
amandahla: Good catch. This is the version used in COS production env and if I'm not mistaken, is being used because is the one that receives more fixes.

merkata (approved):

PR #18, opened 2023-08-09
WIP - Add workload version to juju status

amandahla:
WIP - Not ready for reviews but feel free to take a look.

This PR adds the Synapse version to the workload version so the user can check using the `juju status` command.
The version can be obtained through a request to the [API](https://matrix-org.github.io/synapse/latest/admin_api/version_api.html).

Out of the scope: the docs field was removed from metadata since now Charmcraft checks if the field is not empty. There is no documentation for Synapse yet.

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -8,7 +8,6 @@ description: |
   A charm for the matrix synapse chat server.
   Synapse is a drop in replacement for other chat servers like Mattermost and Slack.
   This charm is useful if you want to spin up your own chat instance.
-docs: ""
 issues: https://github.com/canonical/synapse-operator/issues
 maintainers:
   - launchpad.net/~canonical-is-devops
--- a/src/synapse/api.py
+++ b/src/synapse/api.py
@@ -10,6 +10,7 @@
 import hashlib
 import hmac
 import logging
+import re
 import typing
 
 import requests
@@ -20,6 +21,7 @@ logger = logging.getLogger(__name__)
 
 SYNAPSE_URL = "http://localhost:8008"
 REGISTER_URL = f"{SYNAPSE_URL}/_synapse/admin/v1/register"
+VERSION_URL = f"{SYNAPSE_URL}/_synapse/admin/v1/server_version"
 
 
 class APIError(Exception):
@@ -149,3 +151,35 @@ def _get_nonce() -> str:
     ) as exc:
         logger.error("Failed to request %s : %s", REGISTER_URL, exc)
         raise NetworkError(f"Failed to request {REGISTER_URL}.") from exc
+
+
+def get_version() -> str:
+    """Get version.
+
+    Expected API output:
+    {
+        "server_version": "0.99.2rc1 (b=develop, abcdef123)",
+        "python_version": "3.7.8"
+    }
+
+    Returns:
+        The version returned by Synapse API.
+
+    Raises:
+        NetworkError: if there was an error fetching the version.
+    """
+    try:
+        res = requests.get(VERSION_URL, timeout=5)
+        res.raise_for_status()
+        server_version = res.json()["server_version"]
+        version_match = re.search(r"^([^\s(]+)", server_version)
+        if version_match:
+            return version_match.group(1)
+        return server_version
+    except (
+        requests.exceptions.ConnectionError,
+        requests.exceptions.Timeout,
+        requests.exceptions.HTTPError,
+    ) as exc:
+        logger.error("Failed to request %s : %s", VERSION_URL, exc)
+        raise NetworkError(f"Failed to request {VERSION_URL}.") from exc
--- a/src/synapse/workload.py
+++ b/src/synapse/workload.py
@@ -77,7 +77,7 @@ def check_ready() -> typing.Dict:
     check = Check(CHECK_READY_NAME)
     check.override = "replace"
     check.level = "ready"
-    check.tcp = {"port": SYNAPSE_PORT}
+    check.http = {"url": f"http://localhost:{SYNAPSE_PORT}/_synapse/admin/v1/server_version"}
     # _CheckDict cannot be imported
     return check.to_dict()  # type: ignore
 
--- a/tests/unit/test_synapse_api.py
+++ b/tests/unit/test_synapse_api.py
@@ -87,9 +87,9 @@ def test_generate_mac():
 @mock.patch("synapse.api.requests")
 def test_get_nonce_success(mock_requests):
     """
-    arrange: set User parameters.
-    act: register the user.
-    assert: parameters are passed correctly.
+    arrange: none.
+    act: get nonce.
+    assert: _get_nonce return the correct value.
     """
     nonce_value = "nonce_value"
     mock_response = mock.MagicMock()
@@ -102,8 +102,8 @@ def test_get_nonce_success(mock_requests):
 
 def test_get_nonce_error(monkeypatch: pytest.MonkeyPatch):
     """
-    arrange: set User parameters.
-    act: register the user.
+    arrange: none.
+    act: get nonce.
     assert: NetworkError is raised.
     """
     mock_response_error = requests.exceptions.ConnectionError("Connection error")
@@ -111,3 +111,33 @@ def test_get_nonce_error(monkeypatch: pytest.MonkeyPatch):
     monkeypatch.setattr("synapse.api.requests.get", mock_response)
     with pytest.raises(synapse.APIError, match="Failed to request"):
         synapse.api._get_nonce()
+
+
+@mock.patch("synapse.api.requests")
+def test_get_version_success(mock_requests):
+    """
+    arrange: none.
+    act: get version.
+    assert: get_version return the correct value.
+    """
+    extracted_version = "0.99.2rc1"
+    mock_response = mock.MagicMock()
+    mock_response.json.return_value = {
+        "server_version": f"{extracted_version} (b=develop, abcdef123)",
+        "python_version": "3.7.8",
+    }
+    mock_requests.get.return_value = mock_response
+    assert synapse.api.get_version() == extracted_version
+
+
+def test_get_version_error(monkeypatch: pytest.MonkeyPatch):
+    """
+    arrange: none.
+    act: get version.
+    assert: NetworkError is raised.
+    """
+    mock_response_error = requests.exceptions.ConnectionError("Connection error")
+    mock_response = mock.Mock(side_effect=mock_response_error)
+    monkeypatch.setattr("synapse.api.requests.get", mock_response)
+    with pytest.raises(synapse.APIError, match="Failed to request"):
+        synapse.api.get_version()
--- a/src/charm.py
+++ b/src/charm.py
@@ -15,6 +15,7 @@ from ops.charm import ActionEvent
 from ops.main import main
 
 import actions
+import synapse
 from charm_state import CharmConfigInvalidError, CharmState
 from constants import SYNAPSE_CONTAINER_NAME, SYNAPSE_PORT
 from database_observer import DatabaseObserver
@@ -93,6 +94,11 @@ class SynapseCharm(ops.CharmBase):
         Args:
             event: Event triggering after pebble is ready.
         """
+        try:
+            synapse_version = synapse.get_version()
+            self.unit.set_workload_version(synapse_version)
+        except synapse.APIError as exc:
+            logger.debug("Cannot set workload version at this time: %s", exc)
         self.change_config(event)
 
     def _on_reset_instance_action(self, event: ActionEvent) -> None:
--- a/src/synapse/__init__.py
+++ b/src/synapse/__init__.py
@@ -4,7 +4,7 @@
 """Synapse package is used to interact with Synapse instance."""
 
 # Exporting methods to be used for another modules
-from .api import APIError, register_user  # noqa: F401
+from .api import APIError, get_version, register_user  # noqa: F401
 from .workload import (  # noqa: F401
     ExecResult,
     WorkloadError,
--- a/tests/integration/test_charm.py
+++ b/tests/integration/test_charm.py
@@ -3,7 +3,9 @@
 # See LICENSE file for licensing details.
 
 """Integration tests for Synapse charm."""
+import json
 import logging
+import re
 import typing
 
 import pytest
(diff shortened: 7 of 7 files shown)


src/synapse/api.py
```diff
@@ -149,3 +151,35 @@ def _get_nonce() -> str:
...
+        version_match = re.search(r"^([^\s(]+)", server_version)
+        if version_match:
+            return version_match.group(1)
+        return server_version
+    except (
+        requests.exceptions.ConnectionError,
+        requests.exceptions.Timeout,
+        requests.exceptions.HTTPError,
+    ) as exc:
+        logger.error("Failed to request %s : %s", VERSION_URL, exc)
```
weiiwang01, line 184: When logging exceptions, using `%s` only print the message part for most of exception, without the exception type.

```
print("%s" % RuntimeError("some error")) # print some error
print("%r" % RuntimeError("some error")) # print RuntimeError('some error')
```

But it will be even better if we use `logger.exception` which will print the traceback of the exception as well.

metadata.yaml
```diff
@@ -8,7 +8,6 @@ description: |
   A charm for the matrix synapse chat server.
   Synapse is a drop in replacement for other chat servers like Mattermost and Slack.
   This charm is useful if you want to spin up your own chat instance.
-docs: ""
```
arturo-seijas, line 11: May I suggest you create a discourse post to link here?

src/synapse/api.py
amandahla: Great suggestion! Changed, thanks

src/synapse/api.py
```diff
@@ -147,5 +152,45 @@ def _get_nonce() -> str:
...
+    try:
+        session = Session()
+        retries = Retry(
+            total=3,
+            backoff_factor=3,
+        )
+        session.mount("http://", HTTPAdapter(max_retries=retries))
+        res = session.get(VERSION_URL, timeout=10)
+        res.raise_for_status()
+        server_version = res.json()["server_version"]
```
jdkandersson, line 185: Should the case where this lookup fails be handled gracefully? E.g., what if the `server_version` is not returned? Or no JSON is returned? Or a dict isn't returned?

src/synapse/workload.py
```diff
@@ -69,14 +69,28 @@ class ExecResult(typing.NamedTuple):
...
-    """Return the Synapse container check.
+    """Return the Synapse container ready check.
 
     Returns:
         Dict: check object converted to its dict representation.
     """
     check = Check(CHECK_READY_NAME)
     check.override = "replace"
     check.level = "ready"
+    check.http = {"url": f"http://localhost:{SYNAPSE_PORT}/_synapse/admin/v1/server_version"}
```
jdkandersson, line 80: I think this is a constant in another file also?

src/synapse/workload.py
```diff
@@ -69,14 +69,28 @@ class ExecResult(typing.NamedTuple):
...
+    # _CheckDict cannot be imported
+    return check.to_dict()  # type: ignore
```
jdkandersson, line 82: @benhoyt any comment on this please?

tests/unit/test_synapse_api.py
```diff
@@ -87,9 +87,9 @@ def test_generate_mac():
 @mock.patch("synapse.api.requests")
 def test_get_nonce_success(mock_requests):
     """
-    arrange: set User parameters.
-    act: register the user.
-    assert: parameters are passed correctly.
+    arrange: none.
```
jdkandersson, line 90: So there is no pre-condition at all? What about the fact that certain things are mocked?

tests/unit/test_synapse_api.py
```diff
@@ -102,12 +102,45 @@ def test_get_nonce_success(mock_requests):
 
 def test_get_nonce_error(monkeypatch: pytest.MonkeyPatch):
     """
-    arrange: set User parameters.
-    act: register the user.
+    arrange: none.
```
jdkandersson, line 105: Similar to above

jdkandersson: And I suppose applicable for the others as well

tests/unit/test_synapse_api.py
```diff
@@ -102,12 +102,45 @@ def test_get_nonce_success(mock_requests):
...
     """
     mock_response_error = requests.exceptions.ConnectionError("Connection error")
     mock_response = mock.Mock(side_effect=mock_response_error)
     monkeypatch.setattr("synapse.api.requests.get", mock_response)
     with pytest.raises(synapse.APIError, match="Failed to request"):
         synapse.api._get_nonce()
+
+
+@mock.patch("synapse.api.Session")
+def test_get_version_success(mock_session):
```
jdkandersson, line 117: Perhaps IO missed, it, there seems to be 2 cases, one where the regex matches and one where it doesn't, should there be a test for both cases?

tests/unit/test_synapse_api.py
amandahla: I used Flask PR (https://github.com/canonical/flask-k8s-operator/pull/1) as a reference.

I assumed that arrange would be something related to the charm (deployed, integrated with some database, etc.).
But I can add what is being mocked then.

src/synapse/api.py
amandahla: Sure, I added exceptions to this now.

src/synapse/workload.py
amandahla: Good point, changed.

tests/unit/test_synapse_api.py
amandahla: Arrange changed.

tests/unit/test_synapse_api.py
amandahla: Arrange changed.

tests/unit/test_synapse_api.py
amandahla: Good point, I added a new unit test for that.

src/synapse/workload.py
benhoyt: This is not true anymore. I made `CheckDict` public in ops 2.4 in [as part of this PR](https://github.com/canonical/operator/pull/942/files#diff-d74b54ce28b898e8febb0f33f24abc9bf91bb7793c51ae716bf9a9816d90104d). It's not imported to the top level of `ops`, but it's accessible as `ops.pebble.CheckDict` (similar with `PlanDict` and so on).

src/synapse/api.py
```diff
@@ -182,15 +192,26 @@ def get_version() -> str:
...
-        requests.exceptions.ConnectionError,
-        requests.exceptions.Timeout,
-        requests.exceptions.HTTPError,
-    ) as exc:
-        logger.exception("Failed to request %s : %r", VERSION_URL, exc)
-        raise NetworkError(f"Failed to request {VERSION_URL}.") from exc
+        res_json = res.json()
+        server_version = res_json.get("server_version", None)
+        if server_version is None:
+            # Exception not in docstring because is captured.
```
jdkandersson, line 198: Not following this, where is it handled please?

src/synapse/api.py
```diff
@@ -182,15 +192,26 @@ def get_version() -> str:
...
-        return server_version
-    except (
-        requests.exceptions.ConnectionError,
-        requests.exceptions.Timeout,
-        requests.exceptions.HTTPError,
-    ) as exc:
-        logger.exception("Failed to request %s : %r", VERSION_URL, exc)
-        raise NetworkError(f"Failed to request {VERSION_URL}.") from exc
+        res_json = res.json()
+        server_version = res_json.get("server_version", None)
```
jdkandersson, line 196: What about the case where `res.json()` is not a dict?

src/synapse/api.py
```diff
@@ -182,15 +192,26 @@ def get_version() -> str:
...
+        if server_version is None:
+            # Exception not in docstring because is captured.
+            raise VersionNotFoundError(  # noqa: DCO053
+                f"There is no server_version in JSON output: {res_json}"
+            )
+        version_match = re.search(SYNAPSE_VERSION_REGEX, server_version)
+        if not version_match:
+            # Exception not in docstring because is captured.
```
jdkandersson, line 204: Also not following this?

src/synapse/api.py
amandahla: They both are handled in line:
```
except GetVersionError as exc:
```
So while calling register_user, there is no point to handle this kind of exception. That's why I disabled DC0053.

src/synapse/api.py
amandahla: Closing this one in favor of the first answer.

src/synapse/api.py
amandahla: Changed to check in this call and also the others.

src/synapse/workload.py
amandahla: FIxed, thanks.

src/synapse/api.py
```diff
@@ -69,36 +77,39 @@ def register_user(registration_shared_secret: str, user: User) -> None:
...
-        requests.exceptions.HTTPError,
-    ) as exc:
-        logger.exception("Failed to request %s : %r", REGISTER_URL, exc)
-        raise NetworkError(f"Failed to request {REGISTER_URL}.") from exc
+    except (requests.exceptions.ConnectionError, requests.exceptions.Timeout) as exc:
+        logger.exception("Failed to connect to %s: %r", REGISTER_URL, exc)
+        raise NetworkError(f"Failed to connect to {REGISTER_URL}.") from exc
+    except requests.exceptions.HTTPError as exc:
+        logger.exception("HTTP error from %s: %r", REGISTER_URL, exc)
+        raise NetworkError(f"HTTP error from {REGISTER_URL}.") from exc
```
jdkandersson, line 109: Should we scope catching these exceptions to specifically the lines where the can occur? I.e., lines 102 and 103?

src/synapse/api.py
```diff
@@ -69,36 +77,39 @@ def register_user(registration_shared_secret: str, user: User) -> None:
...
+    except GetNonceError as exc:
+        logger.exception("Failed to get nonce: %r", exc)
+        raise GetNonceError(str(exc)) from exc
```
jdkandersson, line 112: If we just want to re-raise it is possible just to write `raise`, that way we don't make the stack trace confusing by having the same exception listed multiple times. I.e.:

```python
    except GetNonceError as exc:
        logger.exception("Failed to get nonce: %r", exc)
        raise
```

src/synapse/api.py
```diff
@@ -151,18 +162,31 @@ def _get_nonce() -> str:
...
+        return nonce
+    except (requests.exceptions.ConnectionError, requests.exceptions.Timeout) as exc:
+        logger.exception("Failed to connect to %s: %r", REGISTER_URL, exc)
+        raise NetworkError(f"Failed to connect to {REGISTER_URL}.") from exc
+    except requests.exceptions.HTTPError as exc:
+        logger.exception("HTTP error from %s: %r", REGISTER_URL, exc)
+        raise NetworkError(f"HTTP error from {REGISTER_URL}.") from exc
+    except (GetNonceError, requests.exceptions.JSONDecodeError) as exc:
+        logger.exception("Failed to get nonce: %r", exc)
+        raise GetNonceError(str(exc)) from exc
```
jdkandersson, line 189: I'll try and have a go, see what you think:

```python
   try:
        res = requests.get(REGISTER_URL, timeout=5)
        res.raise_for_status()
        res_json = res.json()
    except (requests.exceptions.ConnectionError, requests.exceptions.Timeout) as exc:
        logger.exception("Failed to connect to %s: %r", REGISTER_URL, exc)
        raise NetworkError(f"Failed to connect to {REGISTER_URL}.") from exc
    except requests.exceptions.HTTPError as exc:
        logger.exception("HTTP error from %s: %r", REGISTER_URL, exc)
        raise NetworkError(f"HTTP error from {REGISTER_URL}.") from exc
    # TODO: I think there are also possible exceptions for res.json(), I'm not sure what they are, e.g., if the response was 201 I'm not sure what the behaviour of res.json() is

    if not isinstance(res_json, dict):
        raise GetNonceError(f"Response has unexpected encode: {res_json}")
    nonce = res_json.get("nonce", None)
    if nonce is None:
        raise NonceNotFoundError(
            f"There is no nonce in JSON output: {res_json}"
        )
    return nonce
```

src/synapse/api.py
```diff
@@ -193,6 +217,11 @@ def get_version() -> str:
         res = session.get(VERSION_URL, timeout=10)
         res.raise_for_status()
         res_json = res.json()
+        if not isinstance(res_json, dict):
```
jdkandersson, line 220: See example above for how to simplify this

src/synapse/api.py
```diff
@@ -212,6 +241,6 @@ def get_version() -> str:
     except requests.exceptions.HTTPError as exc:
         logger.exception("HTTP error from %s: %r", VERSION_URL, exc)
         raise NetworkError(f"HTTP error from {VERSION_URL}.") from exc
-    except GetVersionError as exc:
+    except (GetVersionError, requests.exceptions.JSONDecodeError) as exc:
```
jdkandersson, line 244: Ah, probably `requests.exceptions.JSONDecodeError` needs to be handled whenever we try to convert a response to JSON

jdkandersson:
Generally speaking, raising an exception within a function and then handling it within the same function is confusing for the reader and indicates opportunities to improve the error handling flow in a function

src/synapse/api.py
amandahla: You're right, thanks!

src/synapse/api.py
amandahla: You made me realize that it was really confusing before. Now I changed everything based on your idea. Please take a look.
I'm just concerned that I couldn't add Exceptions in docstring that I know will be raised but the lint says it should not be documented.

src/synapse/api.py
amandahla: Changed :)

src/synapse/api.py
amandahla: Changed that too.

src/synapse/api.py
```diff
@@ -129,23 +141,85 @@ def _generate_mac(
...
+    """
+    res = _send_get_request(VERSION_URL, retry=True)
+    try:
+        server_version = res.json()["server_version"]
+    except (requests.exceptions.JSONDecodeError, KeyError, TypeError) as exc:
+        logger.exception("Failed to decode version: %r. Received: %s", exc, res.text)
+        raise GetVersionError(str(exc)) from exc
+    version_match = re.search(SYNAPSE_VERSION_REGEX, server_version)
+    if not version_match:
+        # Exception not in docstring because is captured.
```
jdkandersson, line 221: Perhaps I'm missing something, where is the exception handled in this function please?

amandahla: The exception is actually raised, fixed.

src/synapse/api.py
amandahla: Since the code is updated, I'll close this review but feel free to reopen it.

jdkandersson (approved):

weiiwang01 (approved):

PR #19, opened 2023-08-11
Add SAML integration

amandahla:
Tested with the change merged on this PR https://github.com/canonical/saml-integrator-operator/pull/37

This PR adds SAML integration by using the SAML Integrator charm library.

What happens after merging it:
- Refactor of database observer.
- Add integration with SAML integrator so the user can enable SAML authentication.
- Add new type SAMLConfiguration.
- Add new observer SAMLObserver.
- Add new configuration public_baseurl.
- Add new container Synapse NGINX as a reverse proxy.
- Add unit and integration tests.


lib/charms/saml_integrator/v0/saml.py
```diff
@@ -0,0 +1,303 @@
...
+
+    def _on_relation_changed(self, event: ops.RelationChangedEvent) -> None:
+        """Event emitted when the relation has changed.
+
+        Args:
+            event: event triggering this handler.
+        """
+        if not self.charm.unit.is_leader():
+            return
+        self.on.saml_data_available.emit(event.relation, app=event.app, unit=event.unit)
```
weiiwang01, line 266: @arturo-seijas I have some questions regarding the `saml_data_available` event. Why is it emitted only on the leader unit when the requirer simply reads the reaction data without writing to the relation data? Additionally, considering that relation data might not always exist when the `relation-changed` event occurs, like the first `relation-changed` after `relation-joined`, shouldn't we emit the `saml_data_available` event only when the relation data is present?

arturo-seijas: Good point. I'll be releasing a new version with these changes. Thank you

lib/charms/grafana_k8s/v0/grafana_dashboard.py
```diff
@@ -219,7 +219,7 @@ def __init__(self, *args):
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
 
-LIBPATCH = 32
+LIBPATCH = 33
```
arturo-seijas, line 222: Next time, I suggest updating the libs in a separate PR so that PRs are more atomic and changes can be reverted more easily

arturo-seijas (approved):
LGTM

tests/unit/test_synapse_workload.py
```diff
@@ -61,3 +62,65 @@ def test_enable_metrics_error(monkeypatch: pytest.MonkeyPatch):
...
+    listeners:
+        - type: http
+          port: 8080
+          bind_addresses:
+            - "::"
+    """
+    text_io_mock = io.StringIO(config_content)
+    pull_mock = Mock(return_value=text_io_mock)
+    push_mock = MagicMock()
+    container_mock = MagicMock()
```
jdkandersson, line 84: Any thoughts @weiiwang01 about using the new way to access the container without needing to mock?

src/pebble.py
```diff
@@ -70,10 +76,30 @@ def change_config(self, container: ops.model.Container) -> None:
         try:
             synapse.execute_migrate_config(container=container, charm_state=self._charm_state)
             synapse.enable_metrics(container=container)
+            if self._charm_state.saml_config is not None:
+                logger.debug("pebble.change_config: Enabling SAML")
+                synapse.enable_saml(container=container, charm_state=self._charm_state)
```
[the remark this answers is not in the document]
amandahla: I believe that is not needed here because in this case the plan will be changed (when a configuration changes, the env vars change too). Do you see a different scenario?

src/saml_observer.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def _enable_saml(self) -> None:
+        """Enable  SAML."""
+        container = self._charm.unit.get_container(SYNAPSE_CONTAINER_NAME)
+        if not container.can_connect() or self._pebble_service is None:
+            self._charm.unit.status = ops.MaintenanceStatus("Waiting for pebble")
+            return
+        try:
+            self._pebble_service.enable_saml(container)
+        except Exception as exc:  # pylint: disable=broad-exception-caught
```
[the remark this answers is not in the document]
amandahla: No reason, changed to get PebbleServiceError as expected.

tests/integration/conftest.py
```diff
@@ -259,3 +259,36 @@ async def deploy_prometheus_fixture(
...
+        app = await model.deploy(
+            saml_integrator_app_name,
+            application_name=saml_integrator_app_name,
+            channel="latest/edge",
+            trust=True,
+        )
+        await model.wait_for_idle()
+    entity_id = "https://login.staging.ubuntu.com"
+    metadata_url = "https://login.staging.ubuntu.com/saml/metadata"
+    await app.set_config(  # type: ignore[attr-defined]
```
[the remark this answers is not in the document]
amandahla: Actually, I have no idea why this is here. Removed.

src/pebble.py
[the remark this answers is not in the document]
jdkandersson: Hmm, somewhat dissatisfying that we have cases where we can rely on pebble to perform the restart and others where we have to be explicit. If we change something and pebble doesn't do the restart anymore, will we know that we have do to the restart? Perhaps at least a comment?

src/synapse/workload.py
```diff
@@ -100,6 +101,20 @@ def check_alive() -> ops.pebble.CheckDict:
...
+    """Return the Synapse NGINX container check.
+
+    Returns:
+        Dict: check object converted to its dict representation.
+    """
+    check = Check(CHECK_READY_NAME)
+    check.override = "replace"
+    check.level = "ready"
+    check.tcp = {"port": SYNAPSE_NGINX_PORT}
+    # _CheckDict cannot be imported
```
jdkandersson, line 114: I seem to remember that this is not right anymore?

tests/integration/conftest.py
```diff
@@ -57,6 +57,14 @@ def synapse_image_fixture(pytestconfig: Config):
     return synapse_image
 
 
+@pytest_asyncio.fixture(scope="module", name="synapse_nginx_image")
+def synapse_nginx_image_fixture(pytestconfig: Config):
+    """Get value from parameter synapse-nginx-image."""
+    synapse_nginx_image = pytestconfig.getoption("--synapse-nginx-image")
```
jdkandersson, line 63: Perhaps `--synapse-nginx-image` should be a constant since I think it is in 3 places now?

metadata.yaml
```diff
@@ -22,11 +22,17 @@ containers:
...
+  synapse-nginx:
+    resource: synapse-nginx-image
 
 resources:
   synapse-image:
     type: oci-image
     description: OCI image for Synapse
+  synapse-nginx-image:
+    type: oci-image
+    description: Docker image for nginx Synapse
```
mthaddon, line 34: Can we say "OCI image" rather than "Docker image" here?

src/pebble.py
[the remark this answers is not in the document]
amandahla: It would be nice if this was controlled by the container.replan itself.
I agree that might be confusing in the future. I'll add a comment warning about it similar to the one in enable_saml.

tests/unit/conftest.py
```diff
@@ -228,6 +228,39 @@ def harness_with_postgresql_fixture(
     return harness
 
 
+@pytest.fixture(name="harness_with_saml")
+def harness_with_saml_fixture(
```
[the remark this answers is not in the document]
amandahla: Sorry for the delay. This was a way to force harness reinitialize the charm was would happen in Juju since the charm we rely on charm_state to get information from the observers and this variable is updated only in the __init__.
I don't remember where I saw this way of forcing but I saw some charms doing this way like [here](https://github.com/canonical/prometheus-k8s-operator/blob/main/tests/unit/test_remote_write.py#L290).

tests/integration/conftest.py
amandahla: I only found 2 places where --synapse-nginx-image appears, same for other integration test parameters.

metadata.yaml
amandahla: Sure!

tests/integration/test_charm.py
```diff
@@ -230,3 +230,27 @@ async def test_register_user_action(
         )
         assert response.status_code == 200
         assert response.json()["access_token"]
+
+
+@pytest.mark.asyncio
+async def test_saml_integration(
+    model: Model,
+    synapse_app: Application,
+    saml_integrator_app,  # pylint: disable=unused-argument
```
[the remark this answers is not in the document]
amandahla: Thanks, changed.

tests/unit/test_charm.py
```diff
@@ -116,6 +117,58 @@ def test_traefik_integration(harness_server_name_configured: Harness) -> None:
...
 
+def test_saml_integration_container_restart(
+    harness_server_name_configured: Harness, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    """
+    arrange: start the Synapse charm, set server_name, mock container and enable_saml.
+    act: enable saml via pebble_service as the observer does.
+    assert: The container is restarted.
+    """
+    harness = harness_server_name_configured
```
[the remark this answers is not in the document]
amandahla: I added a line between arrange,act and assert now.

src/synapse/workload.py
amandahla: Oops. Removed.

src/pebble.py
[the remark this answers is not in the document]
jdkandersson: @benhoyt any comment on this?

tests/integration/conftest.py
jdkandersson: It appears in `tests/conftest.py` and I'm guessing it should appear twice in this function, once on this line and once on the next line, though that says `--synapse-image` at the moment which I'm guessing should say `--synapse-nginx-image`

src/pebble.py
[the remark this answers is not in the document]
benhoyt: It looks to me like `synapse.execute_migrate_config`, `synapse.enable_metrics`, and `synapse.enable_saml` only do a `container.push`, not update the plan. If I'm reading that correctly, `container.replan()` won't have any effect -- you'll need to use `container.restart(service)` explicitly.

In general, Pebble doesn't monitor config files for changes, so there are two cases needed:

1) The plan changing, e.g., the service's environment variables or command line. In this case a Pebble `replan` will restart the service.
2) A config file changed but the plan hasn't changed. In this case Pebble will need to be explicitly told to restart the services in question.

Here it's hard to tell -- it's very round-about, because `self.replan(container)` does add a layer, which depends on `synapse.get_environment()` (which in turn depends on the charm state). So if the charm state has changed since the plan was last generated, a replan would be enough.

In this case, since config files are involved as well, I'd probably change this to unconditionally do a restart (or maybe there's a way tell Synapse to do a graceful, zero-downtime restart, like sending SIGHUP?), seeing it's going to happen sometimes anyway. However, the ideal would be to detect if any of the config files changed (during the pull-push sequences) and only restart if they or the plan has changed.

src/pebble.py
```diff
@@ -50,14 +50,28 @@ def __init__(self, charm_state: CharmState):
...
 
         Args:
             container: Charm container.
+            restart: Force restart of the container. Default: False.
         """
+        logger.debug("Replaning the container")
         container.add_layer(SYNAPSE_CONTAINER_NAME, self._pebble_layer, combine=True)
         container.replan()
+        if restart:
+            container.restart(SYNAPSE_SERVICE_NAME)
```
benhoyt, line 64: In the case where a replan is needed and restarts the service *and* `restart` is True, this will restart the service twice in succession -- probably not what you want. See comment below too. I think trying to roll up a "replan maybe restart" into one function here is a bit confusing.

tests/unit/test_synapse_workload.py
weiiwang01: Yes, it's possible to use the new feature in the ops testing harness to prepare files in the container filesystem.
```diff
diff --git a/tests/unit/test_synapse_workload.py b/tests/unit/test_synapse_workload.py
index 816686e..c035774 100644
--- a/tests/unit/test_synapse_workload.py
+++ b/tests/unit/test_synapse_workload.py
@@ -82,17 +82,11 @@ def test_enable_saml_success(harness_with_saml: Harness, monkeypatch: pytest.Mon
             - "::"
           x_forwarded: false
     """
-    text_io_mock = io.StringIO(config_content)
-    pull_mock = Mock(return_value=text_io_mock)
-    push_mock = MagicMock()
-    container_mock = MagicMock()
-    monkeypatch.setattr(container_mock, "pull", pull_mock)
-    monkeypatch.setattr(container_mock, "push", push_mock)
-
-    synapse.enable_saml(container_mock, harness.charm._charm_state)
+    container_root = harness.get_filesystem_root("synapse")
+    synapse_config_file = container_root / "data/homeserver.yaml"
+    synapse_config_file.write_text(config_content, encoding="utf-8")
 
-    assert pull_mock.call_args[0][0] == SYNAPSE_CONFIG_PATH
-    assert push_mock.call_args[0][0] == SYNAPSE_CONFIG_PATH
+    synapse.enable_saml(harness.charm.unit.get_container("synapse"), harness.charm._charm_state)
     saml_relation = harness.model.get_relation("saml")
     assert saml_relation
     saml_relation_data = harness.get_relation_data(saml_relation.id, "saml-integrator")
@@ -115,7 +109,9 @@ def test_enable_saml_success(harness_with_saml: Harness, monkeypatch: pytest.Mon
             },
         },
     }
-    assert push_mock.call_args[0][1] == yaml.safe_dump(expected_config_content)
+    assert synapse_config_file.read_text(encoding="utf-8") == yaml.safe_dump(
+        expected_config_content
+    )
 
 
 def test_enable_saml_error(harness_with_saml: Harness, monkeypatch: pytest.MonkeyPatch):
```

tests/unit/conftest.py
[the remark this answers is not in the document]
benhoyt: So I removed the `harness._framework` internals fiddling, and the test that failed was `test_enable_saml_success`, so I started editing that. A simpler and more "Harness-native" way to write that test would be to set up the relation data this function needs, then call `harness.begin()`. This avoids the re-init problem and avoids this complicated fixture altogether:

```diff
diff --git a/tests/unit/test_synapse_workload.py b/tests/unit/test_synapse_workload.py
index 816686e..e834e26 100644
--- a/tests/unit/test_synapse_workload.py
+++ b/tests/unit/test_synapse_workload.py
@@ -13,8 +13,9 @@ import pytest
 import yaml
 from ops.testing import Harness
 
+from charm import SynapseCharm
 import synapse
-from constants import SYNAPSE_CONFIG_PATH
+from constants import SYNAPSE_CONFIG_PATH, TEST_SERVER_NAME, SYNAPSE_CONTAINER_NAME
 
 
 def test_enable_metrics_success(monkeypatch: pytest.MonkeyPatch):
@@ -67,13 +68,17 @@ def test_enable_metrics_error(monkeypatch: pytest.MonkeyPatch):
         synapse.enable_metrics(container_mock)
 
 
-def test_enable_saml_success(harness_with_saml: Harness, monkeypatch: pytest.MonkeyPatch):
-    """
-    arrange: set mock container with file.
-    act: change the configuration file.
-    assert: new configuration file is pushed and SAML is enabled.
-    """
-    harness = harness_with_saml
+def test_enable_saml_success(monkeypatch: pytest.MonkeyPatch):
+    # Arrange: set up harness and container filesystem
+    harness = Harness(SynapseCharm)
+    harness.update_config({"server_name": TEST_SERVER_NAME})
+    relation_id = harness.add_relation("saml", "saml-integrator")
+    harness.add_relation_unit(relation_id, "saml-integrator/0")
+    harness.update_relation_data(relation_id, "saml-integrator", {
+        "entity_id": "https://login.staging.ubuntu.com",
+        "metadata_url": "https://login.staging.ubuntu.com/saml/metadata",
+    })
+    harness.begin()
     config_content = """
     listeners:
         - type: http
@@ -89,16 +94,12 @@ def test_enable_saml_success(harness_with_saml: Harness, monkeypatch: pytest.Mon
     monkeypatch.setattr(container_mock, "pull", pull_mock)
     monkeypatch.setattr(container_mock, "push", push_mock)
 
+    # Act: write the Synapse config file with SAML enabled
     synapse.enable_saml(container_mock, harness.charm._charm_state)
 
+    # Assert: ensure config file was written correctly
     assert pull_mock.call_args[0][0] == SYNAPSE_CONFIG_PATH
     assert push_mock.call_args[0][0] == SYNAPSE_CONFIG_PATH
-    saml_relation = harness.model.get_relation("saml")
-    assert saml_relation
-    saml_relation_data = harness.get_relation_data(saml_relation.id, "saml-integrator")
-    assert saml_relation_data
-    entity_id = saml_relation_data.get("entity_id")
-    metadata_url = saml_relation_data.get("metadata_url")
     expected_config_content = {
         "listeners": [
             {"type": "http", "x_forwarded": True, "port": 8080, "bind_addresses": ["::"]}
@@ -106,9 +107,9 @@ def test_enable_saml_success(harness_with_saml: Harness, monkeypatch: pytest.Mon
         "saml2_enabled": True,
         "saml2_config": {
             "sp_config": {
-                "metadata": {"remote": [{"url": metadata_url}]},
+                "metadata": {"remote": [{"url": "https://login.staging.ubuntu.com/saml/metadata"}]},
                 "service": {"sp": {"entityId": "https://server-name-configured.synapse.com"}},
-                "name": [entity_id, "en"],
+                "name": ["https://login.staging.ubuntu.com", "en"],
             },
             "user_mapping_provider": {
                 "config": {"mxid_source_attribute": "username", "mxid_mapping": "dotreplace"}
```

[Full code for modified test here.](https://pastebin.canonical.com/p/nHZYt7YBr4/). Still, I found the number of fixture layers here very hard to follow. To understand the setup for this function, we have to go through 3 layers of fixtures:

Side note: not at all related to this PR, but I find pytest's magical match-by-name fixtures hard to follow, especially when they're in a magically-imported file (conftest.py). At least we have ripgrep! :-)

```
test_enable_saml_success, which calls
harness_with_saml, which calls
harness_server_name_configured, which calls
harness_fixture
```

In my experience it's often better to cut through all that and just set up what you need in each test, like I've done above. Even if it's a bit more repetitive, it makes individual tests much easier to understand and modify without breaking other tests. Or you can use very targetted fixtures that set up a specific thing and "composing" them, rather than "inheriting" them like we did here.

Note also how in the modified version I've hard-coded the expected entity_id and metadata_url. I think this simplified and also avoids probing into the relation data (an implementation detail as far as this function is concerned).

One more thing. Now that we have @weiiwang01's excellent `Harness.get_filesystem_root`, I'd simplify further and avoid the mocking of push and pull. [Example of the test with get_filesystem_root here.](https://pastebin.canonical.com/p/229QXwpxS4/)

src/pebble.py
merkata: Good point, thanks @benhoyt ! @amandahla fyi in case you missed it

src/pebble.py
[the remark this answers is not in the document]
merkata: I'm curious about the "In general, Pebble doesn't monitor config files for changes" part - do you know whether this is by design or perhaps Pebble will at some point actively monitor for config changes and act?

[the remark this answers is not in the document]
amandahla: In this case the charm would have to do a replan and restart everytime, right? Or only the restart would update the plan too?
Synapse supports SIGHUP only for specific configurations as mentioned [here](https://github.com/matrix-org/synapse/issues/16071).

src/pebble.py
amandahla: Thanks, I changed it for a restart now :)

metadata.yaml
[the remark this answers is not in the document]
amandahla: It's OCI, changed :)

tests/unit/conftest.py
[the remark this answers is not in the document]
amandahla: Thank you for taking the time to write all these examples! I updated the test_enable_saml_success and the others will be changed in a follow-up PR.

tests/unit/test_synapse_workload.py
amandahla: Thanks, the test was changed now.

tests/integration/conftest.py
amandahla: I see. In this case, I changed the params to be constants then.

tests/unit/test_synapse_workload.py
```diff
@@ -67,55 +68,67 @@ def test_enable_metrics_error(monkeypatch: pytest.MonkeyPatch):
...
-
-    assert pull_mock.call_args[0][0] == SYNAPSE_CONFIG_PATH
-    assert push_mock.call_args[0][0] == SYNAPSE_CONFIG_PATH
-    saml_relation = harness.model.get_relation("saml")
-    assert saml_relation
-    saml_relation_data = harness.get_relation_data(saml_relation.id, "saml-integrator")
-    assert saml_relation_data
-    entity_id = saml_relation_data.get("entity_id")
-    metadata_url = saml_relation_data.get("metadata_url")
+    # This test was given as an example in this comment by Ben Hoyt.
```
jdkandersson, line 77: Does this need to be a comment in the test?

src/pebble.py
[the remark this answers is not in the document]
benhoyt: @merkata Re Pebble monitoring for config file changes. This is not so much "by design" as by it not having been designed yet. I wouldn't rule it out, though it's a non-trivial feature, so I wouldn't hold your breath. If you're interested, you could look at writing a spec for it and see if there's appetite for it?

@amandahla Yep, that's right - replan will restart only if there's a *plan* change.

tests/unit/test_synapse_workload.py
amandahla: Yes, at least until all the tests are refactored and this one is the reference. So anyone can see more details about why this one is different.

arturo-seijas (dismissed):
LGTM

yanksyoon (dismissed):
Some small questions, 🎉  LGTM!

requirements.txt
```diff
@@ -3,3 +3,4 @@ ops == 2.5.1
 pydantic ==1.10.12
 psycopg2-binary ==2.9.7
 requests == 2.31.0
+cosl == 0.0.6
```
arturo-seijas, line 6: Iirc, you just need this for testing. Charmcraft will pull it because it's present in the lib file

metadata.yaml
```diff
@@ -22,11 +22,17 @@ containers:
...
+    resource: synapse-nginx-image
 
 resources:
   synapse-image:
     type: oci-image
     description: OCI image for Synapse
+  synapse-nginx-image:
+    type: oci-image
+    description: OCI image for Synapse NGINX
+    auto-fetch: true
```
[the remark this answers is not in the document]
amandahla: Oh again I forgot this here! :| Thanks!

src/charm.py
```diff
@@ -49,23 +51,37 @@ def __init__(self, *args: typing.Any) -> None:
...
+        """Replan NGINX."""
+        container = self.unit.get_container(SYNAPSE_NGINX_CONTAINER_NAME)
+        if not container.can_connect():
+            self.unit.status = ops.MaintenanceStatus("Waiting for pebble")
+            return
+        self.model.unit.status = ops.MaintenanceStatus("Configuring Synapse NGINX")
+        try:
+            self.pebble_service.replan_nginx(container)
+        except PebbleServiceError as exc:
+            self.model.unit.status = ops.BlockedStatus(str(exc))
```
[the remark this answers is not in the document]
amandahla: Nice suggestion! Changed!

synapse_rock/attributemaps/u1staging.py
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
amandahla: Good catch, is not used. Removed :)

src/synapse/workload.py
```diff
@@ -96,6 +101,19 @@ def check_alive() -> ops.pebble.CheckDict:
...
+def check_nginx_ready() -> ops.pebble.CheckDict:
+    """Return the Synapse NGINX container check.
+
+    Returns:
+        Dict: check object converted to its dict representation.
+    """
+    check = Check(CHECK_READY_NAME)
+    check.override = "replace"
+    check.level = "ready"
+    check.tcp = {"port": SYNAPSE_NGINX_PORT}
```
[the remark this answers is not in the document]
amandahla: I think I didn't get it, could you give an example, please?

requirements.txt
amandahla: I'm trying to remember what happened that I needed to add the cosl here...I'll try to reproduce.

weiiwang01 (approved):

requirements.txt
amandahla: No idea. Removed, thanks.

amandahla: Found it after running a charmcraft clean

```
unit-teste1-0: 12:03:19 WARNING unit.teste1/0.install   File "/var/lib/juju/agents/unit-teste1-0/charm/./src/charm.py", line 22, in <module>
unit-teste1-0: 12:03:19 WARNING unit.teste1/0.install     from observability import Observability
unit-teste1-0: 12:03:19 WARNING unit.teste1/0.install   File "/var/lib/juju/agents/unit-teste1-0/charm/src/observability.py", line 8, in <module>
```
So I added again the cosl library to requirements

arturo-seijas: Doesn't the trace originate in a test?

amandahla: No, teste-1 is the name of the application. It originates from "MetricsEndpointProvider" [here](https://github.com/canonical/synapse-operator/blob/main/src/observability.py#L8tp), from prometheus-scrape libpath 39. In a follow-up PR, I'll update the library again so will get the PYDEPS.

metadata.yaml
```diff
@@ -42,7 +47,7 @@ provides:
 requires:
     ingress:
         interface: ingress
-        limit: 1
+        limit: 2
```
nrobinaubertin, line 50: Not that this is an issue but why would we need two ingresses ?

nrobinaubertin (approved):

metadata.yaml
amandahla: I increased to 2 because I was testing with Traefik and NGINX at the same time ;-)

weiiwang01 (approved):

merkata (approved):

src/synapse/workload.py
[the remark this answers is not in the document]
amandahla: I'll mark as resolved but feel free to suggest and then I can change in a followup PR. :)

PR #30, opened 2023-08-29
Add mjolnir support

amandahla:
### Overview
The [mjolnir](https://github.com/matrix-org/mjolnir) is a moderation tool for Matrix. It runs as a separate bot service that needs an admin user and a room. The room is the management room and mjolnir will only respond to commands originating from it.

The required steps to enable Mjolnir are:
- Get an admin access token.
- Check if the MJOLNIR_MEMBERSHIP_ROOM room is created.
    - Only users from there will be allowed to join the management room.
- Create Mjolnir user or get its access token if already exists.
- Create the management room or get its room id if already exists.
    - The management room will allow only members of MJOLNIR_MEMBERSHIP_ROOM room to join it.
- Make the Mjolnir user admin of this room.
- Create the Mjolnir configuration file.
- Override Mjolnir user rate limit.
- Finally, add Mjolnir pebble layer.

### Rationale
Community team needs a moderation tool to manage Matrix.

### Checklist
 - [x]  The [contributing guide](https://github.com/canonical/is-charms-contributing-guide) was applied
 - [x]  The changes are compliant with [ISD054 - Managing Charm Complexity](https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619)
 - [x]  The documentation is generated using src-docs
 - [ ]  The documentation for charmhub is updated.
 - [x]  The PR is tagged with appropriate label (urgent, trivial, complex)
 - [ ] The src-docs for this repo is still WIP. There is a WIP PR for it.


mjolnir_snap/snapcraft.yaml
```diff
@@ -0,0 +1,37 @@
...
+name: mjolnir
+summary: Mjolnir snap
+description: A moderation tool for Matrix
+version: "1.6.4"
+base: core22
+license: Apache-2.0
+confinement: strict
+apps:
+  mjolnir:
+    command: 'bin/node $SNAP/mjolnir/index.js'
```
merkata, line 13: Nit - on line 7 you have double quotes and here it is single quotes, could it be consistent?

src/charm.py
```diff
@@ -117,6 +123,55 @@ def _on_config_changed(self, event: ops.HookEvent) -> None:
...
+
+    def _enable_mjolnir(self) -> None:
+        """Enable mjolnir service."""
+        container = self.unit.get_container(SYNAPSE_CONTAINER_NAME)
+        if not container.can_connect():
+            self.unit.status = ops.MaintenanceStatus("Waiting for pebble")
+            return
+        self.model.unit.status = ops.MaintenanceStatus("Configuring Mjolnir")
+        try:
+            admin_username = token_hex(16)
```
merkata, line 137: Is 16 here the recommended value like described in [here](https://docs.python.org/3/library/secrets.html) - perhaps put a comment on the line?

src/charm.py
```diff
@@ -117,6 +123,55 @@ def _on_config_changed(self, event: ops.HookEvent) -> None:
         """
         self.change_config(event)
         self._set_workload_version()
+        if self._charm_state.enable_mjolnir:
+            self._enable_mjolnir()
+
+    def _enable_mjolnir(self) -> None:
+        """Enable mjolnir service."""
```
merkata, line 130: Can you expand the description - I see below a (new) admin user is registered to create a token that is than passed when registering the new mjolnir user. Would be good to have the flow described in the docstring.

src/charm.py
```diff
@@ -117,6 +123,55 @@ def _on_config_changed(self, event: ops.HookEvent) -> None:
...
+            mjolnir_access_token = mjolnir_user.access_token
+            room_id = synapse.get_room_id(room_name="management", access_token=admin_access_token)
+            synapse.create_mjolnir_config(
+                container=container, access_token=mjolnir_access_token, room_id=room_id
+            )
+            synapse.override_rate_limit(
+                user=mjolnir_user, access_token=admin_access_token, charm_state=self._charm_state
+            )
+            synapse.deactivate_user(
+                user=admin_user,
```
merkata, line 158: Might discuss variable naming - the admin user here is just a temporary one used to bootstrap mjolnir I believe, so the variables can reflect that imo. Do unregistered users persist somewhere, do you know? Wondering whether that user will be left somewhere (not a big deal with one bot, was just thinking).

src/synapse/api.py
```diff
@@ -223,3 +261,133 @@ def get_version() -> str:
             f"server_version has unexpected content: {server_version}"
         )
     return version_match.group(1)
+
+
+def get_access_token(user: User, server: str, access_token: str) -> str:
+    """Get access token from a user.
+
+    Args:
+        user: user to get access token from.
```
merkata, line 270: Is it from or for?

merkata (dismissed):
Small nits, otherwise looks really good!

mjolnir_snap/snapcraft.yaml
```diff
@@ -0,0 +1,37 @@
...
+    install-mjolnir:
+        after: [node]
+        plugin: nil
+        source: https://github.com/matrix-org/mjolnir.git
+        source-tag: v1.6.4
+        source-type: git
+        override-build: |
+            npm install -g yarn
+            yarn install
+            yarn build
```
gregory-schiano, line 33: Can't the npm plugin be used to do this ?

src/charm.py
```diff
@@ -117,6 +123,55 @@ def _on_config_changed(self, event: ops.HookEvent) -> None:
         """
         self.change_config(event)
         self._set_workload_version()
+        if self._charm_state.enable_mjolnir:
+            self._enable_mjolnir()
```
gregory-schiano, line 127: Shouldn't it be smarter ? What if mjolnir was already enabled, you'll systematically call this method which creates a new user, ...
You should only call this method if mjolnir is not already enabled (probably could be deducted from pebble plan service state / scheduling)

src/charm.py
```diff
@@ -117,6 +123,55 @@ def _on_config_changed(self, event: ops.HookEvent) -> None:
...
+                container,
+                MJOLNIR_USER,
+                True,
+                str(self._charm_state.server_name),
+                admin_access_token,
+            )
+            # Create (or get) the management room
+            # Add the bot to the management room if we are creating it
+            mjolnir_access_token = mjolnir_user.access_token
+            room_id = synapse.get_room_id(room_name="management", access_token=admin_access_token)
```
gregory-schiano, line 150: The room name should probably be a constant

src/charm.py
```diff
@@ -117,6 +123,55 @@ def _on_config_changed(self, event: ops.HookEvent) -> None:
...
+    def _enable_mjolnir(self) -> None:
+        """Enable mjolnir service."""
+        container = self.unit.get_container(SYNAPSE_CONTAINER_NAME)
+        if not container.can_connect():
+            self.unit.status = ops.MaintenanceStatus("Waiting for pebble")
+            return
+        self.model.unit.status = ops.MaintenanceStatus("Configuring Mjolnir")
+        try:
+            admin_username = token_hex(16)
+            admin_user = actions.register_user(container, admin_username, True)
```
gregory-schiano, line 138: I'm wondering if this user shouldn't be systematically created when the charm is start and kept for the whole lifecycle of the application.
It's very likely that we'll need that user for many actions, and creation/disabling a user systematically is probably not the best approach.

src/charm.py
```diff
@@ -117,6 +123,55 @@ def _on_config_changed(self, event: ops.HookEvent) -> None:
...
+            logger.exception("Failed to interact with Synapse workload: %r", exc)
+            self.model.unit.status = ops.BlockedStatus(str(exc))
+            return
+        except synapse.APIError as exc:
+            logger.exception("Failed to interact with Synapse API: %r", exc)
+            self.model.unit.status = ops.BlockedStatus(str(exc))
+            return
+        except actions.RegisterUserError as exc:
+            logger.exception("Failed to register Mjolnir user: %r", exc)
+            self.model.unit.status = ops.BlockedStatus(str(exc))
```
gregory-schiano, line 173: When a charm is in blocked state we expect an action from the user to solve the problem.
If we don't expect any action from the user, and it's a temporary error, don't put in blocked state and either retry or expect another event to retry that action
If we don't expect any action from the user, and the error is unrecoverable, it's preferable to let the exception go through
If we expect an action from the user, it's probably better to state what they should do in the BlockedStatus message

src/synapse/api.py
```diff
@@ -223,3 +261,133 @@ def get_version() -> str:
...
+        res = requests.post(
+            f"{DEACTIVATE_ACCOUNT_URL}/{user_id}", headers=headers, json=data, timeout=5
+        )
+        res.raise_for_status()
+    except (requests.exceptions.ConnectionError, requests.exceptions.Timeout) as exc:
+        logger.exception("Failed to connect to %s: %r", DEACTIVATE_ACCOUNT_URL, exc)
+        raise NetworkError(f"Failed to connect to {DEACTIVATE_ACCOUNT_URL}.") from exc
+    except requests.exceptions.HTTPError as exc:
+        logger.exception("HTTP error from %s: %r", DEACTIVATE_ACCOUNT_URL, exc)
+        raise NetworkError(f"HTTP error from {DEACTIVATE_ACCOUNT_URL}.") from exc
```
gregory-schiano, line 393: This is repeated several time, I'm wondering if you couldn't write a decorator to help there, or make a generic method that wraps requests.request (https://requests.readthedocs.io/en/latest/api/#requests.request)

src/synapse/workload.py
```diff
@@ -167,6 +195,66 @@ def enable_metrics(container: ops.Container) -> None:
...
+            enabled: false
+            port: {MJOLNIR_HEALTH_PORT}
+            address: "0.0.0.0"
+            endpoint: "/healthz"
+            healthyStatus: 200
+            unhealthyStatus: 418
+        pollReports: false
+        displayReports: false
+        """
+    config = yaml.safe_load(default_content)
```
gregory-schiano, line 232: Shouldn't we either use a dataclass that represent the configuration, or a template to avoid such "hardcoded" configuration ?

src/charm.py
amandahla: I've just realized while testing that if we deactivate the account used to get the access token, the access token is deleted as well so this internal admin should be kept. How can we store its password?

src/charm.py
amandahla: @gregory-schiano mentioned in one of his comments that we should consider creating an admin user systematically. At first I thought it could be temporary (I didn't want to add the complexity to keep a user/password) but then I realized that the access token created by the temporary user was deleted as well, making Mjolnir throw an error.

amandahla: I'll mark this one and the next regarding admin user in favor of this comment https://github.com/canonical/synapse-operator/pull/30#discussion_r1315842557 so we can discuss there the best approach. But feel free to reopen-it.

src/charm.py
amandahla: Yes, they persist. When passing the erase parameter, most of the data are erased but some things are left behind and the user is marked as "Erased". I already gave up on this approach :( More discussion about the user here:
https://github.com/canonical/synapse-operator/pull/30#discussion_r1315842557
I'll mark this one as resolved but feel free to re-open it.

src/charm.py
amandahla: Sure! :)

src/synapse/api.py
```diff
@@ -17,13 +17,18 @@
...
 from user import User
 
 logger = logging.getLogger(__name__)
 
-SYNAPSE_URL = "http://localhost:8008"
 REGISTER_URL = f"{SYNAPSE_URL}/_synapse/admin/v1/register"
 VERSION_URL = f"{SYNAPSE_URL}/_synapse/admin/v1/server_version"
+LOGIN_URL = f"{SYNAPSE_URL}/_synapse/admin/v1/users"
+LIST_ROOMS_URL = f"{SYNAPSE_URL}/_synapse/admin/v1/rooms"
+LIST_USERS_URL = f"{SYNAPSE_URL}/_synapse/admin/v2/users?from=0&limit=10&name="
```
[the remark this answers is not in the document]
amandahla: Good point. I didn't find a proper explanation but it seems that they are migrating step by step. So, from the docs latest version:
[List Room API](https://matrix-org.github.io/synapse/latest/admin_api/rooms.html#list-room-api) - v1
[List Accounts API](https://matrix-org.github.io/synapse/latest/admin_api/user_admin_api.html#list-accounts) - v2

src/synapse/api.py
amandahla: From. This is like the admin user is making a login as another user so it will get an access token from this other user.

mjolnir_snap/snapcraft.yaml
amandahla: Since the snap command is "bin/node $SNAP/mjolnir/index.js", I preferred to install the node and use it in the build.

src/charm.py
amandahla: This definitely makes life easier. I changed it to check if the pebble service is already there.

src/charm.py
amandahla: Done!

src/charm.py
amandahla: Nice. Now I changed to block only if the management room does not exist ( I'm considering for now that will be created by an admin user)

src/synapse/api.py
amandahla: Excellent suggestion, changed!

src/synapse/workload.py
amandahla: Something like Jenkins [here](https://github.com/canonical/jenkins-k8s-operator/blob/main/src/jenkins.py#L254)?

amandahla: Changed :)

src/synapse/api.py
amandahla: I changed the parameter name and added a description to the docstring to make it clear what happens.

templates/mjolnir_production.yaml
```diff
@@ -0,0 +1,23 @@
+dataPath: "/data/storage"
```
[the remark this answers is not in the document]
amandahla: This file should be ignored.

src/charm.py
gregory-schiano: I'd rather go for option 2
Either store it in peer relation data, or on the shared storage between the charm and the workload (don't remember where this is mounted)

Option 1 will keep creating admin users, this is not great

amandahla: Agree, changed.

jdkandersson:
For these kind of reviews, running `src-docs` would be quite helpful to get an overview of the interface changes please

mjolnir_snap/snapcraft.yaml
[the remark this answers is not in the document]
amandahla: We are discussing the idea of sharing this with Mjolnir community actually.It's a follow-up action after this is merged/working.

src/actions/register_user.py
```diff
@@ -34,13 +34,22 @@ def __init__(self, msg: str):
...
 
 
-def register_user(container: ops.Container, username: str, admin: bool) -> User:
+# access_token is not a password
+def register_user(
+    container: ops.Container,
+    username: str,
+    admin: bool,
+    server: str = "",
+    admin_access_token: str = "",  # nosec
```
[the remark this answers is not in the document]
amandahla: Passing as string will make it compatible with the following calls and will raise an error on the request as expected.

src/charm.py
```diff
@@ -117,6 +125,123 @@ def _on_config_changed(self, event: ops.HookEvent) -> None:
...
+    def _has_secrets(self) -> bool:
+        """Check if current Juju version supports secrets.
+
+        Returns:
+            If secrets are supported or not.
+        """
+        juju_version = JujuVersion.from_environ()
+        # Because we're only using secrets in a peer relation we don't need to
+        # check if the other end of a relation also supports secrets.
+        return juju_version.has_secrets
```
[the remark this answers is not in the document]
amandahla: I re-used the code from here:
https://github.com/canonical/indico-operator/pull/211
I think having a _has_secrets it's easier to understand and it could be changed if there is no JujuVersion.from_envriron anymore.

src/charm.py
```diff
@@ -117,6 +125,123 @@ def _on_config_changed(self, event: ops.HookEvent) -> None:
...
+            and not peer_relation.data[self.app].get(SECRET_KEY)
+        ):
+            peer_relation.data[self.app].update({SECRET_KEY: admin_user.access_token})
+        elif (
+            peer_relation
+            and self._has_secrets()
+            and not peer_relation.data[self.app].get(SECRET_ID)
+        ):
+            secret = self.app.add_secret({SECRET_KEY: admin_user.access_token})
+            peer_relation.data[self.app].update({SECRET_ID: secret.id})
```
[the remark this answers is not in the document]
amandahla: Same thing, copied from here:
https://github.com/canonical/indico-operator/pull/211 ;-)
Indico is one of my favorite references.

Comment added, also some logger.debug

src/charm.py
```diff
@@ -117,6 +125,123 @@ def _on_config_changed(self, event: ops.HookEvent) -> None:
...
+        if not self._has_secrets():
+            secret_value = peer_relation.data[self.app].get(SECRET_KEY)
+        else:
+            secret_id = peer_relation.data[self.app].get(SECRET_ID)
+            if secret_id:
+                secret = self.model.get_secret(id=secret_id)
+                secret_value = secret.get_content().get(SECRET_KEY)
+        return secret_value
+
+    def _enable_mjolnir(self) -> None:
```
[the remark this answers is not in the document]
amandahla: It's a good idea, I'll add it to the roadmap to discuss how to handle this since it interacts with many parts of the charm.

src/charm.py
```diff
@@ -147,7 +147,11 @@ def _on_leader_elected(self, _: ops.LeaderElectedEvent) -> None:
         if not container.can_connect():
             self.unit.status = ops.MaintenanceStatus("Waiting for pebble")
             return
-        admin_user = actions.register_user(container, SYNAPSE_OPERATOR_USER, True)
+        # Using 16 to create a random value but to  be secure against brute-force attacks, please
+        # check the docs:
+        # https://docs.python.org/3/library/secrets.html#how-many-bytes-should-tokens-use
+        username = token_hex(16)
+        admin_user = actions.register_user(container, username, True)
```
arturo-seijas, line 154: Why a random username?

config.yaml
```diff
@@ -19,3 +19,9 @@ options:
       The public-facing base URL that clients use to access this Homeserver.
       Defaults to https://<server_name>/. Only used if there is integration with
       SAML integrator charm.
+  enable_mjolnir:
```
arturo-seijas, line 22: Thinking about our use case, would it make sense to have it enabled by default?

src/charm.py
amandahla: Kind of a egg-chicken problem.
If the user exists, register_user will try to get the access_token. To do that it needs an admin user. So, to be on the safe side, I decided to use a random user name. So, if for any reason there is no access token on peer data/secret, another user will be created.

src/charm.py
```diff
@@ -115,16 +124,143 @@ def _on_config_changed(self, event: ops.HookEvent) -> None:
...
+            synapse.get_room_id(
+                room_name=MJOLNIR_MEMBERSHIP_ROOM, admin_access_token=admin_access_token
+            )
+        except synapse.RoomNotFoundError:
+            logger.info("Room %s not found, waiting for user action", MJOLNIR_MEMBERSHIP_ROOM)
+            self.model.unit.status = ops.BlockedStatus(
+                f"{MJOLNIR_MEMBERSHIP_ROOM} not found and "
+                "is required by Mjolnir. Please, create it."
+            )
+            event.defer()
```
arturo-seijas, line 227: Do we really need to defer this event here?

src/synapse/api.py
```diff
@@ -223,3 +234,225 @@ def get_version() -> str:
...
+                "%s rooms named with term %s found, checking for exact name.", total, room_name
+            )
+        for room in rooms:
+            if str(room["name"]).upper() == room_name.upper():
+                return room["room_id"]
+    except (requests.exceptions.JSONDecodeError, TypeError, KeyError) as exc:
+        logger.exception("Failed to decode rooms: %r. Received: %s", exc, res.text)
+        raise GetRoomIDError(str(exc)) from exc
+
+    raise RoomNotFoundError(f"Room {room_name} not found. Please, verify if the room exists.")
```
arturo-seijas, line 320: Why not returning None instead?

src/charm.py
```diff
@@ -115,16 +124,143 @@ def _on_config_changed(self, event: ops.HookEvent) -> None:
...
+            return
+        # Not checking if the pebble service is enabled to skip this
+        # in case there is a charm update that changes Mjolnir configuration
+        self.model.unit.status = ops.MaintenanceStatus("Configuring Mjolnir")
+        admin_access_token = self._get_admin_access_token()
+        try:
+            synapse.get_room_id(
+                room_name=MJOLNIR_MEMBERSHIP_ROOM, admin_access_token=admin_access_token
+            )
+        except synapse.RoomNotFoundError:
```
arturo-seijas, line 221: It seems like this is part of the normal flow, so I'd advice against using exceptions to control it. If you returned a None value instead, you could use an `if`

src/synapse/api.py
```diff
@@ -48,26 +57,61 @@ class NetworkError(APIError):
...
-    """Exception raised when getting version via API fails."""
+    """Exception raised when getting version fails."""
 
 
 class VersionUnexpectedContentError(GetVersionError):
     """Exception raised when output of getting version is unexpected."""
 
 
-def register_user(registration_shared_secret: str, user: User) -> None:
+class GetRoomIDError(APIError):
```
arturo-seijas, line 71: I'm not seeing where these specific exceptions are leveraged to justify not raising an `APIError` instead and simplify the code

src/synapse/workload.py
```diff
@@ -57,6 +63,14 @@ class EnableMetricsError(WorkloadError):
     """Exception raised when something goes wrong while enabling metrics."""
 
 
+class EnableMjolnirError(WorkloadError):
```
arturo-seijas, line 66: Is this used?

src/charm.py
```diff
@@ -115,16 +124,143 @@ def _on_config_changed(self, event: ops.HookEvent) -> None:
...
+        container = self.unit.get_container(SYNAPSE_CONTAINER_NAME)
+        if not container.can_connect():
+            self.unit.status = ops.MaintenanceStatus("Waiting for pebble")
+            return
+        # Not checking if the pebble service is enabled to skip this
+        # in case there is a charm update that changes Mjolnir configuration
+        self.model.unit.status = ops.MaintenanceStatus("Configuring Mjolnir")
+        admin_access_token = self._get_admin_access_token()
+        try:
+            synapse.get_room_id(
```
arturo-seijas, line 218: Perhaps this logic would be better placed in the `synapse`package

src/constants.py
```diff
@@ -5,14 +5,23 @@
...
 CHECK_ALIVE_NAME = "synapse-alive"
+CHECK_NGINX_READY_NAME = "synapse-nginx-ready"
+CHECK_MJOLNIR_READY_NAME = "synapse-mjolnir-ready"
 COMMAND_MIGRATE_CONFIG = "migrate_config"
 PROMETHEUS_TARGET_PORT = "9000"
 SYNAPSE_CONFIG_DIR = "/data"
 SYNAPSE_CONFIG_PATH = f"{SYNAPSE_CONFIG_DIR}/homeserver.yaml"
+MJOLNIR_CONFIG_PATH = f"{SYNAPSE_CONFIG_DIR}/config/production.yaml"
+MJOLNIR_SERVICE_NAME = "mjolnir"
+MJOLNIR_USER = "mjolnir"
```
arturo-seijas, line 16: Can you sort these alphabetically?

src/constants.py
```diff
@@ -5,14 +5,23 @@
 
 CHECK_READY_NAME = "synapse-ready"
 CHECK_ALIVE_NAME = "synapse-alive"
+CHECK_NGINX_READY_NAME = "synapse-nginx-ready"
```
arturo-seijas, line 8: I think the constants defined here that are only used in pebble.py would be better defined there

src/charm.py
arturo-seijas: Thanks for the explanation :) Would you mind adding a comment exaplining why? I'm a bit concerned about someone removing this because they lack the reasoning behind it

config.yaml
amandahla: It's a good suggestion but I would say no because Mjolnir is a tool that can be run externally. So, enabling this one inside the charm might cause conflicts.

src/charm.py
amandahla: See the discussion in our channel ;-)

src/synapse/api.py
amandahla: Hmm...good point. I don't have an answer for that, I was following the flow of creating exceptions, perhaps.
I'll change it.

src/charm.py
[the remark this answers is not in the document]
amandahla: At first, I answered that I preferred the _has_secrets in case of JujuVersion changes but then while working on Mjolnir module + collect_unit_status I realized that your suggestion is better since I don't need to import anything from the charm itself. So: changed :)

src/charm.py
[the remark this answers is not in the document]
amandahla: I'll close this because now its outdated, moved to Mjolnir module.

src/charm.py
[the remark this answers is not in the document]
amandahla: At first, my idea was to work on this in a follow-up PR but then after the discussion on MM and Gregory's suggestion to use the [Collect Status event](https://ops.readthedocs.io/en/latest/#ops.CollectStatusEvent), I had to change this now.So: changed too.

src/charm.py
```diff
@@ -117,6 +125,123 @@ def _on_config_changed(self, event: ops.HookEvent) -> None:
...
+
+    def _get_admin_access_token(self) -> str:
+        """Get admin access token.
+
+        Returns:
+            admin access token.
+        """
+        peer_relation = self.model.get_relation(PEER_RELATION_NAME)
+        assert peer_relation is not None  # nosec
+        if not self._has_secrets():
```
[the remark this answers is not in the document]
amandahla: Since this is outdated, I'll close this one. The code was moved.

src/synapse/api.py
amandahla: Changed, thanks :)

src/charm.py
amandahla: Perfect, changed.

src/charm.py
amandahla: David suggested something similar, it was moved to a Mjolnir module now.

src/actions/register_user.py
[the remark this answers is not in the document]
amandahla: Good catch, changed.

src/synapse/api.py
amandahla: I don't remember if this was recommended in one previous review to be honest :-( 
@gregory-schiano @jdkandersson suggestions? Can we use APIError?

src/constants.py
amandahla: They are used in the workload too :)

src/synapse/workload.py
amandahla: Good catch, no. Removed.

src/synapse/api.py
```diff
@@ -17,13 +17,22 @@
...
 from user import User
 
 logger = logging.getLogger(__name__)
 
-SYNAPSE_URL = "http://localhost:8008"
+# The API version that should be used is described in the documentation:
+# https://matrix-org.github.io/synapse/latest/usage/administration/index.html
 REGISTER_URL = f"{SYNAPSE_URL}/_synapse/admin/v1/register"
 VERSION_URL = f"{SYNAPSE_URL}/_synapse/admin/v1/server_version"
+LOGIN_URL = f"{SYNAPSE_URL}/_synapse/admin/v1/users"
```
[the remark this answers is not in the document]
amandahla: Sure, changed.

src/constants.py
amandahla: Sure, changed.

src/charm.py
amandahla: Sure, thanks. Added.

src/synapse/api.py
jdkandersson: It depends on if it is useful to distinguish between APIError and this particular error during the handling of the exception

src/constants.py
arturo-seijas: Ups, my bad

src/synapse/api.py
amandahla: Since this is something that involves other contexts, I created a story (ISD-1070) so this can be addressed in a follow-up PR.

mjolnir_snap/snapcraft.yaml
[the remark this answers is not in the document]
amandahla: I'll close this one since there is no immediate action but feel free to reopen it.

src/charm.py
```diff
@@ -76,7 +64,7 @@ def __init__(self, *args: typing.Any) -> None:
             strip_prefix=True,
         )
         self._observability = Observability(self)
-        self.framework.observe(self.on.leader_elected, self._on_leader_elected)
+        self._mjolnir = Mjolnir(self, charm_state=self._charm_state)
```
jdkandersson, line 67: Something else that comes to mind, should we consider naming this differently? From someone that doesn't know much about Synapse the `Mjolnir` name doesn't tell them that this is about moderation. Perhaps `MjolnirModeration` or something like it?

src/constants.py
```diff
@@ -3,25 +3,27 @@
...
-MJOLNIR_SERVICE_NAME = "mjolnir"
-MJOLNIR_USER = "mjolnir"
+MJOLNIR_HEALTH_PORT = 7777
 MJOLNIR_MANAGEMENT_ROOM = "management"
 MJOLNIR_MEMBERSHIP_ROOM = "moderators"
-MJOLNIR_HEALTH_PORT = 7777
+MJOLNIR_SERVICE_NAME = "mjolnir"
+PEER_RELATION_NAME = "synapse-peers"
+PROMETHEUS_TARGET_PORT = "9000"
+SECRET_ID = "secret-id"  # nosec
```
jdkandersson, line 19: We have a standard for explaining any linting disable rules, what is the reason for disabling in this case?

src/mjolnir.py
jdkandersson, line 1: Nice!

src/mjolnir.py
```diff
@@ -0,0 +1,220 @@
...
+        """Initialize a new instance of the Mjolnir class.
+
+        Args:
+            charm: The charm object that the Mjolnir instance belongs to.
+            charm_state: Instance of CharmState.
+        """
+        super().__init__(charm, "mjolnir")
+        self._charm = charm
+        self._charm_state = charm_state
+        self.framework.observe(charm.on.collect_unit_status, self._on_collect_status)
```
jdkandersson, line 45: Should this be done on the charm rather than here?

jdkandersson (commented):
I'll leave the review here for now since we need to decide whether `Mjolnir` is an observer or a business logic module - based on my initial review it looks like it should be a business logic module

src/charm.py
amandahla: What if I add a comment explaining what is Mjolnir? And more description to the class as well.
I'm afraid that changing it to MjolnirModeration might give the idea that it is something that will moderate "Mjolnir".

src/mjolnir.py
```diff
@@ -0,0 +1,220 @@
...
+
+logger = logging.getLogger(__name__)
+
+MJOLNIR_USER = "mjolnir"
+
+
+class Mjolnir(ops.Object):  # pylint: disable=too-few-public-methods
+    """A class representing the Mjolnir plugin for Synapse application."""
+
+    def __init__(self, charm: ops.CharmBase, charm_state: CharmState):
```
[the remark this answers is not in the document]
amandahla: The charm because Mjolnir uses the peer data/secrets to store an admin access token that is used to further actions.
The charm_state because it checks if enable_mjolnir configuration is set in order to configure the plugin or not.

src/mjolnir.py
```diff
@@ -0,0 +1,220 @@
...
+    PEER_RELATION_NAME,
+    SECRET_ID,
+    SECRET_KEY,
+    SYNAPSE_CONTAINER_NAME,
+)
+from user import User
+
+logger = logging.getLogger(__name__)
+
+MJOLNIR_USER = "mjolnir"
```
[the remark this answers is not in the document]
amandahla: Actually not, changed.

src/constants.py
amandahla: Thanks for pointing that out, added!

src/mjolnir.py
amandahla: Only Mjolnir uses the collect_unit_status in order to check when the moderator room, a requirement for configuring the plugin, is created. This was used to avoid using defer in config-changed event. This was discussed on our MM channel, I can DM you the link to the thread :)

src/charm.py
jdkandersson: comment would be nice

src/mjolnir.py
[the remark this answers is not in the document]
jdkandersson: @gregory-schiano this is an interesting case, I think the peer data/ secret should come from the charm state because it is another source of data that the charm needs to function, much like relation data

src/charm.py
amandahla: Added :)

src/mjolnir.py
```diff
@@ -0,0 +1,225 @@
...
+        # So, to be on the safe side, the user name is randomly generated and if for any reason
+        # there is no access token on peer data/secret, another user will be created.
+        #
+        # Using 16 to create a random value but to  be secure against brute-force attacks, please
+        # check the docs:
+        # https://docs.python.org/3/library/secrets.html#how-many-bytes-should-tokens-use
+        username = token_hex(16)
+        return actions.register_user(container, username, True)
+
+    def _on_collect_status(self, event: ops.CollectStatusEvent) -> None:
```
gregory-schiano, line 113: I don't totally agree with the implementation of the collect_status hook here for few reasons:
- collect_status is a hook that can be implemented on multiple observers, and there's a priority logic implemented on ops side to choose the final status, therefore checking Synapse container status shouldn't be managed from here
- It shouldn't implement business logic, so I think that `_update_peer_data` and `enable_mjolnir` shouldn't be called from here, it's only collecting status, so check if service exists, check if there's a room_id, define the status and the message and do `add_status`

src/charm.py
```diff
@@ -63,6 +64,9 @@ def __init__(self, *args: typing.Any) -> None:
             strip_prefix=True,
         )
         self._observability = Observability(self)
+        # Mjolnir is a moderation tool for Matrix.
+        # See https://github.com/matrix-org/mjolnir/ for more details about it.
+        self._mjolnir = Mjolnir(self, charm_state=self._charm_state)
```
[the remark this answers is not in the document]
merkata: It's defaulting to false, so I believe we should.

src/mjolnir.py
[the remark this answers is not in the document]
merkata: Alright, just laying out my thoughts here so I can see whether I understand:
The mjolnir class now has both observer and business logic and we want to separate these. The class is rewritten to be an observer, the business logic lives separately in a service module (maybe for all future plugins as well?), the state is populated with the necessary data, so it lives in the state and the observer reconciles on changes. Is this correct?

merkata (approved):
I think this looks good. There is a discussion on where exactly to put some pieces, though I consider this a natural refactoring thing. Great work!

src/mjolnir.py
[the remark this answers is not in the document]
jdkandersson: correct

src/mjolnir.py
jdkandersson: Apologies I didn't get back on this, I think this is related to the conversation above so we can close this one for now and discuss there: https://github.com/canonical/synapse-operator/pull/30#discussion_r1324021010

src/mjolnir.py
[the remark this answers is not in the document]
amandahla: In order to provide this feature to the Community team, I suggest that we work on this change in a follow-up PR. WDYT?

IMHO since it's more related to an architecture definition, in a follow-up PR we could have more discussion and it would be easier for other team members to see the changes and discuss them as well.

src/mjolnir.py
amandahla: This definitely is something interesting to be discussed.
I think it could be applied [here](https://github.com/canonical/synapse-operator/pull/30#discussion_r1330005594) as well. What do you think?

src/mjolnir.py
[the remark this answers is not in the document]
jdkandersson: ok sounds good

jdkandersson (approved):

merkata (approved):

src/mjolnir.py
amandahla: I'll mark this one as resolved to be discussed in a follow-up PR.

PR #36, opened 2023-09-13
refactor unit test : update test_enable_metrics_success

Thanhphan1147:
### Overview

Relates to #19, rewrite unit tests to :
* Directly set up the test charm in the test function
* Write files to the charm's containers directly instead of using push/pull mocks

### Rationale

This makes the test functions easier to understand, avoid complicated fixture dependencies

### Juju Events Changes

N/A

### Module Changes

N/A

### Library Changes

N/A

### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -161,3 +161,6 @@ cython_debug/
 # Juju and Rock files
 *.charm
 */*.rock
+
+# Vscode files
+.vscode/
--- a/tests/unit/test_synapse_workload.py
+++ b/tests/unit/test_synapse_workload.py
@@ -18,11 +18,11 @@ from charm import SynapseCharm
 from constants import SYNAPSE_CONFIG_PATH, SYNAPSE_CONTAINER_NAME, TEST_SERVER_NAME
 
 
-def test_enable_metrics_success(monkeypatch: pytest.MonkeyPatch):
+def test_enable_metrics_success():
     """
-    arrange: set mock container with file.
+    arrange: Write config to container.
     act: change the configuration file.
-    assert: new configuration file is pushed and metrics are enabled.
+    assert: new configuration file is present and metrics are enabled.
     """
     config_content = """
     listeners:
@@ -31,17 +31,23 @@ def test_enable_metrics_success(monkeypatch: pytest.MonkeyPatch):
           bind_addresses:
             - "::"
     """
-    text_io_mock = io.StringIO(config_content)
-    pull_mock = Mock(return_value=text_io_mock)
-    push_mock = MagicMock()
-    container_mock = MagicMock()
-    monkeypatch.setattr(container_mock, "pull", pull_mock)
-    monkeypatch.setattr(container_mock, "push", push_mock)
 
-    synapse.enable_metrics(container_mock)
+    # Setup charm
+    harness = Harness(SynapseCharm)
+    harness.update_config({"server_name": TEST_SERVER_NAME, "public_baseurl": TEST_SERVER_NAME})
+    harness.set_can_connect(SYNAPSE_CONTAINER_NAME, True)
+    harness.begin()
 
-    assert pull_mock.call_args[0][0] == SYNAPSE_CONFIG_PATH
-    assert push_mock.call_args[0][0] == SYNAPSE_CONFIG_PATH
+    # write config
+    root = harness.get_filesystem_root(SYNAPSE_CONTAINER_NAME)
+    config_path = root / SYNAPSE_CONFIG_PATH[1:]
+    config_path.parent.mkdir(parents=True)
+    config_path.write_text(config_content)
+
+    # Enable metrics in container
+    synapse_container = harness.model.unit.get_container(SYNAPSE_CONTAINER_NAME)
+    synapse.enable_metrics(synapse_container)
+
     expected_config_content = {
         "listeners": [
             {"type": "http", "port": 8080, "bind_addresses": ["::"]},
@@ -49,7 +55,8 @@ def test_enable_metrics_success(monkeypatch: pytest.MonkeyPatch):
         ],
         "enable_metrics": True,
     }
-    assert push_mock.call_args[0][1] == yaml.safe_dump(expected_config_content)
+
+    assert config_path.read_text() == yaml.safe_dump(expected_config_content)
 
 
 def test_enable_metrics_error(monkeypatch: pytest.MonkeyPatch):


.gitignore
```diff
@@ -161,3 +161,6 @@ cython_debug/
 # Juju and Rock files
 *.charm
 */*.rock
+
+# Vscode files
+.vscode/
```
jdkandersson, line 166: Final new line please?

Thanhphan1147: thanks, I added the missing line

src/constants.py
```diff
@@ -16,3 +16,4 @@
 SYNAPSE_NGINX_PORT = 8080
 SYNAPSE_SERVICE_NAME = "synapse"
 TEST_SERVER_NAME = "server-name-configured.synapse.com"
```
arturo-seijas, line 18: I'd rather have these constants defined in a module located under the `tests` directory as they are not part of the sources

tests/unit/test_charm.py
```diff
@@ -12,24 +12,32 @@
...
-def test_synapse_pebble_layer(harness_server_name_configured: Harness) -> None:
+def test_synapse_pebble_layer(harness: Harness, monkeypatch: pytest.MonkeyPatch) -> None:
     """
     arrange: charm deployed.
     act: start the Synapse charm, set Synapse container to be ready and set server_name.
     assert: Synapse charm should submit the correct Synapse pebble layer to pebble.
     """
-    harness = harness_server_name_configured
+    monkeypatch.setattr(synapse, "get_version", lambda *_args, **_kwargs: "")
+    harness.update_config({"server_name": TEST_SERVER_NAME})
```
arturo-seijas, line 37: Is this executed for every tests and, if so, have you considered moving it to the fixture?

tests/unit/test_database.py
```diff
@@ -15,19 +15,35 @@
...
+) -> None:
     """
     arrange: start the Synapse charm, set Synapse container to be ready and set server_name.
     act: add database relation and erase database.
     assert: erase query is executed.
     """
-    harness = harness_with_postgresql
+    harness = Harness(SynapseCharm)
+    harness.update_config({"server_name": TEST_SERVER_NAME})
+    relation_id = harness.add_relation("database", "postgresql")
```
arturo-seijas, line 35: Iirc there's a way now to do this in a single line

tests/unit/test_database.py
```diff
@@ -52,17 +68,31 @@ def test_erase_database(harness_with_postgresql: Harness, monkeypatch: pytest.Mo
...
     assert: exception is raised.
     """
-    harness = harness_with_postgresql
+    harness = Harness(SynapseCharm)
+    harness.update_config({"server_name": TEST_SERVER_NAME})
+    relation_id = harness.add_relation("database", "postgresql")
+    harness.add_relation_unit(relation_id, "postgresql/0")
+    harness.update_relation_data(
+        relation_id,
+        "postgresql",
```
arturo-seijas, line 88: The config data seems to be the same across the tests, can you extract this somewhere else?

tests/unit/test_reset_instance_action.py
```diff
@@ -3,7 +3,7 @@
 
 """Reset instance action unit tests."""
 
-# pylint: disable=protected-access
+# pylint: disable=protected-access, no-member
```
arturo-seijas, line 6: Which is the problematic line here?

Thanhphan1147: They are these 2 lines : 
- https://github.com/canonical/synapse-operator/blob/c5fa04fa044ca89e11b0513861ae7d6d1432eb05/tests/unit/test_reset_instance_action.py#L54
- https://github.com/canonical/synapse-operator/blob/c5fa04fa044ca89e11b0513861ae7d6d1432eb05/tests/unit/test_reset_instance_action.py#L155

We have used `call_args` in tests in the past but I don't really know why this lint error is only now showing

tests/unit/test_charm.py
Thanhphan1147: Good point, I was hesitating between this and moving it to a fixture because keeping it here I think gives a better idea of what we are doing in the tests. But I'll move it to a fixture.

Thanhphan1147: I have made the changes here : https://github.com/canonical/synapse-operator/pull/36/commits/c331ecb1a5a485378059e08ebc76026e9c7e7c78

tests/unit/test_database.py
Thanhphan1147: Thanks, I have made the changes here : https://github.com/canonical/synapse-operator/pull/36/commits/c331ecb1a5a485378059e08ebc76026e9c7e7c78

tests/unit/test_database.py
Thanhphan1147: I have made the changes here : https://github.com/canonical/synapse-operator/pull/36/commits/c331ecb1a5a485378059e08ebc76026e9c7e7c78

tests/constants.py
```diff
@@ -0,0 +1,7 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""This module defines Test-related constants"""
+
+TEST_SERVER_NAME = "server-name-configured.synapse.com"
+TEST_SERVER_NAME_CHANGED = "pebble-layer-1.synapse.com"
```
jdkandersson, line 7: Consider whether these could be part of a fixture

tests/unit/test_charm.py
jdkandersson: Maybe Arturo was referring to the configuration of setting the server name? That could be a fixture as well since it seems to be done a lot

src/constants.py
Thanhphan1147: I have moved them

tests/unit/test_reset_instance_action.py
arturo-seijas: Perhaps the return value of  `call_args`?

tests/unit/test_database.py
```diff
@@ -96,41 +105,45 @@ def test_connect(
 
     db_client._connect()
 
+    relation_database_password = str(postgresql_relation_data.get("password"))
     query = (
         "dbname='synapse' user='user' host='myhost' "
-        f"password='{datasource_postgresql_password}' connect_timeout=5"
+        f"password='{relation_database_password}' connect_timeout=5"
```
amandahla, line 111: Suggestion: If I am not mistaken, this is the only test that asserts something that is set on relation data.
I think it would be easier to set the relation data similar to this [Discourse PR](https://github.com/canonical/discourse-k8s-operator/pull/105), even by using a constant for all the tests and only changing the password in this one. WDYT?

tests/unit/test_charm.py
Thanhphan1147: I might have missed that. I have also move it to a fixture and did an overall cleanup of those duplicates

tests/unit/test_reset_instance_action.py
Thanhphan1147: I have moved the pylint comments to the right place

tests/constants.py
Thanhphan1147: We also use these values in some of the tests I think, If I put this value directly in a fixture I'd have to find a way to retrieve this value in the test functions. 

What do you think?

tests/unit/test_database.py
Thanhphan1147: Good point, I have re-introduced the postgresql fixture and by default we don't set a password on relation data. We only explicitly set the password in the tests where we need it.

tests/unit/test_mjolnir.py
```diff
@@ -21,15 +21,16 @@
...
-    mock_juju_env, harness_server_name_configured: Harness, monkeypatch: pytest.MonkeyPatch
+    mock_juju_env, harness: Harness, monkeypatch: pytest.MonkeyPatch
 ) -> None:
     """
     arrange: start the Synapse charm, set server_name, mock container and create_admin_user.
     act: call _update_peer_data.
     assert: relation data is updated with access token.
     """
+    harness.update_config({"enable_mjolnir": True})
+    harness.begin_with_initial_hooks()
```
Thanhphan1147, line 32: @amandahla I have also adapt the mjolnir tests to align with the reworked fixtures. Can you take a quick look? Tests seem to run fine on my end.

src/constants.py
```diff
@@ -26,3 +26,4 @@
 SYNAPSE_NGINX_PORT = 8080
 SYNAPSE_PORT = 8008
 SYNAPSE_SERVICE_NAME = "synapse"
+SYNAPSE_URL = "http://localhost:8008"
```
jdkandersson, line 29: Do you think we should use the `SYNAPSE_PORT` variable here so that we don't duplicate the configuration? Also, is this synapse URL always on localhost? Or could it be different on production?

tests/unit/conftest.py
```diff
@@ -164,6 +164,37 @@ def start_cmd_handler(argv: list[str]) -> synapse.ExecResult:
     harness.cleanup()
 
 
+@pytest.fixture(name="harness_with_postgresql")
+def harness_with_postgresql_fixture() -> typing.Generator[Harness, None, None]:
```
jdkandersson, line 168: When naming fixtures it would be good to focus on specifically what is provided, e.g., in this case `postgresql`. The fact that it uses the harness seems to be an implementation detail. For fixtures it is also usually recommended to yield something specifically related to the fixture, e.g., something about `postgresql` in this case. The `harness` can always be retrieved from the original fixture

jdkandersson: Please also consider this for the other fixtures here

jdkandersson (dismissed):
Making good progress, keep it up!

src/constants.py
Thanhphan1147: Didn't notice this when I was aligning my code with main. It seems to be added alongside mjolnir. I'll ask @amandahla about this url

src/constants.py
```diff
@@ -27,4 +27,3 @@
 SYNAPSE_PORT = 8008
 SYNAPSE_SERVICE_NAME = "synapse"
```
arturo-seijas, line 28: `SYNAPSE_SERVICE_NAME` seems to be used only in pebble.py (and in the tests). Can you move it there?

Thanhphan1147: Good point, I'll refactor all of the constants like this :
- If it's only used in 1 file, I'll put it directly in that file
- since most of them are used in `workload` and `api` as you mentioned, and there's a dependency between them, if a constant is used in more than 2 places, I'll put it in `api` (I think it's better to put all imported stuffs in one place in case we need to change them later)

Then I'll remove this file and do the same for the test constants. What do you think? @arturo-seijas  / @amandahla

arturo-seijas: Sounds good to me

src/constants.py
Thanhphan1147: moved those constants and deleted the two constant files

src/constants.py
Thanhphan1147: moved those constants and deleted the two constant files

src/synapse/api.py
```diff
@@ -18,11 +18,31 @@
 from urllib3.util import Retry
 
 from charm_state import CharmState
-from constants import MJOLNIR_MANAGEMENT_ROOM, MJOLNIR_MEMBERSHIP_ROOM, SYNAPSE_URL
 from user import User
 
 logger = logging.getLogger(__name__)
 
+CHECK_ALIVE_NAME = "synapse-alive"
```
arturo-seijas, line 25: I think this is only used in  `pebble.py` can you move it there and make sure there are no more constants defined here that are only used in one file different from this?

src/synapse/api.py
```diff
@@ -18,11 +18,31 @@
...
+CHECK_MJOLNIR_READY_NAME = "synapse-mjolnir-ready"
+CHECK_NGINX_READY_NAME = "synapse-nginx-ready"
+CHECK_READY_NAME = "synapse-ready"
+COMMAND_MIGRATE_CONFIG = "migrate_config"
+SYNAPSE_CONFIG_DIR = "/data"
+MJOLNIR_CONFIG_PATH = f"{SYNAPSE_CONFIG_DIR}/config/production.yaml"
+MJOLNIR_MANAGEMENT_ROOM = "management"
+MJOLNIR_MEMBERSHIP_ROOM = "moderators"
+MJOLNIR_SERVICE_NAME = "mjolnir"
+MJOLNIR_HEALTH_PORT = 7777
```
arturo-seijas, line 35: Can you move to a file under the `tests` directory those constants that are only used for tests?

src/synapse/api.py
Thanhphan1147: I just checked and it's also used in `workflow.py` : https://github.com/canonical/synapse-operator/blob/221a25a0f91b0a62a642979a39b55e6165c47406/src/synapse/workload.py#L94 but maybe the function that is using it is not in the right place as it's also related to pebble.

What do you think about this? should I keep the constant in `api.py` or move it to `pebble.py` together with the `check_alive()` function?

I also did a scan of the constants added to `api.py` and made sure that they are all referenced in more than 2 files.

src/synapse/api.py
Thanhphan1147: I double-checked and most of those constants are used also in `mjolnir.py`, maybe they are also used to deploy mjolnir?.
 
I also made sure that tests-only constants are moved to `tests/unit/conftest.py`

tests/unit/test_synapse_workload.py
```diff
@@ -74,29 +71,15 @@ def test_enable_metrics_error(monkeypatch: pytest.MonkeyPatch):
...
-    harness.update_relation_data(
-        relation_id,
-        "saml-integrator",
-        {
-            "entity_id": "https://login.staging.ubuntu.com",
-            "metadata_url": metadata_url,
-        },
-    )
-    harness.set_can_connect(SYNAPSE_CONTAINER_NAME, True)
+    harness = harness_with_saml
```
amandahla, line 82: This test was supposed to be used as an example. Why it was changed?

tests/unit/conftest.py
Thanhphan1147: Thanks for the tip. I chose this naming because this fixture was independent from the base `harness` fixture. I have reworked this fixture to add postgres on top of the base fixture, moved it to `test_database.py` which is the only place where postgres is tested, and also set it to `autouse`.

tests/unit/test_synapse_workload.py
Thanhphan1147: I reverted my changes.

src/synapse/api.py
arturo-seijas: If those are used only in `mjolnir.py` and test files, can yo move them to `mjolnir.py`?

src/synapse/api.py
arturo-seijas: It'd be probably better to define the in `workload.py` and export it in the `src/synapse/__init__.py` to keep it more cohesive.

Thanhphan1147: Will do that, thanks.

src/synapse/api.py
Thanhphan1147: I did a cleanup of all the constants in https://github.com/canonical/synapse-operator/pull/36/commits/931dfd28cd354872cfc1466bc6d07234fbc06486, I also talked to Amanda to understand which constant does what and moved them to the correct place

src/synapse/api.py
Thanhphan1147: I imported all constants in `src/synapse/__init__.py` and changed all use of constants in the charm to add the `synapse.` prefix

src/mjolnir.py
```diff
@@ -13,20 +13,16 @@
...
-    SECRET_KEY,
-    SYNAPSE_CONTAINER_NAME,
-)
 from user import User
 
 logger = logging.getLogger(__name__)
 
 USERNAME = "mjolnir"
+# Disabling it since these are not hardcoded password
+SECRET_ID = "secret-id"  # nosec
```
arturo-seijas, line 22: Can you order the constants alphabetically?

Thanhphan1147: thanks, they are sorted across all files that defines constant (I finally found an editor extension to do that quickly)

arturo-seijas (approved):
LGTM

arturo-seijas (approved):

amandahla (approved):
LGTM!

PR #44, opened 2023-09-21
WIP - Add well know file to nginx workload container

amandahla:
### Overview

In order to provide delegation, this PR adds a configuration where the user can set a delegation server name. This name will be used to create a server file provided by the NGINX workload container.
More details in [Delegation of incoming traffic](https://matrix-org.github.io/synapse/latest/delegate.html).

### Rationale

This way you can have a server_name like example.com but a URL to your Synapse as synapse.example.com.

### Checklist

### DIFF
--- a/src/synapse/workload.py
+++ b/src/synapse/workload.py
@@ -5,6 +5,7 @@
 
 """Helper module used to manage interactions with Synapse."""
 
+import json
 import logging
 import typing
 
@@ -28,6 +29,7 @@ from constants import (
     SYNAPSE_NGINX_PORT,
     SYNAPSE_PORT,
     SYNAPSE_URL,
+    WELL_KNOW_FILE_PATH,
 )
 
 from .api import VERSION_URL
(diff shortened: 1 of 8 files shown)


src/synapse/workload.py
```diff
@@ -137,6 +139,23 @@ def check_mjolnir_ready() -> ops.pebble.CheckDict:
     return check.to_dict()
 
 
+def create_well_know_file(container: ops.Container, charm_state: CharmState) -> None:
```
jdkandersson, line 142: This function seems to generate a specific well known file (`.well-known/matrix/server`) compared to the function name `create_well_know_file` which seems to read like it can create a range of well known files.

Do you think this should be renamed to something with less implementation detail? Like create configure server name or something like that? The fact that it is creating a well known file seems to be an implementation detail the caller probably doesn't need to know

tests/unit/test_synapse_api.py
jdkandersson, line 1: I couldn't spot a new test to check this configuration? Perhaps also worth an integration test? It would be good to ensure that we check that the code we have written works and get test failures if someone accidentally breaks it or removes the feature. What do you think?

merkata:
The code is good, however I don't think this is how it works. The doc says to host it under `server_name`, so if you have a server name of say ubuntu.com, you will host the file there. The nginx here is an ingress for the actual server, say chat-server.ubuntu.com, right? It should listen on ubuntu.com for this to be effective. See mozilla.org -> https://www.mozilla.org/.well-known/matrix/server with the content of `{ "m.server": "mozilla.modular.im:443" }`. I think [this](https://ubuntu.com/.well-known/matrix/server) is what we're after and should test.

amandahla:
Related:
https://github.com/canonical/synapse-operator/pull/46

PR #46, opened 2023-09-22
Enable serve_server_wellknown

amandahla:
### Overview

Delegation is a Matrix feature that:
- Allows your Synapse server to be configured as example.com so, for example, users will be like user1:example.com
- But the URL can be different, for example, https://synapse.example.com

You need delegation for two scenarios:
1. The one mentioned above so you can route traffic to the expected URL
2. You want federation traffic to use a different port than 8448 (the default)

This PR addresses the situation 2 by enabling the configuration `serve_server_wellknown` making Synapse serve a ".well-known/matrix/server" file with the content:
```
{
    "m.server": "<server_name>:443"
}
```
Where <server_name> is the one configured in the Synapse instance.

### Rationale

Make federation traffic to use a different port than 8448. In this case, the port will be 443.

### Checklist

### DIFF
--- a/src/pebble.py
+++ b/src/pebble.py
@@ -96,6 +96,7 @@ class PebbleService:
         try:
             synapse.execute_migrate_config(container=container, charm_state=self._charm_state)
             synapse.enable_metrics(container=container)
+            synapse.enable_serve_server_wellknown(container=container)
             if self._charm_state.saml_config is not None:
                 logger.debug("pebble.change_config: Enabling SAML")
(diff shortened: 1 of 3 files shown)


nrobinaubertin (approved):
:+1:

yanksyoon (approved):

jdkandersson:
Thanks for submitting this - it seems there aren't any tests? Tests are important to ensure that the feature works as expected and protect from regression. Would it be possible to add tests in this case?

PR #48, opened 2023-09-26
Check if Synapse services are running before configuring Mjolnir

amandahla:
### Overview

While refreshing the charm, it was noticed that it tries to check for the membership room but the service is not up yet so it fails.

### Rationale

This PR will prevent this by checking if all services are running before configuring Mjolnir.

### Checklist


src/mjolnir.py
jdkandersson, line 1: Do you think it is worth writing an integration test that simulates the issue being addressed? Just so we don't have regression in the future and are sure that we have caught and resolved the issue

src/mjolnir.py
```diff
@@ -118,9 +118,16 @@ def _on_collect_status(self, event: ops.CollectStatusEvent) -> None:
...
-            logger.debug("%s service already exists, skipping")
+        mjolnir_service = container.get_services(MJOLNIR_SERVICE_NAME)
+        if mjolnir_service:
+            logger.debug("%s service already exists, skipping", MJOLNIR_SERVICE_NAME)
+            return
+        current_services = container.get_services()
+        all_svcs_running = all(svc.is_running() for svc in current_services.values())
+        if not all_svcs_running:
+            # Synapse must be running so the charm can check if there is
+            # a membership_room.
```
jdkandersson, line 129: This comment currently seems to say that it is possible to check if there is a `membership_room` when the charm is not running? The code seems to return so I'm guessing that it isn't possible to check under these conditions

jdkandersson (approved):

cbartz (approved):
LGTM

src/mjolnir.py
amandahla: Curious, my idea was to clarify that the charm can check for the membership_room only if Synapse is running, otherwise it must wait. I'll rephrase that.

src/mjolnir.py
amandahla: It's a good suggestion but I wonder how to do that since it's something that happens while refreshing the charm. Do you know if there is an integration test for that in our charms? Or have a suggestion for how to fire the upgrade-charm event?

jdkandersson: I believe the Wordpress charm has an upgrade test, so an integration test could deploy a previous version of the charm and then deploy this new version of the charm which should trigger the `upgrade-charm` event, would that work?

jdkandersson (approved):
Looks like you ended up implementing an upgrade test, nice!

src/mjolnir.py
```diff
@@ -118,9 +118,17 @@ def _on_collect_status(self, event: ops.CollectStatusEvent) -> None:
...
+        mjolnir_service = container.get_services(MJOLNIR_SERVICE_NAME)
+        if mjolnir_service:
+            logger.debug("%s service already exists, skipping", MJOLNIR_SERVICE_NAME)
+            return
+        current_services = container.get_services()
+        all_svcs_running = all(svc.is_running() for svc in current_services.values())
+        if not all_svcs_running or not current_services:
+            # The get_membership_room_id does a call to Synapse API in order to get the
+            # membership room id. This only works if Synapse and NGINX are running so that's why
+            # the services are being checked here.
```
cbartz, line 130: I think this should work now, but I wonder what would happen in the future if another service is added, could it be the case that `container.get_services` returns some services but not the synapse service and this construction stops working? Perhaps it would be safer to check for synapse more explicitly.

src/mjolnir.py
```diff
@@ -118,9 +118,17 @@ def _on_collect_status(self, event: ops.CollectStatusEvent) -> None:
...
-            logger.debug("%s service already exists, skipping")
+        mjolnir_service = container.get_services(MJOLNIR_SERVICE_NAME)
+        if mjolnir_service:
+            logger.debug("%s service already exists, skipping", MJOLNIR_SERVICE_NAME)
+            return
+        current_services = container.get_services()
+        all_svcs_running = all(svc.is_running() for svc in current_services.values())
+        if not all_svcs_running or not current_services:
+            # The get_membership_room_id does a call to Synapse API in order to get the
+            # membership room id. This only works if Synapse and NGINX are running so that's why
```
cbartz, line 129: I think NGINX has to be checked using the NGINX container ?

cbartz (commented):
Looks good to me overall, just have some minor comments/questions regarding the service available check.

arturo-seijas (dismissed):
LGTM

src/mjolnir.py
amandahla: Wow...excellent catch! I totally miss that.

Well, let's see:
One: my mistake, we don't need the NGINX container now since the API calls are done directly to Synapse.
Two: it's an edge case but is possible that the container has one or more services running but not the one that we need, the Synapse service.

I think this solves even the readability of this check because it could be changed only to check specifically the Synapse service.
What do you think?

src/mjolnir.py
amandahla: Perfect!

amandahla: (I'll mark this as resolved now because I'll address in the next comment)

src/mjolnir.py
amandahla: I believe so. Right now is doing something like:
- Deploy from charmhub, config server_name and enable Mjolnir (blocked status is expected)
- Refresh using the local charm
- Test
- Remove Synapse and install it again from the local charm
- Proceed with the existing tests

arturo-seijas (dismissed):
LGTM

gregory-schiano (approved):
LGTM

arturo-seijas (approved):
LGTM

PR #59, opened 2023-10-04
Add action to promote a user to an admin

amandahla:
### Overview

To promote a user to an admin you need to interact with the Synapse API. This action makes this easier so the user can use a juju command to do that like:
```
juju run-action synapse/0 promote-user-admin username=user1 --wait
```

Note:
- The [API call ](https://matrix-org.github.io/synapse/latest/admin_api/user_admin_api.html#change-whether-a-user-is-a-server-administrator-or-not)to do that requires an admin access token.
- Mjolnir has the same requirement so the interaction with Synapse and peer data to get this admin access token is already implemented there.
- This PR moves this implementation to a new module called Secret Storage so both, Mjolnir and the new action, can get an admin access token that is stored on peer data.
- Mind that is a little bit different from the way it was implemented on [Flask](https://github.com/canonical/flask-k8s-operator/blob/main/src/secret_storage.py) because to get an admin access token, we need Synapse to be running, register a new user and then we'll have the token we need. Since this is not needed for the charm itself, it will only happen if Mjolnir is enabled or if the user runs the action to make a user an admin. That's why we are not observing the leader elected or relation joined event.
- Why not add the admin access token to charm_state? This would create a cyclic import like this: Cyclic import (actions -> actions.register_user -> synapse -> synapse.api -> charm_state -> secret_storage)

Plus:
- Create a Secret Storage module
- Change all references to admin_access_token to typing.Optional[str]
- Fix typo in src/charm_state.py

### Rationale

Provide an easy way to make a user an admin.

### Checklist


src/secret_storage.py
```diff
@@ -0,0 +1,90 @@
...
+def _update_peer_data(charm: ops.CharmBase, container: ops.model.Container) -> None:
+    """Update peer data if needed.
+
+    Args:
+        charm: The charm object.
+        container: Synapse container.
+    """
+    # If there is no secret, we use peer relation data
+    # If there is secret, then we update the secret and add the secret id to peer data
+    peer_relation = charm.model.get_relation(PEER_RELATION_NAME)
```
gregory-schiano, line 33: You should probably pass the peer relation instead of getting it again as `get_admin_access_token` already gets the relation.
But it's fine testing again the relation as this method could be used elsewhere and forget to check it

src/secret_storage.py
```diff
@@ -0,0 +1,90 @@
...
+        return
+
+    if not JujuVersion.from_environ().has_secrets and not peer_relation.data[charm.app].get(
+        SECRET_KEY
+    ):
+        # we can't create secrets and peer data is empty
+        logger.debug("Updating peer relation data")
+        admin_user = actions.register_user(container, username, True)
+        peer_relation.data[charm.app].update({SECRET_KEY: admin_user.access_token})
+
```
gregory-schiano, line 62: Throw an exception here, it's more defensive

src/secret_storage.py
```diff
@@ -0,0 +1,90 @@
...
+    # there is no access token on peer data/secret, another user will be created.
+    #
+    # Using 16 to create a random value but to  be secure against brute-force attacks,
+    # please check the docs:
+    # https://docs.python.org/3/library/secrets.html#how-many-bytes-should-tokens-use
+    username = token_hex(16)
+    if JujuVersion.from_environ().has_secrets and not peer_relation.data[charm.app].get(SECRET_ID):
+        # we can create secrets and the one that we need was not created yet
+        logger.debug("Adding secret")
+        admin_user = actions.register_user(container, username, True)
```
gregory-schiano, line 50: Since register_user is called for both condition, you should probably make that call once before the if statements

src/secret_storage.py
```diff
@@ -0,0 +1,90 @@
...
+PEER_RELATION_NAME = "synapse-peers"
+# Disabling it since these are not hardcoded password
+SECRET_ID = "secret-id"  # nosec
+SECRET_KEY = "secret-key"  # nosec
+
+logger = logging.getLogger(__name__)
+
+
+def _update_peer_data(charm: ops.CharmBase, container: ops.model.Container) -> None:
+    """Update peer data if needed.
```
gregory-schiano, line 25: I feel like at least part of the comments in the method should be in the docstring as the docstring doesn't tell much about what the method does

src/secret_storage.py
```diff
@@ -0,0 +1,90 @@
...
+    """
+    container = charm.unit.get_container(synapse.SYNAPSE_CONTAINER_NAME)
+    if not container.can_connect():
+        logger.warning("Failed to get secret storage: waiting for pebble")
+        return None
+    peer_relation = charm.model.get_relation(PEER_RELATION_NAME)
+    if not peer_relation:
+        # there is no peer relation so nothing to be done
+        logger.warning("Failed to get secret storage: waiting for peer relation")
+        return None
```
gregory-schiano, line 81: I think it's better to throw an exception instead of returning None, and then the observer / charm manage the exception (probably adapt the charm status)

src/charm.py
```diff
@@ -174,6 +175,34 @@ def _on_register_user_action(self, event: ActionEvent) -> None:
...
+        """
+        results = {
+            "change-user-admin": False,
+        }
+        container = self.unit.get_container(synapse.SYNAPSE_CONTAINER_NAME)
+        if not container.can_connect():
+            event.fail("Failed to connect to container")
+            return
+        try:
+            admin_access_token = secret_storage.get_admin_access_token(self)
```
gregory-schiano, line 192: In the current implementation that could return none (and suggested implementation is to throw an exception). This should be handled here

src/mjolnir.py
```diff
@@ -54,57 +48,14 @@ def _pebble_service(self) -> typing.Any:
...
-            admin_user = self.create_admin_user(container)
-            peer_relation.data[self._charm.app].update({SECRET_KEY: admin_user.access_token})
-
-    def create_admin_user(self, container: ops.model.Container) -> User:
-        """Create an admin user.
-
-        Args:
-            container: Synapse container.
+    @property
+    def _admin_access_token(self) -> typing.Optional[str]:
```
gregory-schiano, line 52: Note that if you change the `get_admin_access_token` method to throw an exception as suggested below, it's better not to use a property (as properties shouldn't thrown exceptions) or to add a try/catch block here

src/secret_storage.py
gregory-schiano, line 1: Since this isn't a listener, and is highly correlated to the actions, I think these 2 methods should be part of the suggested merged `action` module

src/secret_storage.py
```diff
@@ -0,0 +1,90 @@
...
+    _update_peer_data(charm, container)
+    if JujuVersion.from_environ().has_secrets:
+        secret_id = peer_relation.data[charm.app].get(SECRET_ID)
+        if secret_id:
+            secret = charm.model.get_secret(id=secret_id)
+            secret_value = secret.get_content().get(SECRET_KEY)
```
gregory-schiano, line 87: Should probably thrown an exception is the secret_id is not found, and also do some cleanup in the peer databag

src/secret_storage.py
```diff
@@ -0,0 +1,90 @@
...
+    if not container.can_connect():
+        logger.warning("Failed to get secret storage: waiting for pebble")
+        return None
+    peer_relation = charm.model.get_relation(PEER_RELATION_NAME)
+    if not peer_relation:
+        # there is no peer relation so nothing to be done
+        logger.warning("Failed to get secret storage: waiting for peer relation")
+        return None
+    _update_peer_data(charm, container)
+    if JujuVersion.from_environ().has_secrets:
```
gregory-schiano, line 83: You should probably store the result of this in a constant as this is a pretty static value, and this is used multiple times

src/synapse/api.py
```diff
@@ -461,3 +467,25 @@ def _do_request(
...
+) -> None:
+    """Change user to admin.
+
+    Args:
+        user: user to be changed to admin.
+        server: to be used to create the user id.
+        admin_access_token: server admin access token to be used.
+    """
+    authorization_token = f"Bearer {admin_access_token}"
+    headers = {"Authorization": authorization_token}
```
gregory-schiano, line 485: This should be in a private method of the module, this is used multiple times (and will probably be used if the module needs more api calls)
This could also be added to `_do_requests` by adding a `requires_authentication` boolean argument

actions.yaml
```diff
@@ -22,3 +22,12 @@ register-user:
       default: false
   required:
     - username
+change-user-admin:
+  description: |
+    Change whether a user is a server administrator.
```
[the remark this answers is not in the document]
amandahla: I agree, changed.

actions.yaml
```diff
@@ -22,3 +22,12 @@ register-user:
...
+    You need to supply a user name.
+  properties:
+    username:
+      description: |
+        User name to be changed to admin.
```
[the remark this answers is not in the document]
amandahla: I agree, changed.

src/actions/change_user_admin.py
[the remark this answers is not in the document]
amandahla: It makes sense but since there is a file only for reset_instance action, I thought that it would be clearer to keep having one file per action.

src/actions/change_user_admin.py
```diff
@@ -0,0 +1,52 @@
...
+import logging
+import typing
+
+import synapse
+from user import User
+
+logger = logging.getLogger(__name__)
+
+
+class ChangeUserAdminError(Exception):
```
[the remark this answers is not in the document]
amandahla: Maybe but for this PR I'm not changing exceptions that affect other actions/modules because there is a story only to refactor this: ISD-1070

src/actions/change_user_admin.py
```diff
@@ -0,0 +1,52 @@
...
+
+        Args:
+            msg (str): Explanation of the error.
+        """
+        self.msg = msg
+
+
+# admin_access_token is not a password
+def change_user_admin(
+    username: str, server: typing.Optional[str], admin_access_token: typing.Optional[str]
```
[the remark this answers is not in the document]
amandahla: My bad, should be required. Changed.

src/secret_storage.py
amandahla: But this is to work without secrets, using peer data instead. Shouldn't it be compatible with Juju 2?

src/secret_storage.py
amandahla: I thought about that but if the secret or peer data already exists, this would create a useless user.

src/charm.py
```diff
@@ -174,6 +175,34 @@ def _on_register_user_action(self, event: ActionEvent) -> None:
...
+            actions.change_user_admin(
+                username=event.params["username"],
+                server=self._charm_state.synapse_config.server_name,
+                admin_access_token=admin_access_token,
+            )
+            results["change-user-admin"] = True
+        except (PebbleServiceError, actions.ChangeUserAdminError) as exc:
```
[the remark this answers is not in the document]
amandahla: Nothing, removed.

src/charm.py
```diff
@@ -174,6 +175,34 @@ def _on_register_user_action(self, event: ActionEvent) -> None:
...
+            self.model.unit.status = ops.BlockedStatus(str(exc))
```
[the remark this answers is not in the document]
amandahla: Good catch, there is no action for the user to take here in case of a problem. Removing it.

src/secret_storage.py
amandahla: I'll leave it then just in case.

src/secret_storage.py
amandahla: Good point, docstring expanded.

src/secret_storage.py
amandahla: Humm, I see your point and I partially agree because is not an observer but is not an action neither to be put on there. To me looks more like the pebble.py file which is something related to Juju/Internal and provides a way of interacting with that.

src/secret_storage.py
amandahla: Nice, changed.

src/mjolnir.py
amandahla: Nice, changed.

src/secret_storage.py
amandahla: Exception is being raised now.
What do you mean by cleanup? If secret id is not found, what should be done?

src/synapse/api.py
amandahla: Nice suggestion, changed.

jdkandersson (commented):
Have not looked at the tests yet

src-docs/api.md
```diff
@@ -124,14 +129,14 @@ This is a way to do actions on behalf of a user.
...
-<a href="../src/synapse/api.py#L274"><img align="right" style="float:right;" src="https://img.shields.io/badge/-source-cccccc?style=flat-square"></a>
+<a href="../src/synapse/api.py#L337"><img align="right" style="float:right;" src="https://img.shields.io/badge/-source-cccccc?style=flat-square"></a>
 
 ## <kbd>function</kbd> `override_rate_limit`
 
 ```python
 override_rate_limit(
     user: User,
-    admin_access_token: str,
+    admin_access_token: Optional[str],
```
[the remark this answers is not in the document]
amandahla: This was definitely my mistake, I could swear that I changed it already.

src/secret_storage.py
amandahla: Since this has no impact on the feature that this PR intends to provide, I suggest refactoring this, if needed,  in a follow-up PR after discussing the best approach/architecture guidance regarding peer data/secrets.

src/actions/__init__.py
```diff
@@ -3,7 +3,7 @@
 
 """Actions package is used to run actions provided by the charm."""
 
-from .register_user import RegisterUserError, register_user  # noqa: F401
-
 # Exporting methods to be used for another modules
+from .promote_user_admin import PromoteUserAdminError, promote_user_admin  # noqa: F401
```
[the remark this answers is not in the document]
amandahla: The comment above "# Exporting methods to be used for another modules" is related to that.

src/actions/promote_user_admin.py
```diff
@@ -0,0 +1,52 @@
...
+
+    Raises:
+        PromoteUserAdminError: if something goes wrong while promoting the user to
+            be an admin.
+    """
+    try:
+        user = User(username=username, admin=True)
+        synapse.promote_user_admin(user=user, server=server, admin_access_token=admin_access_token)
+    except synapse.APIError as exc:
+        raise PromoteUserAdminError(str(exc)) from exc
```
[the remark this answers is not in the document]
amandahla: Well, I believe this could be applied to every action then?
That way, the charm would handle more than just the action event.

src/actions/__init__.py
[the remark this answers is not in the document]
jdkandersson: oh yes, you are right

src/actions/promote_user_admin.py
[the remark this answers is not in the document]
jdkandersson: As long as it is simple code like this it should be fine, perhaps some of this code could also be out into the `synapse` module

src/secret_storage.py
jdkandersson: For the actions - yes a follow up PR is fine

Are you also referring to the discussion around the purpose of the `secret_storage` module? It would be good to finish that discussion since I think this PR proposes introducing the module?

amandahla: No problem.

Regarding the user creation, this is done on the secret storage module because this is what being stored there and this way prevents a user from being created somewhere else. This would be equivalent to [this](https://github.com/canonical/flask-k8s-operator/blob/main/src/secret_storage.py#L46) on Flask charm but instead of generating a random key, Synapse needs an admin access token obtained after creating an admin user.

src/actions/promote_user_admin.py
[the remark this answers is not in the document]
amandahla: Nice suggestion, changed to the charm.py

src-docs/promote_user_admin.md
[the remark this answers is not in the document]
amandahla: Curious, I removed the file but this md is still here.

src/secret_storage.py
jdkandersson: Please have a think about how that can be changed so that the secret storage module doesn't interact with Synapse to create a user. One idea is to generate the user, if required, in `charm.py` before interacting with the secret storage and pass the created as as an argument to the module. That way, the secret storage has its pre-conditions met and can focus on its core purpose

src/actions/change_user_admin.py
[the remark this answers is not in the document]
amandahla: Marking as resolved since the file was removed.

src/secret_storage.py
amandahla: I see your point and it makes sense but I think we might fall into an egg-chicken situation :)

The charm needs to create a user only if there is no secret/peer data already. So if we change this to be in the charm, then the interaction with peer data/secret will be done there which is not nice as well.

I thought that would be ok for a module to interact with Synapse since the Pebble module does that too.

gregory-schiano: TBH the fact that this module interacts with Synapse doesn't bother me much. The module share the same name as the one in Flask, is inspired by it, but has a radical difference in that the one on Flask is effectively a listener as it hook on some events.

What annoys me more is the fact that this module is not a listener but interacts with the charm, which is in contradiction with our guidelines

Either we ensure this module doesn't interact with the charm and at most interact with Pebble, or we put this code in a listener (like action listener), or we put it in the charm.
But I'm not confident having this module in between.

src/secret_storage.py
gregory-schiano: I meant doing something like this:
```python 
    if not JujuVersion.from_environ().has_secrets and not peer_relation.data[charm.app].get(
        SECRET_KEY
    ):
        # we can't create secrets and peer data is empty
        logger.debug("Updating peer relation data")
        admin_user = actions.register_user(container, username, True)
        peer_relation.data[charm.app].update({SECRET_KEY: admin_user.access_token})
        return

raise Exception("Shouldn't happen")
```

src/secret_storage.py
gregory-schiano: If we have a secret_id in the peer data relation that doesn't exists as a secret, so we should remove that secret_id from the peer relation data

src/actions/change_user_admin.py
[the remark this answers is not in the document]
amandahla: I'll mark this one as resolved but feel free to re-open it.

src-docs/promote_user_admin.md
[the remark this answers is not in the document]
amandahla: I noticed that while generating src-docs, if you remove a file, lazydocs does not remove it from the previously generated docs. So I changed the shell script to remove the directory before calling lazydocs.

nrobinaubertin (approved):

src/secret_storage.py
amandahla: I see, thanks for all the reviews.
I created a get_admin_access_token method in the charm that will be used by the action and Mjolnir. So there is no more secret_storage module and only the charm interacts with Secrets/Peer data.
Still working on fixing the unit tests.

src/secret_storage.py
amandahla: Oh now I see...Well, in that case, shouldn't we let the charm raise an error? because if there is an ID, something happened outside of the charm to delete the expected secret.

src/mjolnir.py
```diff
@@ -162,29 +103,18 @@ def get_membership_room_id(self) -> typing.Optional[str]:
...
-            room_name=synapse.MJOLNIR_MEMBERSHIP_ROOM, admin_access_token=admin_access_token
-        )
-
-    def get_admin_access_token(self) -> str:
-        """Get admin access token.
 
-        Returns:
-            admin access token.
+        Raises:
+            AdminAccessTokenNotFoundError: if there is not admin access token.
```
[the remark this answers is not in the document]
amandahla: Removed ;)

src/secret_storage.py
gregory-schiano: I just checked and if the secret is not found a SecretNotFound exception is raised. My worry is that since the secret_id will stay in the peer relation data and I don't know what the user can do to cleanup these data (unless undeploy / redeploy).
I think that's better to catch the error, cleanup the peer relation data and reraise the error

amandahla: Nice, changed.

gregory-schiano (dismissed):
LGTM

src/secret_storage.py
amandahla: I forgot to come back to this conversation so I returned None instead of raising the Exception. I'll add a note to the Exception story to review this.

src/charm.py
```diff
@@ -174,6 +186,106 @@ def _on_register_user_action(self, event: ActionEvent) -> None:
...
 
+    def _create_admin_user(self) -> typing.Optional[User]:
+        """Create admin user.
+
+        Returns:
+            Admin user with token to be used in Synapse API requests or None if fails.
+        """
+        container = self.unit.get_container(synapse.SYNAPSE_CONTAINER_NAME)
+        if not container.can_connect():
+            logger.error("Failed to create admin user: waiting for pebble")
```
nrobinaubertin, line 197: Should it be "Failed to create admin user: failed to connect to the container" ?

nrobinaubertin (approved):

src/charm.py
amandahla: Good catch. To not lose your approval, I'll fix this in my other open PR [here](https://github.com/canonical/synapse-operator/pull/77).

amandahla: Since I had to add more unit test,  I fixed this here, thanks.

merkata (approved):

nrobinaubertin (approved):

Issue #62, opened 2023-10-10
Disable Password Login


kewisch:
### Enhancement Proposal

We use SSO exclusively. The password login should be disabled in [password_config](https://matrix-org.github.io/synapse/latest/usage/configuration/config_documentation.html#password_config). This will also avoid the profile settings where it looks like users can change their passwords.

kewisch:
@amandahla while #77 is merged, it appears we haven't actually disabled this for the ubuntu server. can you take a look? QA STR:

1. Open private window, go to app.element.io
2. Edit homeserver to be ubuntu.com

Expected:
<img width="696" alt="image" src="https://github.com/canonical/synapse-operator/assets/607198/5c84edc7-fad8-4d86-ae57-6e4db211ad3d">

Actual right now:
<img width="679" alt="image" src="https://github.com/canonical/synapse-operator/assets/607198/ea808b38-54d1-4894-903c-ce06ff4717e5">

amandahla:
This issue has been transferred to a Jira story for further tracking as a feature request. Closing this issue for now. Thanks for your input!

PR #77, opened 2023-10-18
Add enable_password_config

amandahla:
### Overview

Add configuration to change password_config behavior.

### Rationale

This way the administrator can disable password login if the user login via SSO, for example.

### Checklist


amandahla:
Fix #62

merkata (commented):
Looking at the [config](https://matrix-org.github.io/synapse/v1.59/usage/configuration/config_documentation.html) documentation, password_config is nested, so I'm suggesting to expose password_config and deal with enabling/disabling in the function. I imagine if we want to deal with password configuration in the future, if we go with the current route we would add additional fields, like minimum_length_policy_password_config (to get to password_config.policy.minimum_length and change the min pass length for instance). Another thing to consider is when we render the current password_config to update what we're interested in and the rest should remain unchanged, so if you toggle password_config on and off and you have made changes to the policy (like the min pass length) it does not get lost. It would be clearer to me also that a disable_password_config that takes a password_config object changes the enabled parameter and not enable_password_config. I'm not sure how the function is called though - shouldn't pebble be made aware of that change via the on.config_changed path?

config.yaml
```diff
@@ -8,6 +8,11 @@ options:
     description: |
       Configures whether to enable Mjolnir - moderation tool for Matrix.
       Reference: https://github.com/matrix-org/mjolnir
+  enable_password_config:
+    type: boolean
+    default: true
+    description: |
+      Defaults to true. Set to false to disable password authentication.
```
mthaddon, line 15: I think it might be worth explaining why you might want to disable this - e.g. if you have SSO.

amandahla: Good point, changed.

src/synapse/workload.py
```diff
@@ -289,6 +289,24 @@ def enable_metrics(container: ops.Container) -> None:
         raise EnableMetricsError(str(exc)) from exc
 
 
+def disable_password_config(container: ops.Container) -> None:
```
cbartz, line 292: Sorry, but I have not seen where this function is called? Shouldn't it be called when the config is set to false?

gregory-schiano:
Isn't something missing in this PR ?
I see a new config added to the charm_start but it doesn't seem to be used, and there's a method that disable password config but in an "hardcoded" way

src/synapse/workload.py
amandahla: I have no idea what happened, maybe it was lost while merging with previous PRs, but now I added to change_config the call to this function.

nrobinaubertin (approved):

tests/unit/test_synapse_workload.py
```diff
@@ -377,6 +377,51 @@ def test_enable_serve_server_wellknown_error(monkeypatch: pytest.MonkeyPatch):
...
+        "password_config": {
+            "enabled": False,
+        },
+    }
+    assert push_mock.call_args[0][1] == yaml.safe_dump(expected_config_content)
+
+
+def test_disable_password_config_error(monkeypatch: pytest.MonkeyPatch):
+    """
+    arrange: set mock container with file.
```
cbartz, line 411: Nitpicky: I think it is worth mentioning that an error case is set up here.

cbartz (approved):
Looks good, have one nitpicky comment.

tests/unit/test_synapse_workload.py
amandahla: Thanks for pointing that out. I'll change it in a follow-up PR (I'll add to the refactoring issue just in case to not forget).

PR #108, opened 2023-11-21
Add Synapse NGINX pebble ready event handler

amandahla:
### Overview

If for any reason the Synapse NGINX pebble check fails, the container is restarted, a pebble ready event emitted but since there is no handler for it, the pebble plan is not re-created. This PR solves this by adding a handler to this event.

Not related but since is a small change this PR also add a strip to each item of federation domain whitelist as suggested [here](https://github.com/canonical/synapse-operator/pull/94).

### Rationale

Guarantee that Synapse NGINX pebble layer is always planned.

### Juju Events Changes

Add _on_synapse_nginx_pebble_ready

### Module Changes

N/A

### Library Changes

N/A

### Checklist

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -96,7 +96,7 @@ class SynapseCharm(ops.CharmBase):
         """Change configuration."""
         container = self.unit.get_container(synapse.SYNAPSE_CONTAINER_NAME)
         if not container.can_connect():
-            self.unit.status = ops.MaintenanceStatus("Waiting for pebble")
+            self.unit.status = ops.MaintenanceStatus("Waiting for Synapse pebble")
             return
         self.model.unit.status = ops.MaintenanceStatus("Configuring Synapse")
         try:
@@ -106,7 +106,7 @@ class SynapseCharm(ops.CharmBase):
             return
         container = self.unit.get_container(synapse.SYNAPSE_NGINX_CONTAINER_NAME)
         if not container.can_connect():
-            self.unit.status = ops.MaintenanceStatus("Waiting for pebble")
+            self.unit.status = ops.MaintenanceStatus("Waiting for Synapse NGINX pebble")
             return
         self.pebble_service.replan_nginx(container)
         self.model.unit.status = ops.ActiveStatus()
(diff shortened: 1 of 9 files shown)


cbartz (approved):
LGTM

src/charm.py
```diff
@@ -111,7 +104,12 @@ def change_config(self) -> None:
         except PebbleServiceError as exc:
             self.model.unit.status = ops.BlockedStatus(str(exc))
             return
-        self.replan_nginx()
+        container = self.unit.get_container(synapse.SYNAPSE_NGINX_CONTAINER_NAME)
+        if not container.can_connect():
+            self.unit.status = ops.MaintenanceStatus("Waiting for pebble")
```
mthaddon, line 109: Should we maybe say which pebble we're waiting for here? Same comment for others below.

mthaddon (commented):
One comment inline. Also, I think it would be good to have a test/some tests for this change.

amandahla:
> One comment inline. Also, I think it would be good to have a test/some tests for this change.

@mthaddon Any suggestion on how to test this situation? Integration test?
I guess the test would need to have a way to force NGINX check to fail and then try to request Synapse URL. But I don't know how to force this fail. What do you think?

mthaddon:
> > One comment inline. Also, I think it would be good to have a test/some tests for this change.
> 
> @mthaddon Any suggestion on how to test this situation? Integration test? I guess the test would need to have a way to force NGINX check to fail and then try to request Synapse URL. But I don't know how to force this fail. What do you think?

I think we could have a test that if the relevant pebble ready event fires the `replan_nginx` method is called - does that sound right to you?

amandahla:
> > > One comment inline. Also, I think it would be good to have a test/some tests for this change.
> > 
> > 
> > @mthaddon Any suggestion on how to test this situation? Integration test? I guess the test would need to have a way to force NGINX check to fail and then try to request Synapse URL. But I don't know how to force this fail. What do you think?
> 
> I think we could have a test that if the relevant pebble ready event fires the `replan_nginx` method is called - does that sound right to you?

That sounds much easier than what I was thinking :smile: I did a unit test for that.

src/charm.py
amandahla: Sure, changed.

amandahla: I'll close this one but feel free to re-open it.

cbartz (approved):

merkata (approved):
LGTM

PR #112, opened 2023-11-28
Add allow_public_rooms_over_federation configuration

amandahla:
### Overview

While working with federation, one might want toallow public rooms over federation. This configuration allows that.

Reference: https://matrix-org.github.io/synapse/latest/usage/configuration/config_documentation.html#allow_public_rooms_over_federation

### Rationale

Change allow_public_rooms_over_federation configuration.

### Checklist

- [X] The [charm style guide](https://juju.is/docs/sdk/styleguide) was applied
- [X] The [contributing guide](https://github.com/canonical/is-charms-contributing-guide) was applied
- [X] The changes are compliant with [ISD054 - Manging Charm Complexity](https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619)
- [X] The documentation is generated using `src-docs`
- [X] The documentation for charmhub is updated.
- [X] The PR is tagged with appropriate label (`urgent`, `trivial`, `complex`)

### DIFF
--- a/src/pebble.py
+++ b/src/pebble.py
@@ -97,6 +97,8 @@ class PebbleService:
                 synapse.enable_federation_domain_whitelist(
                     container=container, charm_state=self._charm_state
                 )
+            if self._charm_state.synapse_config.allow_public_rooms_over_federation:
+                synapse.enable_allow_public_rooms_over_federation(container=container)
             self.restart_synapse(container)
         except (synapse.WorkloadError, ops.pebble.PathError) as exc:
             raise PebbleServiceError(str(exc)) from exc
--- a/src/synapse/workload.py
+++ b/src/synapse/workload.py
@@ -348,6 +348,24 @@ def enable_federation_domain_whitelist(container: ops.Container, charm_state: Ch
         raise WorkloadError(str(exc)) from exc
 
 
+def enable_allow_public_rooms_over_federation(container: ops.Container) -> None:
+    """Change the Synapse configuration to allow public rooms in federation.
+
+    Args:
+        container: Container of the charm.
+
+    Raises:
+        WorkloadError: something went wrong enabling configuration.
+    """
+    try:
+        config = container.pull(SYNAPSE_CONFIG_PATH).read()
+        current_yaml = yaml.safe_load(config)
+        current_yaml["allow_public_rooms_over_federation"] = True
+        container.push(SYNAPSE_CONFIG_PATH, yaml.safe_dump(current_yaml))
+    except ops.pebble.PathError as exc:
+        raise WorkloadError(str(exc)) from exc
+
+
 def _get_mjolnir_config(access_token: str, room_id: str) -> typing.Dict:
     """Create config as expected by mjolnir.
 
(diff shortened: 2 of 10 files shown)


nrobinaubertin (approved):

src/pebble.py
```diff
@@ -97,6 +97,8 @@ def change_config(self, container: ops.model.Container) -> None:
                 synapse.enable_federation_domain_whitelist(
                     container=container, charm_state=self._charm_state
                 )
+            if self._charm_state.synapse_config.allow_public_rooms_over_federation:
+                synapse.enable_allow_public_rooms_over_federation(container=container)
```
jdkandersson, line 101: What is the reason for putting this in the pebble service please? I realise there are other similar interactions above

I thought that the pebble service is more for generating the pebble service rather than specifically reacting to every event? It doesn't seem like the interactions here are related to pebble, they seem to be related to handling events from the charm?

src/synapse/workload.py
```diff
@@ -348,6 +348,24 @@ def enable_federation_domain_whitelist(container: ops.Container, charm_state: Ch
...
+    Raises:
+        WorkloadError: something went wrong enabling configuration.
+    """
+    try:
+        config = container.pull(SYNAPSE_CONFIG_PATH).read()
+        current_yaml = yaml.safe_load(config)
+        current_yaml["allow_public_rooms_over_federation"] = True
+        container.push(SYNAPSE_CONFIG_PATH, yaml.safe_dump(current_yaml))
+    except ops.pebble.PathError as exc:
+        raise WorkloadError(str(exc)) from exc
```
jdkandersson, line 366: What if it later gets disabled?

amandahla: If is not set, the default is False: https://ems-docs.element.io/books/element-cloud-documentation/page/import-database-and-media-dump
So the configuration is only added to the configuration file if is set to True.

src/pebble.py
amandahla: Any configuration change requires a Pebble Layer replan so that's why this is here. This way the database observer and the config-change event call the same method, avoiding repetitive code.

Discussed here:
https://github.com/canonical/synapse-operator/pull/13#discussion_r1255151747

jdkandersson: So wouldn't it be possible to handle `config-changed` in `charm.py` and the ask pebble to restart the service as a part of the code in `charm.py`? I might have missed it, I don't see any circular dependency issues here since all of the interactions with the synapse module just depend on the container which is available in `charm.py`

src/synapse/workload.py
jdkandersson: So what about if it was once set to `True` and then the user disables it after that? How would the line from the configuration end up getting removed?

amandahla: The configuration file is regenerated for every config-change event. If the user disables it, will emit a config changed event and the configuration file will be regenerated without this line.

src/pebble.py
amandahla: Could you create an issue/story for that so we can discuss a solution? Changing this in this PR would affect how the configuration change is handled, being something out of context here.

jdkandersson: Yes we can make it a separate story

amandahla: Thanks!

PR #113, opened 2023-11-29
Add ip_range_whitelist

amandahla:
### Overview

Add configuration ip_range_whitelist

https://matrix-org.github.io/synapse/latest/usage/configuration/config_documentation.html#ip_range_whitelist

If is not set, nothing is added to the configuration file and the default value is empty.

### Rationale

Add configuration ip_range_whitelist

### Checklist

- [X] The [charm style guide](https://juju.is/docs/sdk/styleguide) was applied
- [X] The [contributing guide](https://github.com/canonical/is-charms-contributing-guide) was applied
- [X] The changes are compliant with [ISD054 - Manging Charm Complexity](https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619)
- [X] The documentation is generated using `src-docs`
- [X] The documentation for charmhub is updated.
- [X] The PR is tagged with appropriate label (`urgent`, `trivial`, `complex`)


config.yaml
```diff
@@ -19,6 +19,13 @@ options:
     type: string
     description: Comma separated list of domains to be allowed to federate.
     default: ''
+  ip_range_whitelist:
```
varshigupta12, line 22: Do we need to "whitelist" here? Can we use allowlist instead just to keep the code inclusive?

amandahla: Unfortunately, this is how the configuration is set in Synapse:
https://matrix-org.github.io/synapse/latest/usage/configuration/config_documentation.html#ip_range_whitelist
I'm afraid that changing it here to a different name might be confusing.

src/charm_state.py
```diff
@@ -196,9 +183,8 @@ def from_charm(
         Raises:
             CharmConfigInvalidError: if the charm configuration is invalid.
         """
-        synapse_config = {k: v for k, v in charm.config.items() if k in KNOWN_CHARM_CONFIG}
         try:
-            valid_synapse_config = SynapseConfig(**synapse_config)  # type: ignore
+            valid_synapse_config = SynapseConfig(**dict(charm.config.items()))  # type: ignore
```
jdkandersson, line 187: Is the `type: ignore` still needed? If so, what is the reason for needing it please?

src/charm_state.py
```diff
@@ -94,6 +80,7 @@ class SynapseConfig(BaseModel):  # pylint: disable=too-few-public-methods
     enable_mjolnir: bool = False
     enable_password_config: bool = True
     federation_domain_whitelist: str | None = Field(None)
+    ip_range_whitelist: str | None = Field(None)
```
jdkandersson, line 83: Should we do some more validation here? I noticed that later we're splitting it by comma, stripping whitespace and I'm guessing relying on some formatting of IP addresses?

tests/unit/test_synapse_workload.py
```diff
@@ -22,6 +22,94 @@
...
+          port: 8080
+          bind_addresses:
+            - "::"
+    """
+    text_io_mock = io.StringIO(config_content)
+    pull_mock = Mock(return_value=text_io_mock)
+    push_mock = MagicMock()
+    container_mock = MagicMock()
+    monkeypatch.setattr(container_mock, "pull", pull_mock)
+    monkeypatch.setattr(container_mock, "push", push_mock)
```
jdkandersson, line 43: What is the reason for needing this? Doesn't harness provide a container with a filesystem now so we don't need to mock anymore?

jdkandersson: @weiiwang01 wrote that feature 🙂

tests/unit/test_synapse_workload.py
```diff
@@ -22,6 +22,94 @@
...
+
+    expected_first_domain = "foo1"
+    expected_second_domain = "foo2"
+    harness.update_config(
+        {"ip_range_whitelist": f"{expected_first_domain},{expected_second_domain}"}
```
jdkandersson, line 48: This is somewhat limited testing, should we add a [pytest parametrize](https://docs.pytest.org/en/7.1.x/how-to/parametrize.html) with a bunch of valid options?

tests/unit/test_synapse_workload.py
amandahla: But would this be applicable here since the test is mocking the container to check how the pull/push are being called?

tests/unit/test_synapse_workload.py
amandahla: I changed my approach actually. Instead of adding unit test for this, I created an integration test that check if the configuration is validated or not.

src/charm_state.py
amandahla: Instead of re-creating the validation that Synapse does, I found out that there is a script that validates the entire configuration file that will be called by the method validate_config. If there is something wrong, the charm will be blocked so the user can review the configuration.

jdkandersson: Does it say specifically what is wrong? There also seem to be charm specific things like that it is a comma-separated list, right?

amandahla: Yes, it does. Here is an example of the output that will be available in the log: `('Invalid IP range provided: banana.', ('ip_range_whitelist',))`

tests/unit/test_synapse_workload.py
amandahla: I'll close this comment since is outdated so new comments can be made in this new version but feel free to re-open it.

tests/unit/test_synapse_workload.py
amandahla: I'll close this comment since is outdated so new comments can be made in this new version but feel free to re-open it.

src/synapse/workload.py
```diff
@@ -366,6 +366,42 @@ def enable_allow_public_rooms_over_federation(container: ops.Container) -> None:
...
+
+    Args:
+        charm_state: Instance of CharmState.
+
+    Returns:
+        IP range whitelist as expected by Synapse or None.
+    """
+    # The field is validated in enable_ip_range_whitelist
+    whitelist = charm_state.synapse_config.ip_range_whitelist
+    return [item.strip() for item in whitelist.split(",")]  # type: ignore[union-attr]
```
jdkandersson, line 380: Since we're doing the logic of splitting and stripping whitespace, should there be a unit test covering:

1. empty list
2. single item list
3. multiple item list
4. item with leading whitespace
5. item with trailing whitespace

tests/unit/test_synapse_workload.py
```diff
@@ -72,6 +72,98 @@ def test_allow_public_rooms_over_federation_error(monkeypatch: pytest.MonkeyPatc
...
+def test_enable_ip_range_whitelist_error(harness: Harness, monkeypatch: pytest.MonkeyPatch):
+    """
+    arrange: set mock container with file.
+    act: update ip_range_whitelist config and call enable_smtp.
+    assert: raise WorkloadError in case of error.
+    """
+    error_message = "Error pulling file"
+    path_error = ops.pebble.PathError(kind="fake", message=error_message)
+    pull_mock = MagicMock(side_effect=path_error)
+    container_mock = MagicMock()
```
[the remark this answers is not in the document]
jdkandersson: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock

tests/unit/test_synapse_workload.py
```diff
@@ -72,6 +72,98 @@ def test_allow_public_rooms_over_federation_error(monkeypatch: pytest.MonkeyPatc
...
+          port: 8080
+          bind_addresses:
+            - "::"
+    """
+    text_io_mock = io.StringIO(config_content)
+    pull_mock = Mock(return_value=text_io_mock)
+    push_mock = MagicMock()
+    container_mock = MagicMock()
+    monkeypatch.setattr(container_mock, "pull", pull_mock)
+    monkeypatch.setattr(container_mock, "push", push_mock)
```
jdkandersson, line 158: Similar to above, could we just construct one MagicMock and rely on that attributes are automatically generated? That way we save a few lines here

src/charm_state.py
amandahla: Sure, this is the message:
```python
lint run-test: commands[5] | mypy /home/amanda/Canonical/synapse-operator/src/ /home/amanda/Canonical/synapse-operator/tests/
src/charm_state.py:191: error: Argument 1 to "SynapseConfig" has incompatible type "**dict[str, str]"; expected "bool"  [arg-type]
src/charm_state.py:191: error: Argument 1 to "SynapseConfig" has incompatible type "**dict[str, str]"; expected "int | None"  [arg-type]
Found 2 errors in 1 file (checked 33 source files)
```

src/synapse/workload.py
```diff
@@ -366,6 +366,42 @@ def enable_allow_public_rooms_over_federation(container: ops.Container) -> None:
         raise WorkloadError(str(exc)) from exc
 
 
+def _create_ip_range_whitelist(charm_state: CharmState) -> list[str]:
```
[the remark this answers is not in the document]
amandahla: Good point, changed.

tests/unit/test_synapse_workload.py
```diff
@@ -72,6 +72,98 @@ def test_allow_public_rooms_over_federation_error(monkeypatch: pytest.MonkeyPatc
...
+    container_mock = MagicMock()
+    monkeypatch.setattr(container_mock, "pull", pull_mock)
+
+    with pytest.raises(synapse.WorkloadError, match=error_message):
+        expected_first_domain = "foo1"
+        expected_second_domain = "foo2"
+        harness.update_config(
+            {"ip_range_whitelist": f"{expected_first_domain},{expected_second_domain}"}
+        )
+        harness.begin()
```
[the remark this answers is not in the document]
amandahla: Since the validation was changed, this test now checks if the charm status is blocked instead of catching an Exception.

tests/unit/test_synapse_workload.py
```diff
@@ -72,6 +72,98 @@ def test_allow_public_rooms_over_federation_error(monkeypatch: pytest.MonkeyPatc
...
+    """
+    text_io_mock = io.StringIO(config_content)
+    pull_mock = Mock(return_value=text_io_mock)
+    push_mock = MagicMock()
+    container_mock = MagicMock()
+    monkeypatch.setattr(container_mock, "pull", pull_mock)
+    monkeypatch.setattr(container_mock, "push", push_mock)
+
+    harness.begin()
+    synapse.enable_ip_range_whitelist(container_mock, harness.charm._charm_state)
```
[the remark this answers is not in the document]
amandahla: Done.

src/synapse/workload.py
```diff
@@ -265,6 +265,27 @@ def execute_migrate_config(container: ops.Container, charm_state: CharmState) ->
...
+    validate_config_result = _exec(
+        container, ["/usr/bin/python3", "-m", "synapse.config", "-c", SYNAPSE_CONFIG_PATH]
+    )
+    if validate_config_result.exit_code:
+        logger.error(
+            "validate config failed, stdout: %s, stderr: %s",
+            validate_config_result.stdout,
+            validate_config_result.stderr,
+        )
+        raise WorkloadError("Validate config failed, please review your charm configuration")
```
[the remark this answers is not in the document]
amandahla: But I changed it to make it more clear to check  the logs.

tests/unit/test_synapse_workload.py
```diff
@@ -108,33 +133,26 @@ def test_enable_ip_range_whitelist_success(ip_range_whitelist: str, harness: Har
             "listeners": [
                 {"type": "http", "port": 8080, "bind_addresses": ["::"]},
             ],
-            "ip_range_whitelist": synapse.workload._create_ip_range_whitelist(
-                harness.charm._charm_state
-            ),
+            "ip_range_whitelist": synapse.workload._create_ip_range_whitelist(ip_range_whitelist),
```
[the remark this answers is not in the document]
amandahla: Good point, changed.

tests/unit/test_synapse_workload.py
```diff
@@ -129,11 +129,12 @@ def test_enable_ip_range_whitelist_success(ip_range_whitelist: str, harness: Har
 
     with open(config_path, encoding="utf-8") as config_file:
         content = yaml.safe_load(config_file)
+        expected_ip_range_whitelist = [item.strip() for item in ip_range_whitelist.split(",")]
```
jdkandersson, line 132: I would suggest making this a hardcoded value in the param since this is essentially doing the same thing as the code. For example:

```python
@pytest.mark.parametrize(
    "ip_range_whitelist, expected_ip_range_whitelist",
    [
...
        pytest.param(
            "10.10.10.10 ",
            ["10.10.10.10"],
            id="single_item_trailing_whitespace",
            marks=pytest.mark.xfail(strict=True),
        ),
...
    ],
)
def test_enable_ip_range_whitelist_success(ip_range_whitelist: str, expected_ip_range_whitelist: list[str], harness: Harness):
```

That way we clearly state the input and expected output for each test and avoid accidental bugs in the test code leading to tests unexpectedly passing

nrobinaubertin (dismissed):

src/synapse/workload.py
```diff
@@ -366,6 +387,41 @@ def enable_allow_public_rooms_over_federation(container: ops.Container) -> None:
...
+def _create_ip_range_whitelist(ip_range_whitelist: str) -> list[str]:
+    """Format IP range whitelist.
+
+    Args:
+        ip_range_whitelist: ip_range_whitelist configuration.
+
+    Returns:
+        IP range whitelist as expected by Synapse or None.
+    """
+    return [item.strip() for item in ip_range_whitelist.split(",")]
```
yanksyoon, line 399: A minor question, would this require to be a list, or could immutable tuple suffice?

yanksyoon (approved):
LGTM!

src/synapse/workload.py
amandahla: Since the values will not be modified, it could be a tuple. I'll modify that in this PR https://github.com/canonical/synapse-operator/pull/119
thanks!

tests/unit/test_synapse_workload.py
amandahla: Thanks,I'll modify that and for following tests in this PR that is related to this configuration
https://github.com/canonical/synapse-operator/pull/119

config.yaml
amandahla: I'll mark this as resolved but feel free to re-open. I can modify that in this ongoing PR
https://github.com/canonical/synapse-operator/pull/119

PR #127, opened 2023-12-14
Add SMTP via the smtp-integrator charm

javierdelapuente:
### Overview

Currently, to configure SMTP in synapse-operator you need to set configuration options. The change allows retrieving SMTP configuration from the SMTP integrator charm.

### Rationale

Leverage the new SMTP integrator charm. As an added benefit, SMTP configuration can be shared by several applications and secrets can be used.

### Module Changes

- New SMTP observer module
- Changes in state. Added SMTPConfiguration
- Changes in charm
- Changes in pebble, added `enable_smtp`
- Changes in workload to write the new configuration to Synapse configuration file

### Library Changes

Library `charms.smtp_integrator.v0.smtp` is included to make use of the `smtp-integrator` charm.

### Checklist


src/smtp_observer.py
```diff
@@ -0,0 +1,147 @@
...
+        self._charm.unit.status = ops.ActiveStatus()
+
+    @property
+    def _pebble_service(self) -> typing.Any:
+        """Return instance of pebble service.
+
+        Returns:
+            instance of pebble service or none.
+        """
+        return getattr(self._charm, "pebble_service", None)
```
yanksyoon, line 147: I think it might help to have this be injected if circular dependency is not a problem.

src/synapse/workload.py
```diff
@@ -601,28 +605,35 @@ def enable_smtp(container: ops.Container, charm_state: CharmState) -> None:
...
+            current_yaml["email"]["smtp_pass"] = smtp_config["password"]
+        current_yaml["email"]["enable_tls"] = smtp_config["enable_tls"]
+        current_yaml["email"]["force_tls"] = smtp_config["force_tls"]
+        current_yaml["email"]["require_transport_security"] = smtp_config[
+            "require_transport_security"
+        ]
         container.push(SYNAPSE_CONFIG_PATH, yaml.safe_dump(current_yaml))
     except ops.pebble.PathError as exc:
-        raise WorkloadError(str(exc)) from exc
+        raise EnableSMTPError(str(exc)) from exc
```
yanksyoon, line 636:
```suggestion
        raise EnableSMTPError("Error reading synpase configuration file (path error)") from exc
```

yanksyoon (approved):
LGTM! Minor questions to understand synapse better :D
(I think it might help if the LICENSE header changes were in a separate PR :D)

src/charm.py
```diff
@@ -38,6 +39,8 @@
 class SynapseCharm(ops.CharmBase):
     """Charm the service."""
 
+    # pylint: disable=too-many-instance-attributes
```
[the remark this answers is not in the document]
javierdelapuente: By default there is a maximum of 7, with the new one there are 8...

javierdelapuente:
> LGTM! Minor questions to understand synapse better :D (I think it might help if the LICENSE header changes were in a separate PR :D)

Totally right, sorry for the LICENSE changes, I updated them to pass the tests

src/smtp_observer.py
```diff
@@ -0,0 +1,147 @@
...
+    def get_relation_as_smtp_conf(self) -> Optional[SMTPConfiguration]:
+        """Get SMTP data from relation.
+
+        Returns:
+            Dict: Information needed for setting environment variables.
+        """
+        try:
+            relation_data: Optional[SmtpRelationData] = self.smtp.get_relation_data()
+        except ValidationError:
+            # ValidationError happens in the smtp(_legacy)relation_created event, as
```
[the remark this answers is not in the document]
javierdelapuente: Sure!

src/smtp_observer.py
javierdelapuente: As it is now, observers are created first, from them the state is built and then the pebble service gets the state injected. I believe this should be better refactored in another PR, changing all the observers.

See https://github.com/canonical/synapse-operator/blob/34c5118a4fc8a070cb6d43a82436e82348fea605/src/charm.py#L51-L64

src/smtp_observer.py
```diff
@@ -0,0 +1,147 @@
...
+        self.model.unit.status = ops.MaintenanceStatus("Preparing the SMTP integration")
+        logger.debug("_on_smtp_data_available: Enabling SMTP")
+        self._enable_smtp()
+
+    def _enable_smtp(self) -> None:
+        """Enable  SMTP."""
+        container = self._charm.unit.get_container(synapse.SYNAPSE_CONTAINER_NAME)
+        if not container.can_connect() or self._pebble_service is None:
+            self._charm.unit.status = ops.MaintenanceStatus("Waiting for Synapse pebble")
+            return
```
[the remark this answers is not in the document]
javierdelapuente: Thanks for the suggestion. I believe you are right, if the event `smtp_data_available` is not processed, it will not be processed until another event that changes the configuration is sent. Added `defer`.

src/synapse/workload.py
javierdelapuente: Reviewing other examples I believe it is that way because the message is set for the blocked status in Juju, and that way it is clearer what happened (it could be because of reading, writing, permissions...).

src/pebble.py
```diff
@@ -131,6 +132,22 @@ def enable_saml(self, container: ops.model.Container) -> None:
...
+
+        Raises:
+            PebbleServiceError: if something goes wrong while interacting with Pebble.
+        """
+        try:
+            logger.debug("pebble.enable_smtp: Enabling SMTP")
+            synapse.enable_smtp(container=container, charm_state=self._charm_state)
+            self.restart_synapse(container)
+        except (synapse.WorkloadError, ops.pebble.PathError) as exc:
+            raise PebbleServiceError(str(exc)) from exc
```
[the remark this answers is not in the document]
javierdelapuente: Reviewing other examples I believe it is that way because the message is set for the blocked status in Juju, and that way it is clearer what happened (a file error, a workload problem...)

src/charm_state.py
```diff
@@ -105,24 +95,24 @@ class Config:  # pylint: disable=too-few-public-methods
...
+            notif_from: the notif_from current value.
             values: values already defined.
 
         Returns:
-            The default value for smtp_notif_from if not defined.
+            The default value for notif_from if not defined.
         """
         server_name = values.get("server_name")
-        if smtp_notif_from is None and server_name:
+        if notif_from is None and server_name:
```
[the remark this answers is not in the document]
javierdelapuente: Having a look at [synapse documentation](https://matrix-org.github.io/synapse/latest/usage/configuration/config_documentation.html)) `notif_from` is mandatory is emails are used, so I think both approaches will fail (as server_name is not a valid email). I'd rather leave it that way, so it is obvious that the user wanted to set it to the empty string for whatever reason.

src/charm_types.py
```diff
@@ -36,3 +36,26 @@ class SAMLConfiguration(typing.TypedDict):
 
     entity_id: str
     metadata_url: str
+
+
+class SMTPConfiguration(typing.TypedDict):
```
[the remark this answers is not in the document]
javierdelapuente: That solution is cleaner than the TypedDict. Changed!

src/charm_state.py
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
```
yanksyoon, line 1: Irrelevant to this PR but I think a shebang line doesn't need to be here  :)

tests/unit/test_smtp_observer.py
```diff
@@ -0,0 +1,109 @@
...
+            "password": "SECRET",  # nosec
+        },
+        SMTPConfiguration(
+            enable_tls=True,
+            force_tls=True,
+            require_transport_security=True,
+            host="127.0.0.1",
+            port=587,
+            user="username",
+            password="SECRET",  # nosec
```
yanksyoon, line 67: I think the current team's convention on these are to use `secrets.token_hex(16)` so that we don't need to use `# nosec` :D

tests/unit/test_smtp_observer.py
```diff
@@ -0,0 +1,109 @@
...
+
+# pylint: disable=protected-access
+
+import pytest
+from charms.smtp_integrator.v0.smtp import AuthType, TransportSecurity
+from ops.testing import Harness
+
+from charm_types import SMTPConfiguration
+
+RELATION_DATA_TO_SMTP_CONF_PARAMS = [
```
yanksyoon, line 14: Perhaps this might be better as a fixture or as a function that generates these to prevent code pollution :D

yanksyoon (approved):
LGTM, minor things that I think can be in a follow up PR! Thanks for the changes :D

config.yaml
```diff
@@ -36,6 +36,10 @@ options:
       Comma separated list of IP address CIDR ranges that should be allowed for
       federation, identity servers, push servers, and for checking key validity
       for third-party invite events.
+  notif_from:
+    type: string
+    description: defines the "From" address to use when sending emails.
+      It must be set if email sending is enabled. Defaults to server_name.
```
amandahla, line 42: - Instead of "email sending" something like "SMTP integration"?
- I think we discussed that before the break: isn't this description mistaken?

From the docs, the default would be "Matrix", right?
https://matrix-org.github.io/synapse/latest/usage/configuration/config_documentation.html#email

src/smtp_observer.py
```diff
@@ -0,0 +1,162 @@
...
+        Arguments:
+            relation_data: The relation data from where to extract the password
+
+        Returns:
+            the password or None if no password found
+        """
+        # If the relation data password_id exists, that means that
+        # Juju version is >= 3.0 and secrets are used for the password.
+        # Otherwise, use the field password as a fallback
+        if relation_data.password_id:
```
amandahla, line 116: You can use the JUJU_HAS_SECRETS variable to make it clear.

src/smtp_observer.py
```diff
@@ -0,0 +1,162 @@
...
+            secret = self.model.get_secret(id=relation_data.password_id)
```
amandahla, line 117: I'm considering what would happen if this raises SecretNotFound error. Since there is no action to be taken, we could let this put the charm in an error state if it happens.

src/smtp_observer.py
```diff
@@ -0,0 +1,162 @@
...
+            event: The event that triggered enabling smtp.
+
+        """
+        container = self._charm.unit.get_container(synapse.SYNAPSE_CONTAINER_NAME)
+        if not container.can_connect() or self._pebble_service is None:
+            self._charm.unit.status = ops.MaintenanceStatus("Waiting for Synapse pebble")
+            # If this event is not deferred, enable_smtp will not be called and
+            # the configuration will not be changed until another event that changes
+            # the configuration is called.
+            event.defer()
```
amandahla, line 146: Usually is not advisable to use defer. See [here](https://github.com/canonical/operator/issues/966) and [here](https://juju.is/docs/sdk/deferring-events-details-and-dilemmas).
In this case, since for every configuration change the Synapse is restarted, this might cause unexpected restarts during the charm lifecycle.

mthaddon: From what I can see, the `enable_smtp` method is called as part of `src/pebble.py` in `change_config`. `_on_synapse_pebble_ready` calls `src/charm.py`'s `change_config`, and this calls `self.pebble_service.change_config(container)`, so it looks to me like this would happen whenever pebble is ready. Have we observed behaviour that suggests this isn't the case (i.e. if pebble isn't ready when we run `src/smtp_observer.py`'s `_enable_smtp` then `enable_smtp` isn't run as part of the pebble ready hook that fires later)?

javierdelapuente: You are right, it looks like it will work on the next pebble ready. Removed defer call.

config.yaml
javierdelapuente: Changed to SMTP integration as it is clearer than "email sending". I did not add the "%(app)'s" info as app_name is not set in the charm (and as said, just defaults to Matrix)

src/smtp_observer.py
javierdelapuente: I thought about that, however in the smtp-integrator charm, there are [two relations](https://github.com/canonical/smtp-integrator-operator/blob/main/metadata.yaml), `smtp` and `smtp-legacy`. `smtp` works only if juju has secrets, but `smtp-legacy` works in all cases. Using juju secrets will break `smtp-legacy`. WDYT?

src/smtp_observer.py
javierdelapuente: Not really sure. Could it happen that there is a relation but the owner (smtp integrator charm)  deletes the secret? If that happens, would just be an option to put the charm in error state until the integration is removed?

tests/unit/test_smtp_observer.py
javierdelapuente: Thanks for the advice!

tests/unit/test_smtp_observer.py
javierdelapuente: If you agree a would leave it there, as putting it has what the tests are really checking (I tried to put it in a function, but was not improving...)

src/smtp_observer.py
amandahla: Actually I thougt about just replacing "if relation_data.password_id" for "if JUJU_HAS_SECRETS". Wouldnt that work?

src/smtp_observer.py
amandahla: I think in this case makes more sense to let the Exception raise and create an error.

tests/unit/test_smtp_observer.py
amandahla: Since there are mostly no data changes from the relation to the configuration, is this test needed?

lib/charms/smtp_integrator/v0/smtp.py
```diff
@@ -0,0 +1,347 @@
...
+
+    def _on_relation_changed(self, event: ops.RelationChangedEvent) -> None:
+        """Event emitted when the relation has changed.
+
+        Args:
+            event: event triggering this handler.
+        """
+        assert event.relation.app
+        relation_data = event.relation.data[event.relation.app]
+        if relation_data:
```
amandahla, line 317: Isn't this validation done by the SMTP integrator charm?

amandahla:
Getting there :rocket: Left a few comments

src/charm_state.py
```diff
@@ -105,24 +93,24 @@ class Config:  # pylint: disable=too-few-public-methods
 
         extra = Extra.allow
 
-    @validator("smtp_notif_from", pre=True, always=True)
+    @validator("notif_from", pre=True, always=True)
     @classmethod
-    def set_default_smtp_notif_from(
-        cls, smtp_notif_from: typing.Optional[str], values: dict
+    def set_default_notif_from(
```
yanksyoon, line 98: Missed this during the initial review, would this function better be named something like
```suggestion
    def get_default_notif_from(
```
since it doesn't quite set any class attributes?

src/smtp_observer.py
```diff
@@ -0,0 +1,148 @@
...
+        return relation_data.password
+
+    def _on_smtp_relation_data_available(self, _: SmtpDataAvailableEvent) -> None:
+        """Handle SMTP data available."""
+        self.model.unit.status = ops.MaintenanceStatus("Preparing the SMTP integration")
+        logger.debug("_on_smtp_data_available: Enabling SMTP")
+        self._enable_smtp()
+
+    def _enable_smtp(self) -> None:
+        """Enable  SMTP."""
```
yanksyoon, line 129:
```suggestion
        """Enable SMTP."""
```

yanksyoon (approved):
LGTM!

tests/unit/test_smtp_observer.py
javierdelapuente: it tests the combinations for tls and transport_security and for auth type if plain, user and password are used. If I remove the test, these code paths will not be tested (or I will have to create several near equal tests otherwise).

src/smtp_observer.py
javierdelapuente: Until  [PR](https://github.com/canonical/smtp-integrator-operator/pull/29) smtp-integrator with secrets was not working correctly.

It now works, so I have updated the tests so that now it runs tests for `smtp` and `smtp-legacy` endpoints and updated the integration tests to run in Juju 3.

If only JUJU_HAS_SECRETS is checked, then this will not work in Juju 3: `juju integrate smtp-integrator:smtp-legacy synapse:smtp`, as secrets are not used in that endpoint in the smtp-integrator charm, so the test will fail. I suppose this could be used for example for offers.

However, now that smtp integrator works with secrets, not checking the data bag fields and instead checking with just JUJU_HAS_SECRETS is possible, but then it will not be possible to use `smtp-legacy` in Juju 3. I do not have a strong opinion, do you prefer to drop the support for `smtp-legacy`?

lib/charms/smtp_integrator/v0/smtp.py
javierdelapuente: A similar check is done in get_relation_as_smtp_conf:
```
    def get_relation_as_smtp_conf(self) -> Optional[SMTPConfiguration]:
...
        if relation_data is None:
            return None
```

However, that's is called when constructing the charm state in all hooks. Not sure if that answers the question

tests/unit/test_smtp_observer.py
amandahla: Well, IMHO this would be more in the scope of the SMTP integrator charm but I don't have a strong opinion about it.

src/smtp_observer.py
amandahla: Lets support both for now since the staging env isn't Juju 3 yet.

lib/charms/smtp_integrator/v0/smtp.py
amandahla: Sorry, you can ignore this reivew, is in the library ;-)

amandahla (approved):
LGTM :rocket:

synapse_rock/rockcraft.yaml
```diff
@@ -48,7 +48,7 @@ parts:
             - python3
             - xmlsec1
         stage-snaps:
-            - mjolnir/latest/edge
+            - mjolnir/latest/stable
```
amandahla, line 51: You can replace it for edge now, snap fixed :)

src/smtp_observer.py
```diff
@@ -0,0 +1,148 @@
...
+            return None
+
+        if relation_data is None:
+            return None
+
+        user = None
+        password = None
+        if relation_data.auth_type == AuthType.PLAIN:
+            user = relation_data.user
+            password = self._get_password_from_relation_data(relation_data)
```
jdkandersson, line 79: This is where we should consider what kind of auth is ok. When we designed the SMTP integrator, we decided to be more flexible and let the users of the integration decide. In general, we should be careful with that we support with security since it reflects poorly on the charming ecosystem if within it it is common to have insecure deployments. Any thoughts @gregory-schiano?

tests/integration/test_charm.py
```diff
@@ -352,29 +352,56 @@ async def test_saml_auth(  # pylint: disable=too-many-locals
...
+        await model.block_until(lambda: "smtp-integrator" not in model.applications, timeout=60)
+        await model.wait_for_idle(status=ACTIVE_STATUS_NAME, idle_period=5)
+
+    smtp_integrator_app = await model.deploy(
+        "smtp-integrator",
+        channel="latest/edge",
+        config={
+            "auth_type": "plain",
+            "host": "127.0.0.1",
+            "password": "SECRET",
```
jdkandersson, line 386: I'm suprised that bandit doesn't complain, why not just use the `secrets` module to generate something?

tests/unit/test_smtp_observer.py
```diff
@@ -0,0 +1,109 @@
...
+            host="127.0.0.1",
+            port=587,
+            user="username",
+            password="SECRET",  # nosec
+        ),
+    ),
+]
+
+
+@pytest.mark.parametrize("relation_data, expected_config", RELATION_DATA_TO_SMTP_CONF_PARAMS)
```
jdkandersson, line 73: You might be interested in taking a look at the following pattern for how to parametrize tests, do you think that would be helpful here? https://github.com/canonical/discourse-gatekeeper/blob/d49ebe76aaec9fc64947a13728edbb650e69048d/tests/unit/test_index.py#L183

tests/unit/test_smtp_observer.py
```diff
@@ -0,0 +1,109 @@
...
+        ),
+    ),
+    (
+        {
+            "host": "127.0.0.1",
+            "port": "25",
+            "auth_type": AuthType.PLAIN,
+            "transport_security": TransportSecurity.STARTTLS,
+            "user": "username",
+            "password": "SECRET",  # no sec
```
jdkandersson, line 39: Similar for these as before, why not just use the `secrets` module to generate a password?

.github/workflows/integration_test.yaml
```diff
@@ -12,5 +12,7 @@ jobs:
       chaos-enabled: false
       chaos-experiments: pod-delete
       trivy-image-config: "trivy.yaml"
-      self-hosted-runner: true
-      self-hosted-runner-label: "edge"
+      # self-hosted-runner: true
+      # self-hosted-runner-label: "edge"
```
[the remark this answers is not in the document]
javierdelapuente: Sure! remove commented code as we don't know when we could use that.

config.yaml
```diff
@@ -36,6 +36,10 @@ options:
       Comma separated list of IP address CIDR ranges that should be allowed for
       federation, identity servers, push servers, and for checking key validity
       for third-party invite events.
+  notif_from:
```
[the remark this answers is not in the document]
javierdelapuente: As `notif_from` is the original name in Synapse configuration, I'd rather leave that name for consistency.

docs/tutorial/getting-started.md
```diff
@@ -140,4 +167,4 @@ model environment you created during this tutorial, use the following command.
 
 ```
 juju destroy model synapse-tutorial -y
-```
\ No newline at end of file
+```
```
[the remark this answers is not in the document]
javierdelapuente: Removed

docs/tutorial/getting-started.md
```diff
@@ -3,6 +3,7 @@
 ## What you’ll do
 - Deploy the Synapse charm.
 - Integrate with the PostgreSQL K8s charm.
+- Integrate with the smtp-integrator charm.
```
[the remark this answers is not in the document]
javierdelapuente: Moved all the smtp integration to a new howto doc

docs/tutorial/getting-started.md
```diff
@@ -61,6 +62,32 @@ synapse                 3.2                           active      1  synapse
 
 The deployment is complete when the status is `Active`.
 
+## Integrate with the smtp-integrator charm
+
+For synapse to use smtp, it uses the smtp-integrator charm.
+```
+juju deploy smtp-integrator --channel edge
+juju config smtp-integrator host=smtp.example.com port=25 user=alice password=whatever auth_type=plain transport_security=none domain=example.com
```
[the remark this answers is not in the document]
javierdelapuente: Moved to the howto.  Set password to <smtp auth password> and transport security as tls.

docs/tutorial/getting-started.md
```diff
@@ -61,6 +62,32 @@ synapse                 3.2                           active      1  synapse
...
+```
+
+Configure the "From" mail for Synapse with:
+```
+juju config synapse notif_from=no-reply@example.com
```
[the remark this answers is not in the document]
javierdelapuente: Done

src/charm.py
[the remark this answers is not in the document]
javierdelapuente: Not sure this is the PR to take action about that. Instance attributes in this case are observers, libraries and state (the addition in this PR is a new observer) and that would require some refactor that I think is not in the scope of the PR. Added explanatory comment for that.

tests/integration/test_charm.py
javierdelapuente: True. Changed to secrets

tests/unit/test_smtp_observer.py
javierdelapuente: done

tests/unit/test_smtp_observer.py
javierdelapuente: Thanks for the info. Changed to a function with `pytest.param`.

src/smtp_observer.py
gregory-schiano: I'm fine not supporting plain auth in synapse charm as it supports STARTTLS and TLS which are more secure standards.
We could mark the status as Blocked with a message stating that we don't support plain auth

merkata (approved):
LGTM! :rocket:

amandahla (approved):
LGTM

merkata (approved):

PR #130, opened 2024-01-02
Create integration test to refresh the charm

javierdelapuente:
### Overview

This PR creates a new integration test to check that refreshing Synapse with Mjolnir enabled works correctly.

To do that, the charm is first installed from Charmhub and then refreshed with the local charm. 

Integrations tests that were using the charm from Charmhub were blocked until https://github.com/canonical/synapse-operator/pull/133 was completed. 

### Rationale

Check that Synapse charm with Mjolnir is refreshed correctly. Original reason in https://github.com/canonical/synapse-operator/pull/48.

### Checklist


gtrkiller (dismissed):
LGTM

tests/integration/conftest.py
```diff
@@ -107,8 +106,8 @@ async def synapse_app_fixture(
     use_existing = pytestconfig.getoption("--use-existing", default=False)
     if use_existing:
         return model.applications[synapse_app_name]
-    if synapse_app_refresh_name in model.applications:
-        await model.remove_application(synapse_app_refresh_name, block_until_done=True)
+    if synapse_app_charmhub_name in model.applications:
+        await model.remove_application(synapse_app_charmhub_name, block_until_done=True)
```
jdkandersson, line 110: What would be the reason that the charm is already deployed in the model? I would have expected any previous tests to remove the application from the model before finishing?

tests/integration/helpers.py
```diff
@@ -0,0 +1,90 @@
...
+    access_token: str,
+):
+    """Create "moderators" room in Synapse
+
+    Args:
+        synapse_ip: Synapse IP
+        access_token: Access token for Synapse to create the room
+
+    """
+    authorization_token = f"Bearer {access_token}"
```
jdkandersson, line 24: Perhaps `authorization_header`?

tests/integration/helpers.py
```diff
@@ -0,0 +1,90 @@
...
+    res = requests.post(
+        f"http://{synapse_ip}:8080/_matrix/client/v3/createRoom",
+        json=room_body,
+        headers=headers,
+        timeout=5,
+    )
+    res.raise_for_status()
+
+
+def get_access_token(synapse_ip, user_username, user_password):
```
cbartz, line 43:
```suggestion
def get_access_token(synapse_ip, user_username, user_password) -> str:
```

tests/integration/helpers.py
```diff
@@ -0,0 +1,90 @@
...
+    )
+    res.raise_for_status()
+    access_token = res.json()["access_token"]
+    assert access_token
+    return access_token
+
+
+async def register_user(
+    synapse_app: Application, user_username: str
+) -> typing.AsyncGenerator[str, None]:
```
cbartz, line 72:
```suggestion
) -> str:
```

cbartz (dismissed):
Looks good, I have some typehint comments.

tests/integration/helpers.py
javierdelapuente: Thanks for the advice. Fixed in another commit

tests/integration/helpers.py
javierdelapuente: Thanks for the advice. Fixed in another commit

tests/integration/conftest.py
javierdelapuente: Thanks for the advice. I see no reason to remove the application in there. If at some point there are conflicts, the tests can be put in another module or change the scope for the fixture.

Removed that code.

tests/integration/conftest.py
```diff
@@ -310,10 +307,12 @@ def user_username_fixture() -> typing.Generator[str, None, None]:
 
 
 @pytest_asyncio.fixture(scope="module", name="user_password")
-async def user_password_fixture(
-    synapse_app: Application, user_username: str
-) -> typing.AsyncGenerator[str, None]:
-    """Return the a username to be created for tests."""
+async def user_password_fixture(synapse_app: Application, user_username: str) -> str:
+    """Return the a username to be created for tests.
```
cbartz, line 311:
```suggestion
    """Register a user and return the new password.
```

tests/integration/helpers.py
javierdelapuente: I would rather leave that name as it it used in several places and I am not sure it is really wrong. Maybe it could be reasonable for another PR to look for other good names like [credentials](https://datatracker.ietf.org/doc/html/rfc6750#section-2.1) or "bearer authentication".

cbartz (approved):

tests/integration/helpers.py
```diff
@@ -0,0 +1,86 @@
...
+        f"http://{synapse_ip}:8080/_matrix/client/r0/login",
+        json={
+            "identifier": {"type": "m.id.user", "user": user_username},
+            "password": user_password,
+            "type": "m.login.password",
+        },
+        timeout=5,
+    )
+    res.raise_for_status()
+    access_token = res.json()["access_token"]
```
Thanhphan1147, line 63: I think you can use `res.json().get("access_token")` here since using brackets will raise `KeyError` if the key is not there

tests/integration/test_charm.py
```diff
@@ -30,22 +31,54 @@
...
     get_unit_ips: typing.Callable[[str], typing.Awaitable[tuple[str, ...]]],
+    synapse_charm: str,
+    synapse_image: str,
+    synapse_nginx_image: str,
 ):
     """
-    arrange: build and deploy the Synapse charm.
-    act: send a request to the Synapse application managed by the Synapse charm.
-    assert: the Synapse application should return a correct response.
+    arrange: build and deploy the Synapse charm from charmhub and enable Mjolnir.
```
Thanhphan1147, line 44: I think we don't deploy the charm from charm hub but rather the locally built one, if that's the case then perhaps you can update this line.

javierdelapuente: Not really, the charm is deployed from Charmhub in the fixture (see https://github.com/canonical/synapse-operator/blob/be4d06e64a02c073575f863e4622fc6036cfb239/tests/integration/conftest.py#L139-L146), and then refreshed with the local one.

tests/integration/helpers.py
javierdelapuente: Sure! That way it will fail in the assert which looks clearer.

cbartz (approved):

tests/integration/helpers.py
```diff
@@ -0,0 +1,86 @@
...
+
+    Returns:
+        The new password for the user
+    """
+    action_register_user: Action = await synapse_app.units[0].run_action(
+        "register-user", username=user_username, admin=True
+    )
+    await action_register_user.wait()
+    assert action_register_user.status == "completed"
+    assert action_register_user.results["register-user"]
```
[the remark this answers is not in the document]
javierdelapuente: Sure. Done!

Thanhphan1147 (approved):
Lgtm!

PR #139, opened 2024-01-17
Add Synapse backup s3-integration

javierdelapuente:
Applicable spec: ISD095

### Overview

This PR creates a new integration with S3 with the goal of being able to create, list, restore
 and delete backups. Those tasks will be done in future PRs. In this one the goal is
to create the integration with a simple test to check that the bucket exists when s3-integration 
is integrated with Synapse. All this functionality will be refined in the next PRs.

### Rationale

Backups will me stored in S3, so the integration is needed.

### Juju Events Changes

`on_credentials_changed` and `on_credentials_gone` from the `charms.data_platform_libs.v0.s3` library.

### Module Changes

New module `backup` that will handle backup functionality and will use S3 credentials.

### Library Changes

Added  `charms.data_platform_libs.v0.s3` library.

### Checklist


docs/reference/integrations.md
```diff
@@ -94,3 +94,17 @@ For the smtp-integrator, insecure configurations with `transport_security=none`
...
 
 See more information in [Charm Architecture](https://charmhub.io/synapse/docs/explanation-charm-architecture).
+
+
+### s3-backup-parameters
+
+_Interface_: s3
+_Supported charms_: [s3-integrator](https://charmhub.io/s3-integrator/)
+
+In order to perform backups, Synapse has to be integrated with the s3-integrator charm using the
```
arturo-seijas, line 104: Can you specify what will be backed up?

src/backup.py
```diff
@@ -0,0 +1,133 @@
...
+from botocore.exceptions import ConnectionError as BotoConnectionError
+from charms.data_platform_libs.v0.s3 import CredentialsChangedEvent, S3Requirer
+from ops.framework import Object
+from pydantic import BaseModel, Field, validator
+
+logger = logging.getLogger(__name__)
+
+S3_CANNOT_ACCESS_BUCKET = "Backup: S3 bucket does not exist or cannot be accessed"
+S3_INVALID_CONFIGURATION = "Backup: S3 configuration is invalid"
+BACK_UP_STATUS_MESSAGES = (S3_INVALID_CONFIGURATION, S3_CANNOT_ACCESS_BUCKET)
```
arturo-seijas, line 20: s/BACK_UP_STATUS_MESSAGES/BACKUP_STATUS_MESSAGES?

src/backup.py
```diff
@@ -0,0 +1,133 @@
...
+        s3_uri_style: The S3 protocol specific bucket path lookup type.
+    """
+
+    access_key: str = Field(alias="access-key")
+    secret_key: str = Field(alias="secret-key")
+    region: Optional[str] = Field(default=None)
+    bucket: str
+    endpoint: Optional[str]
+    path: str = Field(default="")
+    s3_uri_style: str = Field(alias="s3-uri-style", default="host")
```
arturo-seijas, line 42: s/s3_uri_style/s3_uri?

src/backup.py
```diff
@@ -0,0 +1,133 @@
...
+            self._charm.unit.status = ops.BlockedStatus(S3_CANNOT_ACCESS_BUCKET)
+        logger.info("S3 backup correctly configured")
+
+    def _on_s3_credential_gone(self, _: CredentialsChangedEvent) -> None:
+        """Handle s3 credentials gone to reset unit status if it is now correct."""
+        if (
+            isinstance(self._charm.unit.status, ops.BlockedStatus)
+            and self._charm.unit.status.message in BACK_UP_STATUS_MESSAGES
+        ):
+            self._charm.unit.status = ops.ActiveStatus()
```
arturo-seijas, line 103: Doesn't this mean that the charm will reach active status when credentials are removed even if the params are still invalid?

src/backup.py
```diff
@@ -0,0 +1,133 @@
...
+    """
+    session = boto3.session.Session(
+        aws_access_key_id=s3_parameters.access_key,
+        aws_secret_access_key=s3_parameters.secret_key,
+        region_name=s3_parameters.region,
+    )
+    try:
+        s3 = session.resource("s3", endpoint_url=s3_parameters.endpoint)
+    except BotoCoreError:
+        logger.exception("Failed to create S3 session")
```
arturo-seijas, line 123: It might be helpful to print the exception trace too

src/backup.py
```diff
@@ -0,0 +1,133 @@
...
+        return False
+    bucket = s3.Bucket(s3_parameters.bucket)
+    try:
+        bucket.meta.client.head_bucket(Bucket=s3_parameters.bucket)
+    except (ClientError, BotoConnectionError):
+        logger.exception(
```
arturo-seijas, line 129: Same here

metadata.yaml
```diff
@@ -65,6 +65,10 @@ requires:
         interface: smtp
         limit: 1
         optional: true
+    s3-backup-parameters:
```
cbartz, line 68: Maybe just `s3-backup` or `backup`? I believe it is not relevant to a user that the integration will exchange parameters.

docs/reference/integrations.md
```diff
@@ -94,3 +94,17 @@ For the smtp-integrator, insecure configurations with `transport_security=none`
...
+endpoint s3-backup-parameters. Backups will be stored, listed and recovered from the location
+indicated in the S3 compatible object storage provider configuration provided by the integration.
+
+Example s3-backup-parameters integrate command: `juju integrate synapse:s3-backup-parameters s3-integrator`
+
+See more information in [Charm Architecture](https://charmhub.io/synapse/docs/explanation-charm-architecture).
```
cbartz, line 110: Shouldn't the architecture documents be updated when you refer to them?

src/backup.py
```diff
@@ -0,0 +1,133 @@
...
+    region: Optional[str] = Field(default=None)
+    bucket: str
+    endpoint: Optional[str]
+    path: str = Field(default="")
+    s3_uri_style: str = Field(alias="s3-uri-style", default="host")
+
+    @validator("region", always=True)
+    @classmethod
+    def check_region_or_endpoint_set(cls, region: str, values: dict[str, Any]) -> str:
+        """Validate that eather that region or endpoint are set.
```
cbartz, line 47:
```suggestion
        """Validate that either that region or endpoint is set.
```

src/backup.py
```diff
@@ -0,0 +1,133 @@
...
+    def _on_s3_credential_gone(self, _: CredentialsChangedEvent) -> None:
+        """Handle s3 credentials gone to reset unit status if it is now correct."""
+        if (
+            isinstance(self._charm.unit.status, ops.BlockedStatus)
+            and self._charm.unit.status.message in BACK_UP_STATUS_MESSAGES
+        ):
+            self._charm.unit.status = ops.ActiveStatus()
+
+
+def can_use_bucket(s3_parameters: S3Parameters) -> bool:
```
cbartz, line 106: To be compliant with our [contributing guide](https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#function-and-method-Ordering) , the function should appear before it is used, so above `_on_s3_credential_changed`.

tests/unit/test_backup.py
```diff
@@ -0,0 +1,182 @@
...
+from botocore.exceptions import BotoCoreError, ClientError
+from ops.testing import Harness
+
+import backup
+
+
+def test_s3_relation_validation_fails_when_region_and_endpoint_not_set():
+    """
+    arrange: Create s3 relation data without region nor endpoint.
+    act: Create S3Paramaters pydantic BaseModel from relation data.
```
cbartz, line 21:
```suggestion
    act: Create S3Parameters pydantic BaseModel from relation data.
```

src/backup.py
```diff
@@ -0,0 +1,133 @@
...
+            ValueError: if the configuration is invalid.
+        """
+        endpoint = values.get("endpoint")
+        if not region and not endpoint:
+            raise ValueError('one of "region" or "endpoint" needs to be set')
+        return region
+
+
+class SynapseBackup(Object):
+    """Class to manage Synapse backups over S3."""
```
cbartz, line 66: Does this class really manage backups? I believe its more of an observer and should perhaps be renamed to `BackupObserver` (and the module too) to distinguish between buisness and charm/event logic. There seems to be a similar split for the database. See also https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619

cbartz (commented):
Looks good overall, left some comments.

src/backup.py
javierdelapuente: done!

src/backup.py
javierdelapuente: Good point. I have renamed the module and class and created a new module backup for the backup related logic.

src/backup.py
javierdelapuente: Not sure that would be clear, as a S3 URI is something like [S3://bucket-name/key-name](https://repost.aws/questions/QUFXlwQxxJQQyg9PMn2b6nTg/what-is-s3-uri-in-simple-storage-service)

arturo-seijas: So, bucket_pth or smth like that?

src/backup.py
javierdelapuente: `logger.exception` already logs the exception trace ([example execution](https://github.com/canonical/synapse-operator/actions/runs/7568727201/job/20612993165#step:13:518)). Not sure if that is what you are asking.

src/backup.py
javierdelapuente: logger.exception already logs the exception trace ([example execution](https://github.com/canonical/synapse-operator/actions/runs/7568727201/job/20612993165#step:13:518)). Not sure if that is what you are asking.

src/backup.py
javierdelapuente: Good point. However, reviewing the library, the event `credentials_gone` is emitted on `relation_broken`, so I understand that is reasonable. WDYT?

arturo-seijas: Won't running `_on_s3_credential_changed` handle the scenario I mention?

javierdelapuente: I included that logic in the `_on_s3_credential_changed` (so the unit can go back to active it the credentials are fixed without recreating the integration). Does that solve the issue you were seeing?

src/backup.py
cbartz: I think at this stage (without any real business logic) I would have just kept one module and seen `can_use_bucket` as a helper method to validate the relation data, but I suppose that is debatable.

src/backup.py
```diff
@@ -0,0 +1,86 @@
...
+        return False
+    bucket = s3.Bucket(s3_parameters.bucket)
+    try:
+        bucket.meta.client.head_bucket(Bucket=s3_parameters.bucket)
+    except (ClientError, BotoConnectionError):
+        logger.exception(
+            "Bucket %s doesn't exist or you don't have access to it.", s3_parameters.bucket
+        )
+        return False
+    return True
```
cbartz, line 86: Is it clear that a user can also write to a bucket in full? The current checks only see if it is accessible. Maybe there are ACLs that only allow read but not write.

src/backup.py
arturo-seijas: Makes sense

src/backup_observer.py
```diff
@@ -0,0 +1,67 @@
...
+            s3_parameters = S3Parameters(**self._s3_client.get_s3_connection_info())
+        except ValueError:
+            self._charm.unit.status = ops.BlockedStatus(S3_INVALID_CONFIGURATION)
+            return
+
+        if not can_use_bucket(s3_parameters):
+            self._charm.unit.status = ops.BlockedStatus(S3_CANNOT_ACCESS_BUCKET)
+            return
+
+        # At this point the S3 configuration is correct.
```
arturo-seijas, line 53: Why do you need this condition? Why not directly setting the status to Active?

src/backup.py
arturo-seijas: I think you need to execute all the logic in `_on_s3_credential_changed` whenever `self._s3_client.on.credentials_gone` is triggered

javierdelapuente: `credentials_gone` is emitted when the integration with the `s3-integrator` is [broken](https://github.com/canonical/synapse-operator/blob/c473364f8fc7a9ec4c4b58313cc0c371afeb4e96/lib/charms/data_platform_libs/v0/s3.py#L761). I assume that at that point there is no integration with the `s3-integrator` charm. Should the unit be blocked because of S3 issues when there is no integration?

arturo-seijas: Ok, that was not was I was expecting. It is a wrapper around "relation_broken". That's a bit weird, but ignore my comment

src/backup.py
javierdelapuente: I have been checking [`get_bucket_acl`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/get_bucket_acl.html) and testing agains localstack and minio. The responses contain all the grants for the bucket, and then it will be necessary to check it against the acess key. As the next PR will write to the bucket, do you think we could tackle it there?

src/backup_observer.py
javierdelapuente: I believe the unit can be blocked, for example because pebble is not ready, and in that case we would set it to active.

src/backup.py
```diff
@@ -0,0 +1,86 @@
...
+        Raises:
+            ValueError: if the configuration is invalid.
+        """
+        endpoint = values.get("endpoint")
+        if not region and not endpoint:
+            raise ValueError('one of "region" or "endpoint" needs to be set')
+        return region
+
+
+def can_use_bucket(s3_parameters: S3Parameters) -> bool:
```
amandahla, line 59: @gregory-schiano Out of the context of this PR: Do you think that charm integrators could provide a function to validate what they are providing? SMTP, SAML, this one S3...Otherwise, all the charms would need something like this before using the configuration the Integrator provides.

src/backup_observer.py
```diff
@@ -0,0 +1,67 @@
+# Copyright 2024 Canonical Ltd.
```
amandahla, line 1: Minor: It's not a documented pattern but I think it would be nice to keep the Observer's file name/class the same as the integration in the metadata.yaml

src/backup_observer.py
amandahla: I see that this might happen but I think we should set this to ActiveStatus for now since the observer is not dealing with the workload here.

tests/unit/test_backup_observer.py
```diff
@@ -0,0 +1,120 @@
...
+
+import boto3
+import ops
+import pytest
+from ops.testing import Harness
+
+import backup_observer
+
+
+def test_on_s3_credentials_changed_correct(harness: Harness, monkeypatch: pytest.MonkeyPatch):
```
amandahla, line 18: Do you think this and the next 3 tests could be just one using pytest.mark.parametrize?

src/backup.py
nrobinaubertin: Shouldn't they fill the relation data only if the service they provide can reasonably be validated ?

tests/integration/test_charm.py
```diff
@@ -543,3 +543,92 @@ async def test_anonymize_user(
...
+):
+    """
+    arrange: Synapse App deployed and s3-integrator deployed with bucket created.
+    act:  integrate s3-integrator with Synapse.
+    assert: Synapse gets into active status.
+    """
+    s3_integrator_app = await model.deploy(
+        "s3-integrator",
+        channel="latest/edge",
+        config={
```
jdkandersson, line 562: Couldn't the `s3_backup_configuration` just be passed here directly rather than re-mapping it?

tests/integration/test_charm.py
```diff
@@ -543,3 +543,92 @@ async def test_anonymize_user(
...
+            "bucket": s3_backup_configuration["bucket"],
+            "path": s3_backup_configuration["path"],
+            "region": s3_backup_configuration["region"],
+            "s3-uri-style": s3_backup_configuration["s3-uri-style"],
+        },
+    )
+    await model.wait_for_idle(apps=[s3_integrator_app.name], idle_period=5, status="blocked")
+    action_sync_s3_credentials: Action = await s3_integrator_app.units[0].run_action(
+        "sync-s3-credentials",
+        **{
```
jdkandersson, line 573: Why unpack when these could just be arguments?

tests/integration/test_charm.py
```diff
@@ -543,3 +543,92 @@ async def test_anonymize_user(
...
+
+    await model.wait_for_idle(
+        idle_period=30,
+        apps=[synapse_app.name, s3_integrator_app.name],
+        status=ACTIVE_STATUS_NAME,
+    )
+
+    # Test cleanup
+    await model.remove_application("s3-integrator")
+    await model.block_until(lambda: "s3-integrator" not in model.applications, timeout=60)
```
jdkandersson, line 592: The issue with putting test cleanup in the test code is that, if the test fails, the cleanup code is not executed. Clean up code should go into fixtures which will get executed even if the test fails

tests/unit/test_backup.py
```diff
@@ -0,0 +1,73 @@
...
+from unittest.mock import MagicMock
+
+import boto3
+import pytest
+from botocore.exceptions import BotoCoreError, ClientError
+
+import backup
+
+
+def test_s3_relation_validation_fails_when_region_and_endpoint_not_set():
```
jdkandersson, line 16: Are there any other validations we should check?

tests/unit/test_backup.py
```diff
@@ -0,0 +1,73 @@
...
+            "secret-key": token_hex(16),
+            "region": "eu-west-1",
+            "bucket": "bucket_name",
+        }
+    )
+    session = MagicMock()
+    session.resource = MagicMock(side_effect=BotoCoreError())
+    monkeypatch.setattr(boto3.session, "Session", MagicMock(return_value=session))
+
+    assert backup.can_use_bucket(s3_parameters) is False
```
jdkandersson, line 52: what about `assert not ...`?

src/backup.py
cbartz: You can address this in the next PR, but I think you will not see any problems if you are the owner of the bucket.
I think your checks cover most cases, but there may be edge cases (user has read permission but no write permission) or maybe even a false negative check (user has no `s3:ListBucket` permission but `s3:PutObject` permission).
Feel free to resolve this conversation.

tests/unit/test_backup_observer.py
javierdelapuente: Renamed integration from s3-backup to backup as suggested by @cbartz

src/backup_observer.py
javierdelapuente: Renamed the integration from s3-backup to backup as suggested by @cbartz

tests/integration/test_charm.py
javierdelapuente: done!

tests/integration/test_charm.py
javierdelapuente: Refactored to get the s3 integrator from a fixture

tests/unit/test_backup.py
javierdelapuente: Sure!

tests/integration/test_charm.py
javierdelapuente: Refactored a bit to get the credentials from a fixture. Unfortunately, the arguments in this case contain `-`, so they are not valid Python identifiers, and it is not possible to use them as arguments.

tests/unit/test_backup_observer.py
javierdelapuente: Joined. Not sure which option looks better (now it looks a bit like Go)

tests/unit/test_backup.py
javierdelapuente: Yes, the correct case. And a hidden bug was in them...

src/backup_observer.py
javierdelapuente: Changed so now the observer does not check that the previous status was blocked because of S3 problems.

We could review and in case refactor how the status is set in a next PR.

src/backup.py
javierdelapuente: I close it while I think of a better way of checking it (or if it is too complex let it this way as a best effort).

arturo-seijas (approved):

src/backup.py
gregory-schiano: If your question is, should I trust that integrator in validating that the provided information are effectively correct and working ?
Then I'd say indeed these data should be checked when configured on the integrator and the integrator be blocked if they're not.
But effectively I don't think these validations are in place.
That's a good suggestion for improvement

amandahla: Thanks, I'll close this conversation to not block the PR but let me know if there is some place to add suggestions to the architecture guide or if this could be a story.

amandahla (approved):
LGTM

PR #140, opened 2024-01-17
Add Synapse Stats Exporter

amandahla:
### Overview

In order to collect and monitor the number of users and rooms in the Synapse instance, this PR install the synapse-stats-exporter so the metrics can be extracted by Prometheus.

Also, it refactors how the charm handle the admin_access_token.

Before the refactor, the admin access token was handled only by the charm and used by Mjolnir.

But since the Stats Exporter needs it too, it was complex to let this in the charm instead of letting the Synapse module handle it.

So what happens now?

- start event is emitted meaning that (hopefully) the workload was started (not just the container)
- If the admin access token already exists (as a secret or as peer data), start the Stats Exporter and return
- If the admin access token doesn't exist, create it via create_admin_user and save it in the charm_state

- Mjolnir and Stats Exporter will get the admin_access_token via charm_state
- Also, Mjolnir will create the user using the create_user available in the synapse module instead of calling actions.register_user (that didn't make much sense)

This is how the modules depend on each other now:

image

### Rationale

Collect Synapse statistics.

### Juju Events Changes

N/A

### Module Changes

Pebble module has a new layer.

### Library Changes

N/A

### Checklist

Documentation should be updated after the exporter being available.


src/charm.py
```diff
@@ -181,6 +181,94 @@ def _on_synapse_nginx_pebble_ready(self, _: ops.HookEvent) -> None:
...
+
+    def _on_start(self, _: ops.HookEvent) -> None:
+        """Handle start event."""
+        if self.get_admin_access_token():
+            self._start_synapse_stats_exporter()
+            return
+        # Since Synapse is ready, the charm can create the admin access token
+        container = self.unit.get_container(synapse.SYNAPSE_CONTAINER_NAME)
+        if not container.can_connect():
+            self.unit.status = ops.MaintenanceStatus("Waiting for Synapse pebble")
```
javierdelapuente, line 261: Not sure handling this logic only in this event will be problematic. I just tested on my machine, and randomly the event `pebble_ready` comes after the `start` event. If that happens, then the exported will not be planned.

image

src/charm.py
```diff
@@ -107,7 +107,7 @@ def change_config(self) -> None:
             return
         self.model.unit.status = ops.MaintenanceStatus("Configuring Synapse")
         try:
-            self.pebble_service.change_config(container)
+            self.pebble_service.change_config(container=container)
```
merkata, line 110: Just out of curiosity, why the change here, is it purely cosmetic?

src/charm.py
```diff
@@ -181,6 +181,94 @@ def _on_synapse_nginx_pebble_ready(self, _: ops.HookEvent) -> None:
...
         self._set_unit_status()
 
+    def _get_peer_relation(self) -> typing.Optional[ops.Relation]:
+        """Get peer relation.
+
+        Returns:
+            Relation or not if is not found.
+        """
+        peer_relation = self.model.get_relation(PEER_RELATION_NAME)
+        return peer_relation
```
merkata, line 191: Do we usually do this? I'd rather `return self.model.get_relation(PEER_RELATION_NAME)` directly, though this is perfectly fine as well.

src/charm.py
```diff
@@ -181,6 +181,94 @@ def _on_synapse_nginx_pebble_ready(self, _: ops.HookEvent) -> None:
...
+            secret = self.app.add_secret({SECRET_KEY: access_token})
+            peer_relation.data[self.app].update({SECRET_ID: secret.id})
+        else:
+            logger.debug("Adding peer data")
+            peer_relation.data[self.app].update({SECRET_KEY: access_token})
+
+    def _on_start(self, _: ops.HookEvent) -> None:
+        """Handle start event."""
+        if self.get_admin_access_token():
+            self._start_synapse_stats_exporter()
```
merkata, line 256: Doesn'g mjolnir require an access token as well? Should it be added here?

src/mjolnir.py
```diff
@@ -57,11 +56,11 @@ def _admin_access_token(self) -> typing.Optional[str]:
         Returns:
             admin access token or None if fails.
         """
-        get_admin_access_token = getattr(self._charm, "get_admin_access_token", None)
-        if not get_admin_access_token:
+        access_token = self._charm_state.synapse_config.admin_access_token
+        if not access_token:
             logging.error("Failed to get method get_admin_access_token.")
```
merkata, line 61: That error can be reworked now due to the change.

src/synapse/admin.py
```diff
@@ -0,0 +1,74 @@
...
+    # access_token.
+    # But to do that it needs an admin user and we don't have one yet.
+    # So, to be on the safe side, the user name is randomly generated and if for any reason
+    # there is no access token on peer data/secret, another user will be created.
+    #
+    # Using 16 to create a random value but to  be secure against brute-force attacks,
+    # please check the docs:
+    # https://docs.python.org/3/library/secrets.html#how-many-bytes-should-tokens-use
+    username = token_hex(16)
+    return create_user(container, username=username, admin=True)
```
merkata, line 41: `container=container` for consistency sake?

synapse_rock/rockcraft.yaml
```diff
@@ -49,6 +49,9 @@ parts:
             - xmlsec1
         stage-snaps:
             - mjolnir/latest/edge
+            - synapse-stats-exporter/latest/edge
+            - matrix-appservice-irc/latest/edge
+            - synapse-stats-exporter/latest/edge
```
merkata, line 54: duplicate with line 52?

merkata (commented):
Good start, left some minor comments, interested to see the tests on this.

src/charm.py
amandahla: My bad, I changed this while trying the idea of passing the token as a parameter and then left as a named parameter. Reverted, thanks!

src/charm.py
amandahla: Agree, changed!

src/charm.py
amandahla: Yes, Mjolnir requires it too.

- If Mjolnir is enabled, the charm will instantiate a Mjolnir class passing state as parameter
- The class observes the collect status event that will check for the requirements and enable Mjolnir when everything is set

src/mjolnir.py
amandahla: Right now if the admin access token is None, the charm will be put on MaintenanceStatus. Do you think it should be changed?

self._charm.unit.status = ops.MaintenanceStatus(
                "Failed to get admin access token. Please, check the logs."
            )

src-docs/charm.py.md
```diff
@@ -18,7 +18,11 @@ Charm for Synapse on kubernetes.
 ## <kbd>class</kbd> `SynapseCharm`
 Charm the service. 
 
+<<<<<<< HEAD
```
nrobinaubertin, line 21: left-over from merge

src/pebble.py
```diff
@@ -73,6 +73,22 @@ def replan_mjolnir(self, container: ops.model.Container) -> None:
...
+        """Replan Synapse StatsExporter service.
+
+        Args:
+            container: Charm container.
+        """
+        access_token = self._charm_state.synapse_config.admin_access_token
+        if not access_token:
+            logger.warning("Admin Access Token is none, skipping Stats Exporter start.")
+        layer = self._stats_exporter_pebble_layer
+        layer["services"][synapse.STATS_EXPORTER_SERVICE_NAME]["environment"] = {
```
nrobinaubertin, line 86: It would make more sens to me to have that in the function that creates the stats_exporter layer. But then it couldn't be a property anymore. Why are we defining those create-layer-functions as properties ?

src-docs/charm.py.md
amandahla: Thanks, for some reason I lost my pre-commit config :| Now is fixed.

src/pebble.py
amandahla: IMHO The main reason for being properties is to dynamically create the environment, for example, every time the attribute is called. I think there is no significant change in performance but makes the code easier to read.

nrobinaubertin (dismissed):

src/pebble.py
nrobinaubertin: I'm not sure that I like the property decorator in those cases but it's a stylistic choice (I'm not sure that I like decorators in general).

javierdelapuente (dismissed):

amandahla:
I'm closing this PR in favor of splitting the changes into smaller ones. This is already too messy.

PR #160, opened 2024-01-26
feat(irc-bridge): Add IRC bridge module

merkata:
### Overview

Adds an IRC bridge module. This is part of a couple of smaller PRs and is not feature complete, note it is a PR against the feature branch.

### Rationale

Adheres to managing charm complexity.

### Juju Events Changes

n/a
### Module Changes

Adds irc bridge module.
### Library Changes

n/a
### Checklist

### DIFF
--- a/src/irc_bridge.py
+++ b/src/irc_bridge.py
@@ -1,0 +1,89 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Provide the IRC bridge class to represent the matrix-appservice-app plugin for Synapse."""
+
+# disabling due the fact that collect status does many checks
+# pylint: disable=too-many-return-statements
+
+import logging
+import typing
+
+import ops
+
+import actions
+import synapse
+from charm_state import CharmState
+
+logger = logging.getLogger(__name__)
+
+IRC_SERVICE_NAME = "irc"
+
+class IRCBridge(ops.Object):  # pylint: disable=too-few-public-methods
+    """A class representing the IRC bridge plugin for Synapse application.
+
+    See https://github.com/matrix-org/matrix-appservice-irc/ for more details about it.
+    """
+
+    def __init__(self, charm: ops.CharmBase, charm_state: CharmState):
+        """Initialize a new instance of the IRC bridge class.
+
+        Args:
+            charm: The charm object that the IRC bridge instance belongs to.
+            charm_state: Instance of CharmState.
+        """
+        super().__init__(charm, "mjolnir")
+        self._charm = charm
+        self._charm_state = charm_state
+        self.framework.observe(charm.on.collect_unit_status, self._on_collect_status)
+
+    @property
+    def _pebble_service(self) -> typing.Any:
+        """Return instance of pebble service.
+
+        Returns:
+            instance of pebble service or none.
+        """
+        return getattr(self._charm, "pebble_service", None)
+
+    def _on_collect_status(self, event: ops.CollectStatusEvent) -> None:
+        """Collect status event handler.
+
+        Args:
+            event: Collect status event.
+        """
+        if not self._charm_state.synapse_config.enable_irc_bridge:
+            return
+        container = self._charm.unit.get_container(synapse.SYNAPSE_CONTAINER_NAME)
+        if not container.can_connect():
+            self._charm.unit.status = ops.MaintenanceStatus("Waiting for Synapse pebble")
+            return
+        irc_service = container.get_services(IRC_SERVICE_NAME)
+        if irc_service:
+            logger.debug("%s service already exists, skipping", IRC_SERVICE_NAME)
+            return
+        self.enable_irc_bridge()
+        event.add_status(ops.ActiveStatus())
+
+    def enable_irc_bridge(self) -> None:
+        """Enable irc service.
+
+        The required steps to enable the IRC bridge are:
+         - Create the IRC bridge configuration file.
+         - Create the IRC bridge registration file.
+         - Finally, add IRC bridge pebble layer.
+
+        """
+        container = self._charm.unit.get_container(synapse.SYNAPSE_CONTAINER_NAME)
+        if not container.can_connect():
+            self._charm.unit.status = ops.MaintenanceStatus("Waiting for Synapse pebble")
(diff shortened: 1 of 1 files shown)


src/irc_bridge.py
```diff
@@ -0,0 +1,87 @@
...
+    @property
+    def _pebble_service(self) -> typing.Any:
+        """Return instance of pebble service.
+
+        Returns:
+            instance of pebble service or none.
+        """
+        return getattr(self._charm, "pebble_service", None)
+
+    def _on_collect_status(self, event: ops.CollectStatusEvent) -> None:
```
amandahla, line 49: Cool, I think this is the first time that I saw the same event being observed twice in a charm. Could happen a conflict between this one and the one in Mjolnir?

amandahla (approved):
LGTM

src/irc_bridge.py
merkata: Oh not sure about it, still no integ test - good catch! @gregory-schiano wdyt?

src/irc_bridge.py
```diff
@@ -0,0 +1,87 @@
...
+        container = self._charm.unit.get_container(synapse.SYNAPSE_CONTAINER_NAME)
+        if not container.can_connect():
+            self._charm.unit.status = ops.MaintenanceStatus("Waiting for Synapse pebble")
+            return
+        self._charm.model.unit.status = ops.MaintenanceStatus("Configuring IRC bridge")
+        server_name = self._charm_state.synapse_config.server_name
+        synapse.create_irc_bridge_app_registration(container=container)
+        synapse.create_irc_bridge_config(container=container, server_name=server_name)
+        self._pebble_service.replan_irc_bridge(container)
+        self._charm.model.unit.status = ops.ActiveStatus()
```
javierdelapuente, line 87: Does it make sense that we also set the unit status to active here (it is set in the collect_status)?

src/irc_bridge.py
```diff
@@ -0,0 +1,87 @@
...
+
+        Args:
+            event: Collect status event.
+        """
+        # pylint:disable=duplicate-code
+        if not self._charm_state.synapse_config.enable_irc_bridge:
+            return
+        container = self._charm.unit.get_container(synapse.SYNAPSE_CONTAINER_NAME)
+        if not container.can_connect():
+            self._charm.unit.status = ops.MaintenanceStatus("Waiting for Synapse pebble")
```
javierdelapuente, line 60: I think it should be event.add_status(ops.MaintenanceStatus("Waiting for Synapse pebble")), otherwise it can be overwritten by the other collect status.

src/irc_bridge.py
```diff
@@ -0,0 +1,87 @@
...
+            return
+        irc_service = container.get_services(IRC_SERVICE_NAME)
+        if irc_service:
+            logger.debug("%s service already exists, skipping", IRC_SERVICE_NAME)
```
javierdelapuente, line 64: Now that we are going to use the collect_status, should be add a ActiveStatus in all the correct paths?

tests/unit/test_irc_bridge.py
```diff
@@ -0,0 +1,238 @@
...
+    harness.update_config({"enable_irc_bridge": True})
+    harness.begin_with_initial_hooks()
+    harness.set_leader(True)
+    container: ops.Container = harness.model.unit.get_container(synapse.SYNAPSE_CONTAINER_NAME)
+    monkeypatch.setattr(container, "get_services", MagicMock(return_value=MagicMock()))
+    enable_irc_bridge_mock = MagicMock()
+    monkeypatch.setattr(IRCBridge, "enable_irc_bridge", enable_irc_bridge_mock)
+
+    event_mock = MagicMock()
+    harness.charm._irc_bridge._on_collect_status(event_mock)
```
javierdelapuente, line 48: There is a `harness.evaluate_status` function, maybe that is more realistic than directly calling the event handler?

src/irc_bridge.py
merkata: confirmed it won't be a conflict - have a look at https://discourse.charmhub.io/t/events-collect-app-status-and-collect-unit-status/12562

src/irc_bridge.py
merkata: I'm not sure tbh - the whole logic in this file is interesting:
Ops will send the collect status at the end of a hook (and we assume synapse should be running after the hook), then we check here whether it is running - if not, we set a global Maintenance. If it is up, to me any other part that checks that synapse is up should be able to clear that Maintenance. Further down we set the status to irc bridge with an `event.add`, after running the enable_irc_bridge (where we again check that synapse is running...). This would need refactoring imo.

src/irc_bridge.py
merkata: Can you elaborate on which are correct paths in your opinion? Have mjolnir and irc bridge depend on synapse and add Maintenance statuses via event.add in case the synapse container is not reachable?

src/irc_bridge.py
merkata: To me setting this here clears the Maintenance on line 82.

javierdelapuente: As this code is only called from here: 
```
        self.enable_irc_bridge()
        event.add_status(ops.ActiveStatus())
```
and if nothing else fails in another collect status it will go to Active, not sure if it is necessary to set active here.

src/irc_bridge.py
javierdelapuente: Not sure neither TBH. We set it here in MaintenanceStatus (but the irc bridge is not configured), and this handler ends here. If any other collect_status handler runs a `add_status(ActiveStatus)`, it will end in active, but the irc bridge is not enabled. What should be the status?

javierdelapuente: Pinging @gregory-schiano as he pointed me to this comment https://github.com/canonical/synapse-operator/pull/30#discussion_r1324745492

src/irc_bridge.py
javierdelapuente: I mean, from the point of view of this `IRCBridge.collect_status`, at this point everything is correct, same as if irc was correctly enabled. But maybe it adds nothing. I close the conversation.

PR #165, opened 2024-01-30
Create backup action

javierdelapuente:
Applicable spec: ISD095

### Overview

This PR creates the action `create-backup` to back up Synapse, using the s3 integration to get
S3 credentials and the `backup_passphrase` config variable to use as the encryption
key. 

command using pipes, like `tar -c <paths> | gpg --symmetric <options> | aws s3 cp <options>`.  For
that command to work, the gpg `passphrase` to use was previously copied to the 
environment variables.

Besides, the aws s3 command has an option to give the expected size of the object to upload (`--expected-size` ). This is required so the aws s3 cp can calculate the correct size for the multipart uploads. This is done in this PR using a bare `du` command with all the paths to back up.

### Rationale

### Juju Events Changes

Added `create-backup` action.

### Module Changes

The entry point for the logic is the `backup_observer`, which register a handler for the `create-backup` action.

All the rest of the logic is inside the `backup` module, that performs several actions in the workload to 

### Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 jsonschema ==4.21.1
-ops ==2.9.0
+ops ==2.10.0
 pydantic ==1.10.14
 psycopg2-binary ==2.9.9
 requests == 2.31.0


javierdelapuente (commented):
test

PR #171, opened 2024-01-31
Create backup action

javierdelapuente:
Applicable spec: ISD095

### Overview

This PR creates the action `create-backup` to back up Synapse, using the s3 integration to get
S3 credentials and the `backup_passphrase` config variable to use as the encryption
key. 

command using pipes, like `tar -c <paths> | gpg --symmetric <options> | aws s3 cp <options>`.  For
that command to work, the gpg `passphrase` to use was previously copied to the 
environment variables.

Besides, the aws s3 command has an option to give the expected size of the object to upload (`--expected-size` ). This is required so the aws s3 cp can calculate the correct size for the multipart uploads. This is done in this PR using a bare `du` command with all the paths to back up.

### Rationale

### Juju Events Changes

Added `create-backup` action.

### Module Changes

The entry point for the logic is the `backup_observer`, which register a handler for the `create-backup` action.

All the rest of the logic is inside the `backup` module, that performs several actions in the workload to 

### Checklist


src/backup.py
```diff
@@ -3,18 +3,47 @@
...
+
+# For the data directory, inside the "media" directory, all directories starting
+# with local_ will be backed up. The directories starting with "remote_" are from
+# other server and is it not necessary to back them up.
+MEDIA_DIR = "media"
+LOCAL_DIR_PATTERN = "local_*"
+
+# A smaller value will minimise memory requirements. A bigger value can make the transfer faster.
+S3_MAX_CONCURRENT_REQUESTS = 1
+PASSPHRASE_FILE = "/root/.gpg_passphrase"  # nosec
```
javierdelapuente, line 35: Not sure this is the right place. Any idea?

arturo-seijas: Do you mean the key location? This would imply running synapse as root user, which contravenes the least privilege principle

arturo-seijas: Perhaps under the application directory (`/srv/synapse/...` I guess) or under `$HOME` (wherever that is)?

javierdelapuente: I have updated the location of the passphrase file to Synapse config dir. I haven't changed to run it as the Synapse user as it is managed by the startup script and it is not fully configured (has no home dir for example).

WDYT?

arturo-seijas (dismissed):

src/backup.py
```diff
@@ -128,3 +157,203 @@ def can_use_bucket(self) -> bool:
...
+        paths: Paths to check.
+
+    Returns:
+        Total size in bytes.
+
+    Raises:
+        BackupError: If there was a problem calculating the size.
+    """
+    command = "set -euxo pipefail; du -bsc " + paths_to_args(paths) + " | tail -n1 | cut -f 1"
+    exec_process = container.exec([BASH_COMMAND, "-c", command])
```
javierdelapuente, line 207: try catch

javierdelapuente:
closed as it will be done in another branch

PR #172, opened 2024-01-31
Refactor S3 Client for Backup

javierdelapuente:
### Overview

As a first step before the main [PR](https://github.com/canonical/synapse-operator/pull/171) for the backup action, the S3 interface is changed for the backup.

The interface for s3 now is just the function `can_use_bucket`. Other functions will be needed, like `list_objects`/`list_backups`, `delete_object`/`delete_backup`, and they will all use the same boto s3 client. A class is used to provide a simpler interface and will share the low level boto3 client. The other methods will be created in next PRs.

The low level client with S3 has been changed to use s3 client interface, as sometimes in the project (in the tests) the client interface was used and other times it was the resource interface.

Other some minor improvements have been made (use of the host/path property in S3 and putting tests in their correct directories)

### Rationale

Split the main backup [PR](https://github.com/canonical/synapse-operator/pull/171) into manageable chunks.

### Checklist

### DIFF
--- a/src/backup.py
+++ b/src/backup.py
@@ -7,11 +7,16 @@ import logging
 from typing import Any, Optional
 
 import boto3
+from botocore.config import Config
 from botocore.exceptions import BotoCoreError, ClientError
 from botocore.exceptions import ConnectionError as BotoConnectionError
 from pydantic import BaseModel, Field, validator
 
 logger = logging.getLogger(__name__)
+
+
+class S3Error(Exception):
+    """Generic S3 Exception."""
 
 
 class S3Parameters(BaseModel):
@@ -24,7 +29,8 @@ class S3Parameters(BaseModel):
         bucket: The bucket name.
         endpoint: The endpoint used to connect to the object storage.
         path: The path inside the bucket to store objects.
-        s3_uri_style: The S3 protocol specific bucket path lookup type.
+        s3_uri_style: The S3 protocol specific bucket path lookup type. Can be "path" or "host".
+        addressing_style: S3 protocol addressing style, can be "path" or "virtual".
     """
 
     access_key: str = Field(alias="access-key")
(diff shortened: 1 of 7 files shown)


gtrkiller (approved):
LGTM

tests/unit/conftest.py
```diff
@@ -67,6 +68,10 @@ def wait_output(self):
                 stderr=self._stderr,
             )
 
+        def wait(self):
```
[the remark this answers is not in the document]
javierdelapuente: oops, not yet, it it in the other PR...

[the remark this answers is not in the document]
arturo-seijas: Can you remove it so that PRs are more atomic?

arturo-seijas (approved):

src/backup.py
```diff
@@ -7,13 +7,18 @@
...
 from pydantic import BaseModel, Field, validator
 
 logger = logging.getLogger(__name__)
 
 
+class S3Error(Exception):
+    """Generic S3 Exception."""
+
+
 class S3Parameters(BaseModel):
```
amandahla, line 22: Not blocker: But I've just realized that this reminds me of a discussion once @gregory-schiano 
Why is S3Parameters a class?

javierdelapuente: That's a good question.

Initially, following the [contributing guide](https://github.com/canonical/is-charms-contributing-guide), a NamedTuple or similar could be worth just to group arguments. S3Parameters will not only be used for the S3Client, but in next PRs it will be also used for the workload (to set environment variables and to use as arguments in exec commands to the container).

However, as in SynapseConfig or ProxyConfig, it is using `pydantic` for some validations/parsing, and that's the reason it is not a NamedTuple. I think splitting that entity into two (one for parsing/validation) and another for internal use in backup is not worth for this simple case, and where S3 is a stable protocol.

amandahla (approved):
LGTM

src/backup.py
amandahla: Hmm...Lets see how it goes, closing since is not a blocker

PR #184, opened 2024-02-06
Create backup action

javierdelapuente:
Applicable spec: ISD095

### Overview

This PR creates the action create-backup to back up Synapse, using the s3 integration to get
S3 credentials and the backup_passphrase config variable to use as the encryption
key.

command using pipes, like tar -c <paths> | gpg --symmetric <options> | aws s3 cp <options>. For
that command to work, the gpg passphrase to use was previously copied to the
environment variables.

Besides, the aws s3 command has an option to give the expected size of the object to upload (--expected-size ). This is required so the aws s3 cp can calculate the correct size for the multipart uploads. This is done in this PR using a bare du command with all the paths to back up.

### Rationale

### Juju Events Changes

Added create-backup action.

### Module Changes

The entry point for the logic is the backup_observer, which register a handler for the create-backup action.

All the rest of the logic is inside the backup module, that performs several actions in the workload to

### Checklist


src/backup.py
```diff
@@ -128,3 +161,255 @@ def can_use_bucket(self) -> bool:
             )
             return False
         return True
+
+
+def create_backup(
+    container: ops.Container,
+    s3_parameters: S3Parameters,
+    passphrase: str,
+    backup_id: Optional[str] = None,
```
arturo-seijas, line 170: I don't see this argument being passed. It's probably worth removing in: YAGNI

src/backup.py
```diff
@@ -128,3 +161,255 @@ def can_use_bucket(self) -> bool:
...
+    Raises:
+       BackupError: If there was an error creating the backup.
+    """
+    if backup_id is None:
+        backup_id = "backup-" + datetime.datetime.now().strftime(BACKUP_ID_FORMAT)
+
+    _prepare_container(container, s3_parameters, passphrase)
+    paths_to_backup = _get_paths_to_backup(container)
+    logger.info("Paths to back up: %s.", list(paths_to_backup))
+    if not paths_to_backup:
```
arturo-seijas, line 189: Is it unreasonable to think there might be no files to backup? Perhaps it would be better to log a warning instead of raising an exception here or even silently return. Is there any issue if the backup is executed passing an empty list of paths?

src/backup.py
```diff
@@ -128,3 +161,255 @@ def can_use_bucket(self) -> bool:
...
+    logger.info("Backup command output: %s. %s.", stdout, stderr)
+    return backup_id
+
+
+def _prepare_container(
+    container: ops.Container, s3_parameters: S3Parameters, passphrase: str
+) -> None:
+    """Prepare container for create or restore backup.
+
+    This means preparing the required aws configuration and the gpg passphrase file.
```
arturo-seijas, line 219: s/preparing the required aws configuration/configuring the aws client

src/backup.py
javierdelapuente: I understand that this should never happen. As Synapse should have started, at least the keys should be there... it should be an error

src/backup.py
```diff
@@ -128,3 +161,255 @@ def can_use_bucket(self) -> bool:
...
+        stdout, stderr = exec_process.wait_output()
+    except (APIError, ExecError) as exc:
+        raise BackupError("Cannot calculate size of paths. du failed.") from exc
+
+    logger.info(
+        "Calculating size of paths. Command: %s. stdout: %s. stderr: %s", command, stdout, stderr
+    )
+
+    try:
+        return int(stdout)
```
arturo-seijas, line 320: Is this even possible considering we should have received a 0 exit code?

javierdelapuente: I understand this should not happen...

synapse_rock/rockcraft.yaml
```diff
@@ -13,6 +13,13 @@ license: Apache-2.0
 platforms:
   amd64:
 parts:
+    add-user:
```
arturo-seijas, line 16: Rockcraft now injects an user to run services. You can see an example at https://github.com/canonical/discourse-k8s-operator/blob/main/discourse_rock/rockcraft.yaml#L9

tests/integration/test_charm.py
```diff
@@ -583,3 +584,70 @@ async def test_synapse_enable_s3_backup_integration_no_bucket(
...
+
+    synapse_unit: Unit = next(iter(synapse_app.units))
+    backup_action: Action = await synapse_unit.run_action("create-backup")
+    await backup_action.wait()
+
+    assert backup_action.status == "completed"
+    assert "backup-id" in backup_action.results
+    bucket_name = s3_backup_configuration["bucket"]
+    path = s3_backup_configuration["path"].strip("/")
+    object_key = path + "/" + backup_action.results["backup-id"]
```
arturo-seijas, line 621: How about an f-string here?

synapse_rock/rockcraft.yaml
javierdelapuente: I do not think we can use that, as we need the user to have uid 991 and gid 991 and also to have a home (for gpg to work).

src/backup.py
```diff
@@ -128,3 +161,255 @@ def can_use_bucket(self) -> bool:
...
+
+    Returns:
+       a S3 path, only including the protocol if the bucket is included.
+    """
+    # Removing the root slash in path makes this code compatible with
+    # MinIO. MinIO looks like it does not have a slash as the first element,
+    # and if the file was uploaded with one, it gets removed.
+    path = prefix.strip("/")
+    if object_name:
+        path = str(pathlib.Path(f"{path}/{object_name}"))
```
arturo-seijas, line 396: Why instantiating a path to convert it again to a string?

javierdelapuente: I was thinking in only having valid paths. That is it, cpnverting "path//subpath" to "path/subpath". It is because of this: https://github.com/minio/minio/issues/5958

arturo-seijas: Oh! Makes sense. Would you mind adding a comment explaining the reasoning?

arturo-seijas (dismissed):

src/backup.py
```diff
@@ -3,18 +3,51 @@
...
+AWS_COMMAND = "/aws/dist/aws"
+
+# The configuration files to back up consist in the signing keys
+# plus the sqlite db if it exists.
+BACKUP_FILE_PATTERNS = ["*.key", "homeserver.db*"]
+
+# For the data directory, inside the "media" directory, all directories starting
+# with local_ will be backed up. The directories starting with "remote_" are from
+# other server and is it not necessary to back them up.
+MEDIA_DIR = "media"
```
nrobinaubertin, line 31: This will be dynamic in a future PR and should be read from the homeserver.yaml file

src/backup.py
arturo-seijas: I think it would make sense to directly let the charm error if this happens

javierdelapuente: Removed the try catch that should not happen!

src/backup.py
```diff
@@ -128,3 +161,253 @@ def can_use_bucket(self) -> bool:
...
+    except ValueError as exc:
+        raise BackupError("Cannot calculate size of paths. Wrong stdout.") from exc
+
+
+def _build_backup_command(
+    s3_parameters: S3Parameters,
+    backup_id: str,
+    backup_paths: Iterable[str],
+    passphrase_file: str,
+    expected_size: int = int(1e10),
```
nrobinaubertin, line 327: Why is there a default value for expected_size ?

arturo-seijas (approved):

src/backup.py
```diff
@@ -128,3 +161,253 @@ def can_use_bucket(self) -> bool:
...
+    paths = _paths_to_args(backup_paths)
+    tar_command = f"tar -c {paths}"
+    gpg_command = f"gpg --batch --no-symkey-cache --passphrase-file '{passphrase_file}'"
+    gpg_command += " --symmetric"
+    aws_command = f"{AWS_COMMAND} s3 cp --expected-size={expected_size} - "
+
+    s3_url = _s3_path(
+        prefix=s3_parameters.path, object_name=backup_id, bucket=s3_parameters.bucket
+    )
+    aws_command += f"'{s3_url}'"
```
nrobinaubertin, line 352: It's minor but consider using the pattern:
```
a = (f"start of the string"
     f"next part of the string"
     )
```
For the long strings. This will also force you to regroup them (`aws_command` is in two places right now)

src/synapse/workload.py
```diff
@@ -31,10 +31,12 @@
 SYNAPSE_CONFIG_PATH = f"{SYNAPSE_CONFIG_DIR}/homeserver.yaml"
 SYNAPSE_CONTAINER_NAME = "synapse"
 SYNAPSE_DATA_DIR = "/data"
+SYNAPSE_GROUP = "synapse"
 SYNAPSE_NGINX_CONTAINER_NAME = "synapse-nginx"
 SYNAPSE_NGINX_PORT = 8080
 SYNAPSE_NGINX_SERVICE_NAME = "synapse-nginx"
 SYNAPSE_SERVICE_NAME = "synapse"
+SYNAPSE_USER = "synapse"
```
nrobinaubertin, line 39: Will this be only used for backups ? Then it should be called `synapse_backup` (or just `backups` ?) in my opinion

src/backup.py
javierdelapuente: removed, it is not used

src/synapse/workload.py
javierdelapuente: It is the same user as used by the Synapse server. I think it can confuse if I name it as `synapse_backup` when it is the same one (the 991 uid).

nrobinaubertin: It make sense if it's the same

PR #193, opened 2024-02-09
Feat/add irc bridge

merkata:
### Overview

Adds an IRC bridge app end to end - the app is packed in the rock, there is a new req for a dedicated db, tests are updated.

### Rationale

Requested by Community for IRC integration.

### Juju Events Changes

IRC bridge acts on `collect_status` much like mjolnir (being neutral about this approach though it needs rethinking).

### Module Changes

New module for irc bridge, adjustments to workload and pebble, db as well so the class can be used by the new db as well.

### Library Changes

n/a

### Checklist


src/database_observer.py
```diff
@@ -30,18 +30,19 @@ class DatabaseObserver(Object):
 
     _RELATION_NAME = "database"
 
-    def __init__(self, charm: CharmBase):
+    def __init__(self, charm: CharmBase, relation_name: str = _RELATION_NAME):
```
nrobinaubertin, line 33: I propose to remove the default (we don't want it to silently fail in a subtle way)

src/synapse/workload.py
```diff
@@ -545,6 +571,117 @@ def create_mjolnir_config(container: ops.Container, access_token: str, room_id:
...
+            "-r",
+            "-f",
+            IRC_BRIDGE_REGISTRATION_PATH,
+            "-u",
+            "http://localhost:",
+            str(SYNAPSE_PORT),
+            "-c",
+            IRC_BRIDGE_CONFIG_PATH,
+            "-l",
+            "my_bot",
```
nrobinaubertin, line 637: s/my/irc ?

templates/irc_bridge_production.yaml
```diff
@@ -0,0 +1,141 @@
...
+    minUserActiveDays: 1
+    inactiveAfterDays: 30
+  ircHandler:
+    mapIrcMentionsToMatrix: "on" # This can be "on", "off", "force-off".
+    powerLevelGracePeriod: 1000
+  permissions:
+    '@merkata:ubuntu.com': admin
+    '@amandahla:ubuntu.com': admin
+    '@kewisch:ubuntu.com': admin
+    '@ilvipero:ubuntu.com': admin
```
nrobinaubertin, line 130: If we're making a charm for everyone to use, this shouldn't be hardcoded

tox.ini
```diff
@@ -117,7 +117,7 @@ commands =
 description = Run integration tests
 deps =
     macaroonbakery==1.3.2 # fixes "TypeError: Descriptors cannot be created directly." protobuf error
-    juju >=3.0
+    juju >=3.0,<=3.3.0
```
nrobinaubertin, line 120: This workaround is not needed anymore

pyproject.toml
```diff
@@ -11,7 +11,7 @@ skips = ["*/*test.py", "*/test_*.py", "*tests/*.py"]
 branch = true
 
 [tool.coverage.report]
-fail_under = 92
+fail_under = 90
```
jdkandersson, line 14: A reduction? I think our standards say that we shouldn't go down, do you think that is applicable here?

src-docs/irc_bridge.py.md
```diff
@@ -0,0 +1,89 @@
...
+A class representing the IRC bridge plugin for Synapse application. 
+
+See https://github.com/matrix-org/matrix-appservice-irc/ for more details about it. 
+
+<a href="../src/irc_bridge.py#L41"><img align="right" style="float:right;" src="https://img.shields.io/badge/-source-cccccc?style=flat-square"></a>
+
+### <kbd>function</kbd> `__init__`
+
+```python
+__init__(charm: CharmBase, charm_state: CharmState)
```
jdkandersson, line 25: Is this an observer? If not, what is the reason for needing the `charm` object please?

jdkandersson: I think it is intended to be an observer, perhaps include that in the name of the class?

src/charm_state.py
```diff
@@ -58,6 +58,7 @@ class SynapseConfig(BaseModel):  # pylint: disable=too-few-public-methods
 
     Attributes:
         allow_public_rooms_over_federation: allow_public_rooms_over_federation config.
+        enable_irc_bridge: enable_irc_bridge config.
```
jdkandersson, line 61: Perhaps describe what it is intended to enable? I'm not sure the description provides any more information?

src/charm_state.py
```diff
@@ -161,11 +165,13 @@ def proxy(self) -> "ProxyConfig":
             no_proxy=no_proxy,
         )
 
+    # pylint: disable=too-many-arguments
```
jdkandersson, line 168: Perhaps consider grouping some of these into sub classes? Or otherwise, should we include an explanatory comment as to why the disable is necessary please?

src/irc_bridge.py
```diff
@@ -0,0 +1,154 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Provide the IRC bridge class to represent the matrix-appservice-app plugin for Synapse."""
+
+# disabling due the fact that collect status does many checks
+# pylint: disable=too-many-return-statements
```
jdkandersson, line 7: Perhaps group the checks into separate functions?

src/irc_bridge.py
```diff
@@ -0,0 +1,154 @@
...
+            charm: The charm object that the IRC bridge instance belongs to.
+            charm_state: Instance of CharmState.
+        """
+        super().__init__(charm, "irc-bridge")
+        self._charm = charm
+        self._charm_state = charm_state
+        self.framework.observe(charm.on.collect_unit_status, self._on_collect_status)
+
+    @property
+    def _pebble_service(self) -> typing.Any:
```
jdkandersson, line 54: Is it possible to define a type here that is more specific than `Any`?

src/irc_bridge.py
```diff
@@ -0,0 +1,154 @@
...
+        """
+        return getattr(self._charm, "pebble_service", None)
+
+    def _on_collect_status(self, event: ops.CollectStatusEvent) -> None:
+        """Collect status event handler.
+
+        Args:
+            event: Collect status event.
+        """
+        # pylint:disable=duplicate-code
```
jdkandersson, line 68: What is the duplicate code here please? Perhaps consider combining it or describing why it isn't possible?

src/irc_bridge.py
```diff
@@ -0,0 +1,154 @@
...
+            return
+        container = self._charm.unit.get_container(synapse.SYNAPSE_CONTAINER_NAME)
+        if not container.can_connect():
+            self._charm.unit.status = ops.MaintenanceStatus("Waiting for Synapse pebble")
+            return
+        irc_service = container.get_services(IRC_SERVICE_NAME)
+        if irc_service:
+            logger.debug("%s service already exists, skipping", IRC_SERVICE_NAME)
+            return
+        self.enable_irc_bridge()
```
jdkandersson, line 82: Is this method only intended to be called from here? If so, perhaps it could be private?

src/irc_bridge.py
```diff
@@ -0,0 +1,154 @@
...
+        Returns:
+            The database connection string.
+        """
+        db_connect_string = (
+            "postgres://"
+            + f"{self._charm_state.irc_bridge_datasource['user']}"  # type: ignore
+            + f":{self._charm_state.irc_bridge_datasource['password']}"  # type: ignore
+            + f"@{self._charm_state.irc_bridge_datasource['host']}"  # type: ignore
+            + f":{self._charm_state.irc_bridge_datasource['port']}"  # type: ignore
+            + f"/{self._charm_state.irc_bridge_datasource['db']}"  # type: ignore
```
jdkandersson, line 152: What are the `type: ignore` for here please?

src/pebble.py
```diff
@@ -257,6 +266,35 @@ def _mjolnir_pebble_layer(self) -> ops.pebble.LayerDict:
...
+    def _irc_bridge_pebble_layer(self) -> ops.pebble.LayerDict:
+        """Generate pebble config for the irc bridge service.
+
+        Returns:
+            The pebble configuration for the irc bridge service.
+        """
+        command_params = (
+            f"-c {synapse.IRC_BRIDGE_CONFIG_PATH}"
+            + f" -f {synapse.IRC_BRIDGE_REGISTRATION_PATH}"
+            + f" -p {synapse.IRC_BRIDGE_HEALTH_PORT}"
```
jdkandersson, line 279: Is the `+` required here? I think it should work without it and it will get auto-merged by the compilation step

src/synapse/workload.py
jdkandersson, line 1: Do you think it would be worth it to create an `irc` submodule instead of putting it into the workload module?

src/synapse/workload.py
```diff
@@ -149,6 +162,19 @@ def check_mjolnir_ready() -> ops.pebble.CheckDict:
...
+    """Return the Synapse IRC bridge service check.
+
+    Returns:
+        Dict: check object converted to its dict representation.
+    """
+    check = Check(CHECK_IRC_BRIDGE_READY_NAME)
+    check.override = "replace"
+    check.level = "ready"
+    check.http = {"url": f"http://localhost:{IRC_BRIDGE_HEALTH_PORT}"}
+    return check.to_dict()
```
jdkandersson, line 175: This module probably shouldn't be `ops` aware - perhaps better to put this function into the layers that are `ops` / `juju` aware?

src/synapse/workload.py
```diff
@@ -545,6 +571,117 @@ def create_mjolnir_config(container: ops.Container, access_token: str, room_id:
...
+    """Create config as expected by irc bridge.
+
+    Args:
+        server_name: server name of the Synapse instance.
+        db_connect_string: database connection string.
+
+    Returns:
+        IRC Bridge configuration
+    """
+    with open("templates/irc_bridge_production.yaml", encoding="utf-8") as irc_config_file:
```
jdkandersson, line 584: What about pathlib.read_text https://docs.python.org/3/library/pathlib.html#pathlib.Path.read_text?

src/synapse/workload.py
```diff
@@ -545,6 +571,117 @@ def create_mjolnir_config(container: ops.Container, access_token: str, room_id:
...
+        container: Container of the charm.
+
+    Raises:
+        CreateIRCBridgeRegistrationError: error creating irc bridge app registration.
+    """
+    try:
+        _get_irc_bridge_app_registration(container=container)
+        _add_app_service_config_field(container=container)
+    except ops.pebble.PathError as exc:
+        raise CreateIRCBridgeRegistrationError(str(exc)) from exc
```
jdkandersson, line 662: It is kind of confusing that calling a non-ops function might raise a `ops` error. I'm guessing these bubble up, perhaps handle them directly in the function?

templates/irc_bridge_production.yaml
```diff
@@ -0,0 +1,141 @@
...
+    enabled: false
+advanced:
+  maxHttpSockets: 1000
+  maxTxnSize: 10000000
+sentry:
+  enabled: false
+  dsn: "https://<key>@sentry.io/<project>"
+database:
+  engine: "postgres"
+  connectionString: ""
```
jdkandersson, line 141: Final new line?

jdkandersson (commented):
Haven't looked at the tests yet, thanks for the progress so far!

src/charm.py
```diff
@@ -48,13 +49,15 @@ def __init__(self, *args: typing.Any) -> None:
         """
         super().__init__(*args)
         self._backup = BackupObserver(self)
+        self._irc_bridge_database = DatabaseObserver(self, relation_name="irc-bridge-database")
```
amandahla, line 52: Out of curiosity: Is there a reason for this not to be set only if enable_irc_bridge is true? Together with self._irc_bridge.

src/irc_bridge.py
```diff
@@ -0,0 +1,154 @@
...
+            container: The container to create the PEM file in.
+
+        Raises:
+            PEMCreateError: If the PEM file creation fails.
+        """
+        self._charm.model.unit.status = ops.MaintenanceStatus("Creating PEM file for IRC bridge")
+        pem_create_command = [
+            "/bin/bash",
+            "-c",
+            "[[ -f /data/config/passkey.pem ]] || "
```
amandahla, line 124: Minor: naming the file with something related to IRC bridge? 
And BTW What happens if this file is lost?

src/charm_state.py
```diff
@@ -161,11 +165,13 @@ def proxy(self) -> "ProxyConfig":
...
     @classmethod
     def from_charm(
         cls,
         charm: ops.CharmBase,
         datasource: typing.Optional[DatasourcePostgreSQL],
```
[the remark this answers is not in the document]
merkata: Yes, it can be optional and is covered in the tests as well (datasource=None).

src/irc_bridge.py
merkata: Renamed, thanks!
If the file is lost, a new one can be generated, this is used for storing the irc password for the server, the pass is supplied as a matrix command in the admin room. Still this could be covered by our backup and restore, though I feel this is a minor thing to address now.

src/charm.py
merkata: Nope, it should be within the condition, thanks!

src/synapse/workload.py
merkata: Nice, refactored!

templates/irc_bridge_production.yaml
merkata: done

src-docs/irc_bridge.py.md
merkata: Correct, done.

pyproject.toml
merkata: Reverted

tox.ini
merkata: Reverted

templates/irc_bridge_production.yaml
merkata: I agree

src/synapse/workload.py
merkata: True, done.

src/database_observer.py
merkata: Do you have an example of how it could fail, we might want to put that as a test? The default is how we normally use that, all our charms have that constant of `database`, for synapse it also has custom logic to prepare the database.

src/synapse/workload.py
merkata: Do you mind having a look at the whole file and perhaps make a proposal on where it is better to place ops aware functions? Reason is the module has lots of ops logic and if/when we refactor, I prefer this is done in one pass if possible. Thanks!

src/pebble.py
merkata: True, omitted.

merkata: TIL also, thanks!

src/irc_bridge.py
merkata: Done, made private.

src/irc_bridge.py
merkata: It is a check that synapse is ready, we use it for both mjolnir and now irc, though it is a three line can_connect code that I don't feel putting in a helper would add any benefit, so I went the route of commenting, let me know what you think.

src/irc_bridge.py
merkata: Yup, good call, turned it into a PebbleService

src/irc_bridge.py
merkata: Wouldn't that add overhead particular to this setup? The `return` conditions add a logger or a MaintenanceStatus before returning, otherwise I would group them here as well and return once.

src/charm_state.py
merkata: Added a comment - this is due to the second datasource and given it might be temporary and we move this in a separate charm I'm not tempted to group these and refactor and rewrite the tests associated.

src/charm_state.py
merkata: Done.

src/database_observer.py
nrobinaubertin: I was thinking of the possibility that a dev forgets to specify the relation_name value while initializing a new DatabaseObserver. That could cause havoc without being immediately noticeable.  
By removing the default, we make sure that such errors cannot happen.

src/irc_bridge.py
jdkandersson: There is an interesting article regarding [can connect](https://discourse.charmhub.io/t/handling-temporary-errors-in-charms/13247) perhaps take a look and see if that changes things

src/irc_bridge.py
jdkandersson: Let's leave this for now since I think this function will need to be reworked anyway

src/synapse/workload.py
jdkandersson: You are right there is a lot of `ops` usage in here - that should all be moved out and these kind of modules shouldn't depend on `ops` and instead provide more generic functions and classes to support the charm

src/admin_access_token.py
```diff
@@ -22,7 +22,7 @@
 SECRET_KEY = "secret-key"  # nosec
 
 
-class AdminAccessTokenService:
+class AdminAccessTokenService:  # pragma: no cover
```
jdkandersson, line 25: Perhaps include a statement here as to why this doesn't need to be covered for unit tests please?

src/irc_bridge.py
```diff
@@ -51,7 +52,7 @@ def __init__(self, charm: ops.CharmBase, charm_state: CharmState):
         self.framework.observe(charm.on.collect_unit_status, self._on_collect_status)
 
     @property
-    def _pebble_service(self) -> typing.Any:
+    def _pebble_service(self) -> typing.Optional[PebbleService]:
```
jdkandersson, line 55: What would be the conditions where pebble service is `None` please? Making anything possibly `None` makes it's use more complicated because the user needs to check for whether it is `None` so if it can be avoided that it could be `None` that simplifies code a lot

src/irc_bridge.py
```diff
@@ -0,0 +1,167 @@
...
+        """Collect status event handler.
+
+        Args:
+            event: Collect status event.
+            charm_state: Instance of CharmState.
+        """
+        if not charm_state.synapse_config.enable_irc_bridge:
+            return
+        if charm_state.irc_bridge_datasource is None:
+            ops.MaintenanceStatus("Waiting for irc bridge db relation")
```
yanksyoon, line 77: i think you meant to put `self.unit.status` = :D

src/irc_bridge.py
```diff
@@ -0,0 +1,167 @@
...
+            "-c",
+            "[[ -f /data/config/irc_passkey.pem ]] || "
+            + "openssl genpkey -out /data/config/irc_passkey.pem "
+            + "-outform PEM -algorithm RSA -pkeyopt rsa_keygen_bits:2048",
+        ]
+        logger.info("Creating PEM file for IRC bridge.")
+        try:
+            exec_process = container.exec(
+                pem_create_command,
+                environment={},
```
yanksyoon, line 140: I'm wondering if `environment={}` intends to do something?

src/irc_bridge.py
```diff
@@ -0,0 +1,167 @@
...
+        """
+        if charm_state.irc_bridge_datasource is None:
+            return ""
+        db_connect_string = (
+            "postgres://"
+            + f"{charm_state.irc_bridge_datasource['user']}"
+            + f":{charm_state.irc_bridge_datasource['password']}"
+            + f"@{charm_state.irc_bridge_datasource['host']}"
+            + f":{charm_state.irc_bridge_datasource['port']}"
+            + f"/{charm_state.irc_bridge_datasource['db']}"
```
yanksyoon, line 165: I think we don't quite need the `+` operators here :D

src/synapse/workload.py
```diff
@@ -545,6 +572,117 @@ def create_mjolnir_config(container: ops.Container, access_token: str, room_id:
...
+        container,
+        [
+            "/bin/node",
+            "/app/app.js",
+            "-r",
+            "-f",
+            IRC_BRIDGE_REGISTRATION_PATH,
+            "-u",
+            "http://localhost:",
+            str(SYNAPSE_PORT),
```
yanksyoon, line 634: Perhaps this can be grouped into a single `f"http://localhost:{SYNAPSE_PORT}"`

src/synapse/workload.py
```diff
@@ -545,6 +572,117 @@ def create_mjolnir_config(container: ops.Container, access_token: str, room_id:
...
+        container: Container of the charm.
+
+    Raises:
+        WorkloadError: something went wrong updating the configuration.
+    """
+    try:
+        config = container.pull(SYNAPSE_CONFIG_PATH).read()
+        current_yaml = yaml.safe_load(config)
+
+        current_yaml["app_service_config_files"] = [f"{IRC_BRIDGE_REGISTRATION_PATH}"]
```
yanksyoon, line 679: I think `IRC_BRIDGE_REGISTRATION_PATH` is already a string :D

src/database_observer.py
merkata: The default db is part of the "default" behavior, meaning that if you want to initialize a DatabaseObserver, normally you do it for Synapse and it is the main db, the relation is "database" by convention and that name is set across several charms. The metadata file guards you against relating twice (and I wrote a test to cover that, so if the limit of 1 is raised, it will fail), so it won't fail silently, it will fail loud as it is supposed to and we keep some user ergonomics to point at a DatabaseObserver and it does the job without specifying a name. I'm still torn though as I get your point. @gregory-schiano @jdkandersson @weiiwang01 ?

src/irc_bridge.py
merkata: I don't recall placing it, let's say it is there to see if people review PRs thoroughly - you get the bonus points :tada: and thanks for spotting.

src/irc_bridge.py
merkata: @yanksyoon you have the eye of the :tiger:

src/irc_bridge.py
merkata: Has been refactored away.

src/admin_access_token.py
merkata: Done, I noticed it isn't tested by accident and I annotated it with an additional TODO comment. Thanks!

src/irc_bridge.py
```diff
@@ -0,0 +1,165 @@
...
+            container: The container to create the PEM file in.
+
+        Raises:
+            PEMCreateError: If the PEM file creation fails.
+        """
+        self._charm.model.unit.status = ops.MaintenanceStatus("Creating PEM file for IRC bridge")
+        pem_create_command = [
+            "/bin/bash",
+            "-c",
+            "[[ -f /data/config/irc_passkey.pem ]] || "
```
amandahla, line 131: Heads up: I believe the /data/config is created only if Mjolnir is enabled.
https://github.com/canonical/synapse-operator/blob/main/src/synapse/workload.py#L543

amandahla (approved):
LGTM

src/database_observer.py
jdkandersson: I don't have a strong opinion, would probably lean towards not having the default for safety

src/synapse/workload.py
merkata: I did a first pass and changed a lot of synapse config related to not depend on ops, please have a look. I propose to continue in another refactor PR, this will take time.

src/synapse/workload.py
merkata: Yes, this can end up inside src/irc_bridge, though I think it should be part of a bigger refactor on src/synapse and seeing what belongs where - you raised some good points!

src/charm.py
```diff
@@ -54,21 +54,11 @@ def __init__(self, *args: typing.Any) -> None:
...
-            self._charm_state = CharmState.from_charm(
-                charm=self,
-                datasource=self._database.get_relation_as_datasource(),
-                irc_bridge_datasource=self._irc_bridge_database.get_relation_as_datasource(),
-                saml_config=self._saml.get_relation_as_saml_conf(),
-                smtp_config=self._smtp.get_relation_as_smtp_conf(),
-            )
-        except CharmConfigInvalidError as exc:
-            self.model.unit.status = ops.BlockedStatus(exc.msg)
-            return
```
jdkandersson: I'm confused - if this is removed, where is `_charm_state` set instead?

src/irc_bridge.py
```diff
@@ -32,44 +32,49 @@ def __init__(self, message: str):
...
 
-class IRCBridge(ops.Object):  # pylint: disable=too-few-public-methods
+class IRCBridgeObserver(ops.Object):  # pylint: disable=too-few-public-methods
     """A class representing the IRC bridge plugin for Synapse application.
 
     See https://github.com/matrix-org/matrix-appservice-irc/ for more details about it.
     """
 
-    def __init__(self, charm: ops.CharmBase, charm_state: CharmState):
+    def __init__(self, charm: CharmBaseWithState) -> None:
```
jdkandersson, line 41: I think the guidance is that charm state should be passed as an argument? And that we should also defer instantiation until the event hook going forward so it might not be needed at construct time anymore?

src/irc_bridge.py
```diff
@@ -32,44 +32,49 @@ def __init__(self, message: str):
...
 
         Returns:
-            instance of pebble service or none.
+            The charm object.
         """
-        return getattr(self._charm, "pebble_service", None)
+        return self._charm
 
-    def _on_collect_status(self, event: ops.CollectStatusEvent) -> None:
+    @inject_charm_state
```
jdkandersson, line 62: Oh ok I think I remember this, I think the guidance is changed to not use a decorator and instead construct it in the method? @gregory-schiano perhaps a PR to take a look at

gregory-schiano: Well the implementation on Synapse crossed the finalisation of our decision. For coherence purpose I'd keep it this way, and we'll tackle that in a separate track when we'll ensure the new guidance is implemented on all our charms

amandahla (approved):

gtrkiller (approved):
LGTM, thanks!

src/irc_bridge.py
merkata: answered below

src/charm.py
merkata: answered below

src/irc_bridge.py
merkata: That article is great, thanks for sharing! I'm keeping a note on this for a refactoring PR.

PR #195, opened 2024-02-09
Make admin_access_token available via new class Admin Access Token Service

amandahla:
### Overview

This refactor is needed for other layers that might require an admin access token (like Synapse Stats Exporter).
Also, this makes the charm in accord with the ISD054.

### Rationale

Currently, the admin access token is only used by Mjolnir in a way that is not possible to get it from the state. This change will make getting an admin access token (and its management) easier by adding a new approach: creates a Admin Access Token Service class that will be instantiated in the charm and then passed as parameter to Mjolnir.

### Juju Events Changes

N/A

### Module Changes

synapse has a new file admin
new class: Admin Access Token Service

### Library Changes

N/A

### Checklist

Changing the documentation for charmhub is not needed since this is an internal change.


src/charm.py
```diff
@@ -173,13 +174,31 @@ def _on_config_changed(self, _: ops.HookEvent) -> None:
...
+        if not container.can_connect():
+            self.unit.status = ops.MaintenanceStatus("Waiting for Synapse pebble")
+            return
+        time.sleep(3)
+        admin_user = synapse.create_admin_user(container)
+        if not admin_user:
+            logger.error("Failed to create admin user.")
+            return
+        self._set_admin_access_token(admin_user.access_token)
+        self._charm_state.synapse_config.admin_access_token = admin_user.access_token
```
javierdelapuente, line 188: should not this be set in more places?

src/charm.py
```diff
@@ -173,13 +174,31 @@ def _on_config_changed(self, _: ops.HookEvent) -> None:
...
+        # Mjolnir is a moderation tool for Matrix.
+        # See https://github.com/matrix-org/mjolnir/ for more details about it.
+        if self._charm_state.synapse_config.enable_mjolnir:
+            self._mjolnir = Mjolnir(self, charm_state=self._charm_state)
```
javierdelapuente, line 192: I believe at this point there could be two Mjolnir instances, each one of them registered to the collect_unit_status event (one created in `Charm.__init__` and one here).

src/charm.py
```diff
@@ -173,13 +174,31 @@ def _on_config_changed(self, _: ops.HookEvent) -> None:
     def _on_synapse_pebble_ready(self, _: ops.HookEvent) -> None:
         """Handle synapse pebble ready event."""
         self.change_config()
+        # Since Synapse is ready, the charm can create the admin access token
+        container = self.unit.get_container(synapse.SYNAPSE_CONTAINER_NAME)
+        if not container.can_connect():
+            self.unit.status = ops.MaintenanceStatus("Waiting for Synapse pebble")
+            return
+        time.sleep(3)
```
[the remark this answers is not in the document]
amandahla: Yesterday, I had a completely different idea to remove the dependency between user creation and events. I'll mark this and the following comments as resolved because it has changed a lot by adding a "Token Service". Reviews are welcome! :)

src/charm.py
```diff
@@ -173,13 +174,31 @@ def _on_config_changed(self, _: ops.HookEvent) -> None:
...
+        admin_user = synapse.create_admin_user(container)
```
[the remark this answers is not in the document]
amandahla: Yesterday, I had a completely different idea to remove the dependency between user creation and events. I'll mark this and the following comments as resolved because it has changed a lot by adding a "Token Service". Reviews are welcome! :)

src/charm.py
amandahla: Removed ;)
Yesterday, I had a completely different idea to remove the dependency between user creation and events. I'll mark this and the following comments as resolved because it has changed a lot by adding a "Token Service". Reviews are welcome! :)

src/charm.py
amandahla: Removed ;)

pyproject.toml
```diff
@@ -11,7 +11,7 @@ skips = ["*/*test.py", "*/test_*.py", "*tests/*.py"]
 branch = true
 
 [tool.coverage.report]
-fail_under = 95
+fail_under = 92
```
gtrkiller, line 14: Why it is necessary to diminish the coverage percentage? Wouldn't be better to just add more tests?

src/synapse/admin.py
```diff
@@ -0,0 +1,74 @@
...
+
+from user import User
+
+from .api import register_user
+from .workload import get_registration_shared_secret
+
+logger = logging.getLogger(__name__)
+
+
+def create_admin_user(container: ops.Container) -> typing.Optional[User]:
```
javierdelapuente, line 22: Nice to have this outside of the charm.py file!

pyproject.toml
javierdelapuente, line 14: I think this is because of admin_access_token.py file. If there is a reason not to test it, could it be better to include a #pragma: no cover there with possibly the reason? (https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#test-coverage)

javierdelapuente (commented):
LGTM, nice moving code outside of the main charm, a couple of comments. 

Not for this PR, but it would be nice to have a check for the token validity (sometimes there are 401 in the integ tests in some cases).

nrobinaubertin (approved):

pyproject.toml
nrobinaubertin: It is not always better to have higher coverage. More tests means that refactoring or any change will be harder

nrobinaubertin: I don't think that there is a lot of added value to test something that is just a few logic around using juju internals

amandahla: From the guide, I think we are still covered:
https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#test-coverage

But more tests can be added later.

amandahla: From the guide, I think we are still covered:
https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#test-coverage

But more tests can be added later.

javierdelapuente: Sure, feel free to close this comment, just that the other option, putting "# pragma: no cover" in the files/places you think are not necessary to test, will not change the required test coverage for future PRs.

javierdelapuente (approved):

src/synapse/admin.py
amandahla: Thanks!

PR #196, opened 2024-02-09
Delete backup action

javierdelapuente:
### Checklist

### DIFF
--- a/src/backup.py
+++ b/src/backup.py
@@ -176,6 +176,26 @@ class S3Client:
             return False
         return True
 
+    def delete_backup(self, backup_id: str) -> None:
+        """Delete a backup stored in S3 in the current s3 configuration.
+
+        Args:
+            backup_id: backup id to delete.
+
+        Raises:
+            S3Error: If there was an error deleting the backup.
+        """
+        backups = self.list_backups()
+        if not any(backup.backup_id == backup_id for backup in backups):
+            raise S3Error(f"Backup with id {backup_id} does not exist")
+
+        object_key = _s3_path(prefix=self._s3_parameters.path, object_name=backup_id)
+
+        try:
+            self._client.delete_object(Bucket=self._s3_parameters.bucket, Key=object_key)
+        except ClientError as exc:
+            raise S3Error(f"Cannot delete backup_id {backup_id} from bucket") from exc
+
     def list_backups(self) -> list[S3Backup]:
         """List the backups stored in S3 in the current s3 configuration.
 
--- a/src/backup_observer.py
+++ b/src/backup_observer.py
@@ -44,6 +44,7 @@ class BackupObserver(Object):
         self.framework.observe(
             self._charm.on.restore_backup_action, self._on_restore_backup_action
         )
+        self.framework.observe(self._charm.on.delete_backup_action, self._on_delete_backup_action)
 
     def _on_s3_credential_changed(self, _: CredentialsChangedEvent) -> None:
         """Check new S3 credentials set the unit to blocked if they are wrong."""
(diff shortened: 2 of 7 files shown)


src/backup.py
```diff
@@ -176,6 +176,45 @@ def can_use_bucket(self) -> bool:
...
+        Raises:
+            S3Error: If there was an error checking the backup.
+        """
+        object_key = _s3_path(prefix=self._s3_parameters.path, object_name=backup_id)
+        try:
+            self._client.head_object(Bucket=self._s3_parameters.bucket, Key=object_key)
+        except ClientError as exc:
+            if "Error" in exc.response and exc.response["Error"].get("Code") == "404":
+                return False
+            logger.exception("Error checking object")
```
arturo-seijas, line 213: This message seems a bit vague to me

src/backup_observer.py
```diff
@@ -168,6 +169,16 @@ def _on_restore_backup_action(self, event: ActionEvent) -> None:
             event.fail("Wrong S3 configuration on restore backup action. Check S3 integration.")
             return
 
+        try:
+            s3_client = backup.S3Client(s3_parameters)
+            if not s3_client.exists_backup(backup_id):
+                event.fail(f"backup-id {backup_id} does not exist")
```
arturo-seijas, line 175: I'm wondering if we should fail here or exit with success and still print this output so that the action is idempotent

src/backup_observer.py
```diff
@@ -176,10 +187,39 @@ def _on_restore_backup_action(self, event: ActionEvent) -> None:
...
+            s3_parameters = backup.S3Parameters(**self._s3_client.get_s3_connection_info())
+        except ValueError:
+            logger.exception("Wrong S3 configuration in delete backup action")
+            event.fail("Wrong S3 configuration in delete backup action. Check S3 integration.")
+            return
+
+        try:
+            s3_client = backup.S3Client(s3_parameters)
+            if not s3_client.exists_backup(backup_id):
+                event.fail("Backup to delete does not exist.")
```
arturo-seijas, line 217: I'm wondering if we should fail here or exit with success and still print this output so that the action is idempotent

javierdelapuente: I was thinking in `rm` command that fails if the file does not exist. Open to change it if you still think that this action should behave otherwise...

arturo-seijas: I'm having a similar discussion [her](https://github.com/canonical/digest-squid-auth-helper/pull/2#discussion_r1484034353). So it'd be a good idea to agree on an approach @gregory-schiano

nrobinaubertin: I answered on the other side also but same thing: I would issue a warning (or log it) and finish with success

javierdelapuente: Changed so the action does not fail (`status: completed` instead of `status: failed`), but the result message says the backup does not exist and a warning is logged.

arturo-seijas (dismissed):

src/backup.py
```diff
@@ -176,6 +176,45 @@ def can_use_bucket(self) -> bool:
...
+            True if the backup-id exists, False otherwise
+
+        Raises:
+            S3Error: If there was an error checking the backup.
+        """
+        object_key = _s3_path(prefix=self._s3_parameters.path, object_name=backup_id)
+        try:
+            self._client.head_object(Bucket=self._s3_parameters.bucket, Key=object_key)
+        except ClientError as exc:
+            if "Error" in exc.response and exc.response["Error"].get("Code") == "404":
```
cbartz, line 211: Could we use 401 and 403 to indicate invalid credentials/insufficient permissions and let the user know with an appropriate message/BlockedStatus? However, this may apply to all s3 operations and may be outside the scope of this PR.

tests/unit/test_backup.py
```diff
@@ -172,6 +172,126 @@ def test_can_use_bucket_bucket_error(s3_parameters_backup, monkeypatch: pytest.M
...
+
+    key = f"{s3_parameters_backup.path.strip('/')}/{backup_id}"
+    delete_object_mock.assert_called_once_with(Bucket=s3_parameters_backup.bucket, Key=key)
+
+
+def test_delete_backup_boto_client_error(s3_parameters_backup, monkeypatch: pytest.MonkeyPatch):
+    """
+    arrange: Create a S3Client. On delete_object return a boto exception.
+    act: Run delete_backup.
+    assert: The function delete_object throws and exception.
```
cbartz, line 208:
```suggestion
    assert: The function delete_object throws an exception.
```

cbartz (approved):
LGTM

src/backup.py
javierdelapuente: It is a good point. I would leave it for another PR as it would probably be better if checked in another function like `can_use_bucket` and maybe change a bit how it works.

cbartz (approved):

arturo-seijas (approved):

PR #197, opened 2024-02-09
Remove COS integration test

amandahla:
### Overview

This PR remove COS during the integration tests.

### Rationale

Since COS integration is not required for Synapse to work, we decided to remove it from the integration tests to make them faster. 

### Checklist


.github/workflows/integration_test.yaml
```diff
@@ -8,7 +8,7 @@ jobs:
     uses: canonical/operator-workflows/.github/workflows/integration_test.yaml@main
     secrets: inherit
     with:
-      extra-arguments: --localstack-address 172.17.0.1
+      extra-arguments: --localstack-address 172.17.0.1 -m "not cos"
```
cbartz, line 11: Would it make more sense to remove these tests? If they're not run, they'll probably stop working after a while. As they are checked in version control, they can be restored later if you wish.

amandahla: We are planning to run them before publishing to edge :)

cbartz: Gotcha, could you please add this to the PR description?

cbartz (approved):

.github/workflows/integration_test.yaml
amandahla: Sure, thanks for raising this.

nrobinaubertin:
How will we test them before pushing to edge ? Will there be an action ?

amandahla:
> How will we test them before pushing to edge ? Will there be an action ?

I was planning to create a branch in operator-workflows and change the publish charm by adding a step to run skipped integration tests but for this COS case I'm very tempted to just remove them. What do you think?

nrobinaubertin:
> > How will we test them before pushing to edge ? Will there be an action ?
> 
> I was planning to create a branch in operator-workflows and change the publish charm by adding a step to run skipped integration tests but for this COS case I'm very tempted to just remove them. What do you think?

I actually agree, the COS tests that we currently have don't seem to really test the synapse charm per se. Those tests could be only done in staging

javierdelapuente:
> > How will we test them before pushing to edge ? Will there be an action ?
> 
> I was planning to create a branch in operator-workflows and change the publish charm by adding a step to run skipped integration tests but for this COS case I'm very tempted to just remove them. What do you think?

I am not sure we should test at that point. I would rather delete it (of leave the test where it is). At that point the code is in main, and we will have broken CI/CD there that will require another PR....

jdkandersson:
> > > How will we test them before pushing to edge ? Will there be an action ?
> > 
> > 
> > I was planning to create a branch in operator-workflows and change the publish charm by adding a step to run skipped integration tests but for this COS case I'm very tempted to just remove them. What do you think?
> 
> I am not sure we should test at that point. I would rather delete it (of leave the test where it is). At that point the code is in main, and we will have broken CI/CD there that will require another PR....

@merkata did we discuss writing a spec for e2e tests for how to do COS testing? the comment from @javierdelapuente is quite reasonable that we should not have a test that isn't run

gtrkiller (approved):

cbartz (approved):

PR #212, opened 2024-02-20
WIP. Trying to not create the charm state in init

javierdelapuente:
### Checklist

### DIFF
--- a/src/charm_state.py
+++ b/src/charm_state.py
@@ -6,6 +6,7 @@ import dataclasses
 import itertools
 import os
 import typing
+from abc import ABC, abstractmethod
 
 import ops
 
@@ -21,6 +22,55 @@ from pydantic import (  # pylint: disable=no-name-in-module,import-error
 )
 
 from charm_types import DatasourcePostgreSQL, SAMLConfiguration, SMTPConfiguration
+
+
+def inject_charm_state(  # pylint: disable=protected-access
+    method: typing.Callable[[typing.Any, typing.Any, "CharmState"], None]
+) -> typing.Callable[[typing.Any, typing.Any], None]:
+    """Workaround until this is fixed https://github.com/canonical/operator/issues/1129.
+
+    Args:
+        method: observer method to wrap and inject the charm_state
+
+    Returns:
+        the function wrapper
+    """
+
+    def wrapper(theself: typing.Any, event: typing.Any) -> None:
+        """Add the charm_state argument to the function.
+
+        Args:
+            theself: the self in the Object
+            event: the event for the observer
+
+        Returns:
+            The value returned from the original function. That is, None.
+
+        Raises:
+            TypeError: if the function wrapped is invalid
+        """
+        try:
+            if hasattr(theself, "build_charm_state"):
+                charm_state = theself.build_charm_state()
+            elif hasattr(theself, "_charm") and hasattr(theself._charm, "build_charm_state"):
+                charm_state = theself._charm.build_charm_state()
+            else:
+                raise TypeError("Cannot inject charm state into observer method")
+        except CharmConfigInvalidError as exc:
+            theself.model.unit.status = ops.BlockedStatus(exc.msg)
+            return None
+        return method(theself, event, charm_state)
+
+    setattr(wrapper, "__name__", method.__name__)
+    return wrapper
+
+
+class CharmBaseWithState(ops.CharmBase, ABC):
+    """CharmBase than can build a CharmState."""
+
+    @abstractmethod
+    def build_charm_state(self) -> "CharmState":
+        """Build charm state."""
 
 
 class CharmConfigInvalidError(Exception):
--- a/src/mjolnir.py
+++ b/src/mjolnir.py
@@ -11,9 +11,10 @@ import typing
 
 import ops
 
+import pebble
 import synapse
 from admin_access_token import AdminAccessTokenService
-from charm_state import CharmState
+from charm_state import CharmBaseWithState, CharmState, inject_charm_state
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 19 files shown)


src/charm_state.py
```diff
@@ -23,6 +24,69 @@
...
+
+        try:
+            charm_state = charm.build_charm_state()
+        except CharmConfigInvalidError as exc:
+            charm.model.unit.status = ops.BlockedStatus(exc.msg)
+            return None
+        return method(instance, event, charm_state)
+
+    # This is necessary for ops to work
+    setattr(wrapper, "__name__", method.__name__)
```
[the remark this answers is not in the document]
nrobinaubertin: Maybe add the issue link as a comment

javierdelapuente:
> @gregory-schiano @jdkandersson I wonder if we are not getting to the point of going back and using [stored state ](https://juju.is/docs/sdk/stored-state-uses-limitations) due complexity of managing the state in the charm.

Thank for the input @amandahla. Can you expand on why you think there is an increase in the complexity related to the stored state? The current PR does not change how the CharmState instance is built, only when it is built (instead of the `__init__` method, in every observer that needs it).

gregory-schiano:
> > @gregory-schiano @jdkandersson I wonder if we are not getting to the point of going back and using [stored state ](https://juju.is/docs/sdk/stored-state-uses-limitations) due complexity of managing the state in the charm.
> 
> Thank for the input @amandahla. Can you expand on why you think there is an increase in the complexity related/similar to the stored state? The current PR does not change how the CharmState instance is built, only when it is built (instead of the `__init__` method, in every observer that needs it).

Indeed I agree with Javier, also I don't see how using a stored state will help as the concept of CharmState (a meaningful object representing the workload state composed of relation data, configuration data, and workload information) will still be needed.

This PR aims at fixing the design issues we having with the state being instantiated in the constructor before the events are observed leading to issues if the state fails to load

amandahla:
> > > @gregory-schiano @jdkandersson I wonder if we are not getting to the point of going back and using [stored state ](https://juju.is/docs/sdk/stored-state-uses-limitations) due complexity of managing the state in the charm.
> > 
> > 
> > Thank for the input @amandahla. Can you expand on why you think there is an increase in the complexity related/similar to the stored state? The current PR does not change how the CharmState instance is built, only when it is built (instead of the `__init__` method, in every observer that needs it).
> 
> Indeed I agree with Javier, also I don't see how using a stored state will help as the concept of CharmState (a meaningful object representing the workload state composed of relation data, configuration data, and workload information) will still be needed. The stored state will become another source of data (mostly replacing workload information)
> 
> This PR aims at fixing the design issues we having with the state being instantiated in the constructor before the events are observed leading to issues if the state fails to load

Actually this PR just made me think more about state and how is turning out to be more like a layer in front of ops since we are using it instead of getting information directly from relations/unit/configuration.

I remember StoredState because it was easier to understand that we are getting information from somewhere but now having the state concept is a little bit tricky if you are not familiar with event lifecycle/relations.

I wonder if instead of passing state around wouldn't be easier to just use ops and get what we need in the time we need it.

amandahla:
@javierdelapuente approving the PR but just using it to discuss the theory ;)

amandahla (approved):
LGTM nice work with Python!

yanksyoon (approved):
I believe we've had this discussion in GitHub runner as well, it would help the team greatly if we have a post written on the reasons to do this (if we don't already have it) :D

javierdelapuente:
> I believe we've had this discussion in GitHub runner as well, it would help the team greatly if we have a post written on the reasons to do this (if we don't already have it) :D

Do you mean creating the CharmState outside of `__main__` or more about the reasons for the CharmState?

amandahla (approved):

merkata:
> I believe we've had this discussion in GitHub runner as well, it would help the team greatly if we have a post written on the reasons to do this (if we don't already have it) :D

Agree we might want to expand and structure that idea in a post, that's good!

amandahla (dismissed):
LGTM

javierdelapuente:
> @javierdelapuente I meant having initialization of CharmState outside of **init**, most likely for each function just before they get used. Here's a reference to [setup_state](https://github.com/canonical/github-runner-operator/pull/199/files#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307cR202) outside of charm `__init__` on github-runner-operator charm. The reason for this here was to fully complete initialization and register the handlers without failure. The bug we encountered was that the actions wouldn't get registered due to `CharmState.from_charm` raising exceptions and thus not completing the handler registration (`framework.observe`s).

I will have a look and update ISD014, as it recommends to initialize the state in the init method. Besides the issue of the handlers not registered (number 2 in the overview), we also had another problem (data library needs the event to refresh data in the relation and so the data is not yet there in init, number 1), and I believe using secrets-changed will be difficult to manage this way (number 3).

amandahla (approved):

src/mjolnir.py
```diff
@@ -167,7 +166,7 @@ def enable_mjolnir(self, admin_access_token: str) -> None:
             USERNAME,
             True,
             admin_access_token,
-            str(self._charm_state.synapse_config.server_name),
+            str(charm_state.synapse_config.server_name),
```
yanksyoon, line 169: Out of curiosity, is `str()` cast required? I think this field is already of str type :D

yanksyoon (approved):
LGTM :D

src/mjolnir.py
yanksyoon: Please feel free to resolve this in follow up PR since it seems that it has good amount of approvals already :)

javierdelapuente: I will solve it in a next PR as it it not related to this PR. Thanks for the input!

PR #217, opened 2024-02-21
Add Synapse Stats Exporter pebble layer

amandahla:
### Overview

This PR:
- Add Synapse Stats Exporter pebble layer if PostgreSQL is available

If the configurations are set, the Synapse Stats Exporter will be enabled to collect statistics. Synapse Stats Exporter is prepared to collect metrics from the database.

Since the Exporter collects data directly from the database there is no need to create a user depending on the Synapse API. Also, since we own Synapse Stats Exporter, we can create custom queries to collect data.

### Rationale

Monitor Synapse statistics via Prometheus + Grafana. The statistics available so far are the number of users and the number of rooms.

### Juju Events Changes

Config Changed Event checks if the datasource is available and then sets the Synapse Stats Exporter's pebble layer.

### Module Changes

New pebble layer added for starting Synapse Stats Exporter together with Synapse.

### Library Changes

N/A

### Checklist


config.yaml
```diff
@@ -60,6 +60,15 @@ options:
       Synapse server name. Must be set to deploy the charm. Corresponds to the
       server_name option on Synapse configuration file and sets the
       public-facing domain of the server.
+  stats_exporter_user:
```
arturo-seijas, line 63: Is this a synapse user? If so, will it be possible to create it on deployment? Otherwise we'll have a chicken egg situation, as I understand it

amandahla: Thanks, we discussed a little bit about it in the daily so I decided to change the approach again and now the Synapse Stats Exporter connects to the database to get metrics so there is no need of API user.

jdkandersson:
I'm hoping to understand please:
* What is the reason for needing the database for statistics? And what are the statistics, is it things like user interactions?
* Why not include the docs in this PR? I'm guessing that before closing the story the docs need to be finished?

src/charm.py
```diff
@@ -222,6 +222,11 @@ def _on_reset_instance_action(self, event: ActionEvent, charm_state: CharmState)
             return
         try:
             self.model.unit.status = ops.MaintenanceStatus("Resetting Synapse instance")
+            try:
+                container.stop(pebble.STATS_EXPORTER_SERVICE_NAME)
+            except (ops.pebble.Error, RuntimeError) as e:
```
jdkandersson, line 227: Are we sure that the service was stopped if an exception is raised?

src/pebble.py
```diff
@@ -69,6 +71,31 @@ def replan_mjolnir(container: ops.model.Container) -> None:
...
+            "PROM_SYNAPSE_HOST": datasource["host"],
+            "PROM_SYNAPSE_PORT": datasource["port"],
+            "PROM_SYNAPSE_USER": datasource["user"],
+            "PROM_SYNAPSE_PASSWORD": datasource["password"],
+        }
+        try:
+            container.add_layer(STATS_EXPORTER_SERVICE_NAME, layer, combine=True)
+            container.start(STATS_EXPORTER_SERVICE_NAME)
+        except ops.pebble.Error as e:
+            logger.debug("Ignoring error while restarting Synapse Stats Exporter")
```
jdkandersson, line 95: What is the reason for ignoring please?

src/charm.py
amandahla: No, but if the container dont stop, the action will fail anyway since if there are open connections in the database:
"database "synapse" is being accessed by other users"
https://github.com/canonical/synapse-operator/actions/runs/8207941891/job/22450994126

jdkandersson: Perhaps if we encounter an error we fail at that action at that point then?

src/pebble.py
jdkandersson: Ok, do you think that is worth adding as a comment? Ignoring errors is unusual so anyone reading the code might have a similar thought 🙂

src/charm.py
amandahla: Good suggestion, I changed to fail the event at this point.

src/pebble.py
amandahla: Sure, added!

docs/explanation/charm-architecture.md
```diff
@@ -74,6 +74,11 @@ forward non-static traffic to it.
 
 The workload that this container is running is defined in the [Synapse ROCK](https://github.com/canonical/synapse-operator/tree/main/synapse_rock).
 
+If Synapse is integrated with PostgreSQL, [Synapse Stats Exporter](https://github.com/canonical/synapse_stats_exporter) will be enabled.
+Synapse Stats Exporter listens to non-TLs POR `9877` and will be configured as a
```
jdkandersson, line 78: `Port`?

jdkandersson (approved):

tests/integration/test_charm.py
```diff
@@ -66,6 +66,29 @@ async def test_synapse_validate_configuration(synapse_app: Application):
...
 
+async def test_enable_stats_exporter(
+    synapse_app: Application,
+    synapse_app_name: str,
+    get_unit_ips: typing.Callable[[str], typing.Awaitable[tuple[str, ...]]],
+) -> None:
+    """
+    arrange: Synapse is integrated with Postgresql.
+    act:  request Synapse Stats Exporter URL.
+    assert: the Synapse application is active and the Exporter returns as expected.
```
cbartz, line 77: The statement "the Synapse application is active" is not asserted after the act phase, am I right? This is a bit confusing.

cbartz (commented):
Looks good, one small comment.

docs/explanation/charm-architecture.md
yanksyoon:
```suggestion
Synapse Stats Exporter listens to non-TLS port `9877` and will be configured as a
```
also TLS :D

yanksyoon (approved):
Minor comments, LGTM!

tests/integration/test_charm.py
amandahla: Good catch, changed.

javierdelapuente (approved):

nrobinaubertin (approved):

PR #235, opened 2024-03-06
Add refresh admin access token when invalid

javierdelapuente:
### Overview

This PR does mainly two things:
 - Always uses a retry when connecting to the workload in the socket connection timeout. This allows to wait for the workload when it is starting. For the read timeout, it will use retry as it was before this PR.
 - When getting the admin access token, check if it is valid (no 401 on using it for dummy whoami request). If it is not valid, create/recreate it. This fixes the token after actions like reset_instance, restore backup or integrate with a database.

### Rationale

This PR fixes a bug that happens when an admin access token is in the peer relation and reset_instance or restore backup or integrate with a database happen, in which case the token is invalid and is not refreshed anymore. In this case, several things stop working, like Mjolnir, room creation...

Changing the retry policy to retry on connect was added because the new integration test was failing for that reason.

### Module Changes

New function in synapse.api that checks if the token is valid making a request to `_matrix/client/v3/account/whoami`

Some refactor in the `AdminAccessTokenService` to refresh the token if it is not valid.

### Checklist

### DIFF
--- a/src/admin_access_token.py
+++ b/src/admin_access_token.py
@@ -12,6 +12,7 @@ import ops
 from ops.jujuversion import JujuVersion
 
 import synapse
+from exceptions import PeerRelationNotFoundError
 
 logger = logging.getLogger(__name__)
 
@@ -61,13 +62,16 @@ class AdminAccessTokenService:
 
         Returns:
             Admin access token or None if admin token was not found or there was an error.
+
+        Raise:
+            PeerRelationNotFoundError: if the peer relation does not exist.
         """
         peer_relation = self._model.get_relation(PEER_RELATION_NAME)
         if not peer_relation:
             logger.error(
                 "Failed to get admin access token: no peer relation %s found", PEER_RELATION_NAME
             )
-            return None
+            raise PeerRelationNotFoundError()
         admin_access_token = None
         if JUJU_HAS_SECRETS:
             secret_id = peer_relation.data[self._app].get(SECRET_ID)
@@ -93,13 +97,16 @@ class AdminAccessTokenService:
 
         Returns:
             admin access token or None if there was an error.
+
+        Raise:
+            PeerRelationNotFoundError: if the peer relation does not exist.
         """
         peer_relation = self._model.get_relation(PEER_RELATION_NAME)
         if not peer_relation:
             logger.error(
                 "Failed to get admin access token: no peer relation %s found", PEER_RELATION_NAME
             )
-            return None
+            raise PeerRelationNotFoundError()
         admin_user = synapse.create_admin_user(container)
         if not admin_user:
             logger.error("Error creating admin user to get admin access token")
--- a/src/admin_access_token.py
+++ b/src/admin_access_token.py
@@ -80,8 +80,6 @@ class AdminAccessTokenService:
                     del peer_relation.data[self._app][SECRET_ID]
                     return None
         else:
-            # There is no Secrets support and none relation data was created
-            # So lets create the user and store its token in the peer relation
             secret_value = peer_relation.data[self._app].get(SECRET_KEY)
             if secret_value:
                 admin_access_token = secret_value
--- a/src/admin_access_token.py
+++ b/src/admin_access_token.py
@@ -104,7 +104,7 @@ class AdminAccessTokenService:
             return None
         admin_user = synapse.create_admin_user(container)
         if not admin_user:
-            logger.error("Error getting admin access token")
+            logger.error("Error creating admin user to get admin access token")
             return None
         if JUJU_HAS_SECRETS:
             logger.debug("Adding/ secret")
--- a/src/admin_access_token.py
+++ b/src/admin_access_token.py
@@ -112,12 +112,12 @@ class AdminAccessTokenService:
             logger.error("Error creating admin user to get admin access token")
             return None
         if JUJU_HAS_SECRETS:
-            logger.debug("Adding/ secret")
+            logger.debug("Adding admin_access_token secret to peer relation")
             secret = self._app.add_secret({SECRET_KEY: admin_user.access_token})
             peer_relation.data[self._app].update({SECRET_ID: secret.id})
             admin_access_token = admin_user.access_token
         else:
-            logger.debug("Adding peer data")
+            logger.debug("Adding admin_access_token to peer relation")
             peer_relation.data[self._app].update({SECRET_KEY: admin_user.access_token})
             admin_access_token = admin_user.access_token
         return admin_access_token
--- a/src/admin_access_token.py
+++ b/src/admin_access_token.py
@@ -41,13 +41,26 @@ class AdminAccessTokenService:
         self._model = model
 
     def get(self, container: ops.Container) -> typing.Optional[str]:
-        """Get admin access token.
+        """Get admin access token from peer relation.
+
+        If the admin token is not valid or it does not exist it creates one.
 
         Args:
             container: Workload container.
 
         Returns:
             admin access token or None if fails.
+        """
+        admin_access_token = self._get_from_peer_relation()
+        if admin_access_token is None or not synapse.is_token_valid(admin_access_token):
+            admin_access_token = self._renew_token(container)
+        return admin_access_token
+
+    def _get_from_peer_relation(self) -> typing.Optional[str]:
+        """Get admin access token from peer relation.
+
+        Returns:
+            Admin access token or None if not found or failure.
         """
         peer_relation = self._model.get_relation(PEER_RELATION_NAME)
         if not peer_relation:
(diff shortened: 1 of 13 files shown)


src/admin_access_token.py
```diff
@@ -66,27 +79,38 @@ def get(self, container: ops.Container) -> typing.Optional[str]:
...
+
+        Returns:
+            admin access token or None if there was an error.
+        """
+        peer_relation = self._model.get_relation(PEER_RELATION_NAME)
+        if not peer_relation:
+            logger.error(
+                "Failed to get admin access token: no peer relation %s found", PEER_RELATION_NAME
+            )
+            return None
```
arturo-seijas, line 102: I'm wondering if it would be better to raise exceptions for these conditions instead of returning None

javierdelapuente: I did not want to change the interface to the module, but I believe you are right, that is too bad to let it happen without failing. Updated to raise an exception. Thanks for the idea!

tests/integration/test_charm.py
```diff
@@ -395,3 +395,58 @@ async def test_synapse_with_mjolnir_from_refresh_is_up(
         f"http://{synapse_ip}:{synapse.MJOLNIR_HEALTH_PORT}/healthz", timeout=5
     )
     assert mjolnir_response.status_code == 200
+
+
+async def test_admin_token_refresh(model: Model, synapse_app: Application):
+    """
+    arrange: Build and deploy the Synapse charm from charmhub.
+             Create an user.
```
cbartz, line 403:
```suggestion
             Create a user.
```

tests/unit/test_admin_access_token.py
```diff
@@ -71,16 +73,62 @@ def test_get_admin_access_token_no_secrets(
...
+    monkeypatch.setattr(admin_access_token, "JUJU_HAS_SECRETS", juju_has_secrets)
+
+    # Get admin access token
+    harness.begin_with_initial_hooks()
+    monkeypatch.setattr(synapse, "create_admin_user", create_admin_user_mock)
+    initial_admin_access_token = harness.charm.token_service.get(MagicMock)
+    is_token_valid_mock.assert_not_called()
+    assert initial_admin_access_token == initial_token
+
+    # Get admin access token. Should not be refreshed as it is valid.
```
cbartz, line 128:
```suggestion
    # Get admin access token. Should not be refreshed if it is valid.
```

cbartz (commented):
Looks good, have a few nitpicks. Could you put the reason for the PR in the rationale next time? Thank you.

src/admin_access_token.py
```diff
@@ -41,20 +42,36 @@ def __init__(self, app: ops.Application, model: ops.Model):
...
+        Raise:
+            PeerRelationNotFoundError: if the peer relation does not exist.
+        """
         peer_relation = self._model.get_relation(PEER_RELATION_NAME)
         if not peer_relation:
             logger.error(
                 "Failed to get admin access token: no peer relation %s found", PEER_RELATION_NAME
             )
-            return None
+            raise PeerRelationNotFoundError()
```
arturo-seijas, line 74: Isn't line 75 unreachable now?

src/admin_access_token.py
```diff
@@ -66,27 +83,41 @@ def get(self, container: ops.Container) -> typing.Optional[str]:
...
+        peer_relation = self._model.get_relation(PEER_RELATION_NAME)
+        if not peer_relation:
+            logger.error(
+                "Failed to get admin access token: no peer relation %s found", PEER_RELATION_NAME
+            )
+            raise PeerRelationNotFoundError()
+        admin_user = synapse.create_admin_user(container)
+        if not admin_user:
+            logger.error("Error creating admin user to get admin access token")
+            return None
```
arturo-seijas, line 113: How about raising an exception here too?

src/admin_access_token.py
javierdelapuente: it is reacheble (otherwise this function would always fail)

src/admin_access_token.py
javierdelapuente: I believe there is code in Synapse that depends on that, I would rather not do it in this PR

javierdelapuente:
> Looks good, have a few nitpicks. Could you put the reason for the PR in the rationale next time? Thank you.

Oops sorry, added the rationale part!

src/admin_access_token.py
```diff
@@ -41,20 +42,36 @@ def __init__(self, app: ops.Application, model: ops.Model):
...
+
+        If the admin token is not valid or it does not exist it creates one.
 
         Args:
             container: Workload container.
 
         Returns:
             admin access token or None if fails.
         """
+        admin_access_token = self._get_from_peer_relation()
```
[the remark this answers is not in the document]
cbartz: This was introduced in https://github.com/canonical/is-charms-contributing-guide/pull/68, so it might be worth discussing with @jdkandersson.

src/admin_access_token.py
arturo-seijas: Oops, my bad

cbartz (approved):

arturo-seijas (dismissed):

src/admin_access_token.py
```diff
@@ -66,27 +83,41 @@ def get(self, container: ops.Container) -> typing.Optional[str]:
...
+            logger.error(
+                "Failed to get admin access token: no peer relation %s found", PEER_RELATION_NAME
+            )
+            raise PeerRelationNotFoundError()
+        admin_user = synapse.create_admin_user(container)
+        if not admin_user:
+            logger.error("Error creating admin user to get admin access token")
+            return None
+        if JUJU_HAS_SECRETS:
+            logger.debug("Adding/ secret")
```
merkata, line 115: There's a weird slash character here.

merkata (approved):
Great!

src/admin_access_token.py
javierdelapuente: oops. Fixed. Thanks!

src/admin_access_token.py
amandahla, line 55: Shouldnt catch PeerRelationNotFound?

src/synapse/api.py
```diff
@@ -491,3 +506,20 @@ def promote_user_admin(
...
+    Args:
+        access_token: server access token to be used.
+
+    Returns:
+        If the token is valid or not.
+    """
+    try:
+        _do_request("GET", WHOAMI_URL, admin_access_token=access_token, retry=True)
+    except UnauthorizedError:
+        logger.info("Invalid access token ")
```
amandahla, line 523: super tiny minor comment: space after token

src/synapse/api.py
```diff
@@ -126,18 +131,26 @@ def _do_request(
...
         Response from the request.
     """
     try:
         session = requests.Session()
+        # By default always retry on connect (and in server errors). Failing on connect
+        # should mean that the server has not started to process the request.
+        # Only retry on read if retry set to True, as in a non-idempotent operation
+        # this may have undesired effects.
         retries = Retry(
+            connect=3,
```
amandahla, line 146: I don't remember: it shows in the logs when try to retry, right?

src/admin_access_token.py
javierdelapuente: Thanks for bringing this back to the foreground.

I left the exception uncaught as I thought that was a situation that is clearly something very wrong (not having a peer relation).

Reviewing the documentation peer relations events are part of the [setup phase](https://juju.is/docs/sdk/charm-lifecycle#heading--notes-on-the-setup-phase), and occur only after `install`. It also says:
```
The reason for this is that a peer relation cannot be created or destroyed ‘manually’ at arbitrary times, they either exist or not, and if they do exist, then we know it from the start.
```

Checking manually, I see that in the `install` event we do not have the peer relation (or at least it is None using `charm.model.get_relation("synapse-peers")`). It is in all other events in the setup phase, and it disappears in the teardown phase, after the `synapse_peers_relation_broken` event.

As we are using the collect status ops event to invoke logic in the container, that means that we may in effect setting the unit to error in at least the teardown phase (haven't tested it).

As a precaution, I am going to change it to return None instead of raising/catching that exception. This was the same logic that was before the PR. I will add a note to the story about the collect status to review these cases in that story.

amandahla: Good investigation, thanks.

src/synapse/api.py
amandahla: Honestly, I'm curious about how the charm will behave since many events depend on API calls. I don't know if there is some kind of charm event timeout if, for example, a request takes all the retries to be completed.

javierdelapuente: AFAIK, juju events by default do not have a timeout, so retrying three times should not be a problem (it should not add more than a few seconds, then it raises)

amandahla (approved):
LGTM :rocket:

cbartz (approved):

PR #243, opened 2024-03-11
Block while scaling if Synapse is not integrated with Redis

amandahla:
### Overview

This PR checks for Redis integration before proceeding with the principal/worker configuration.

This is done by using the peer relation. 

So, for every unit that joins the peer relation, Redis is checked and if not present, the charm is blocked. 

Also, for every unit that departs from the relation, if only one unit remains, then there is no need of Redis and the charm can be set as Active.

### Rationale

Redis is required for horizontal scaling.

### Juju Events Changes

_on_relation_joined
_on_relation_departed

### Module Changes

N/A

### Library Changes

N/A

### Checklist


src/charm.py
```diff
@@ -182,6 +189,45 @@ def _on_config_changed(self, _: ops.HookEvent, charm_state: CharmState) -> None:
...
 
+    @inject_charm_state
+    def _on_relation_joined(self, _: ops.HookEvent, charm_state: CharmState) -> None:
+        """Handle Synapse peer relation joined event.
+
+        Args:
+            charm_state: The charm state.
+        """
+        if charm_state.redis_config is None:
+            logger.debug("Scaling is not allowed without Redis integration.")
```
javierdelapuente, line 200: Won't this happen also if there is only one unit in the application?

amandahla: I hope not. I'm considering this:
“For a peer relation, <peer relation name>-relation-joined will only be emitted if the scale is larger than 1. In other words, applications with a scale of 1 do not see peer relation joined/departed events. If you are using peer data as a means for persistent storage, then use peer relation-created instead.”
https://juju.is/docs/sdk/relation-name-relation-joined-event

javierdelapuente: TIL

javierdelapuente (approved):
LGTM

src/charm.py
```diff
@@ -182,6 +189,45 @@ def _on_config_changed(self, _: ops.HookEvent, charm_state: CharmState) -> None:
...
+
+    @inject_charm_state
+    def _on_relation_departed(self, event: RelationDepartedEvent, charm_state: CharmState) -> None:
+        """Handle Synapse peer relation departed event.
+
+        Args:
+            event: relation departed event.
+            charm_state: The charm state.
+        """
+        if not self.unit.is_leader() or event.departing_unit == self.unit:
```
arturo-seijas, line 215: Not sure I get the `event.departing_unit == self.unit` part

amandahla: This check is because all units including the one departing receive this event. Thats why the charm will ignore it if not the leader or if is the unit removed/departing.

Reference:
https://juju.is/docs/sdk/relation-name-relation-departed-event#heading--observing-this-event-in-ops

arturo-seijas: Shouldn't it be the other way around? `if not self.unit.is_leader() or event.departing_unit != self.unit`

src/charm.py
```diff
@@ -182,6 +189,45 @@ def _on_config_changed(self, _: ops.HookEvent, charm_state: CharmState) -> None:
...
+        if peer_relation is not None:
+            # if more than 1, check for Redis integration
+            synapse_units = len(peer_relation.units)
+            if synapse_units > 1:
+                logger.debug("Found %d units, checking for Redis integration.", synapse_units)
+                if charm_state.redis_config is None:
+                    logger.debug("Scaling is not allowed without Redis integration.")
+                    self.model.unit.status = ops.BlockedStatus(
+                        "Redis integration not found. Please, verify it."
+                    )
```
javierdelapuente, line 228: I did not see it before, I think you missed a return here.

amandahla: Thanks!

src/charm.py
arturo-seijas: Oh, I see. Makes sense to me

PR #252, opened 2024-03-18
Chore/refactaroo for Ops in synapse/workload

merkata:
### Overview

Refactors some bits from workload to not rely on `ops` - currently the charm passes a container to src/pebble which then passes it to src/synapse/workload and the workload makes a lot of (duplicate) things that are perhaps better handled at one layer and not across. 

Note that this is against the irc bridge branch, as the proposal to refactor came in there and if this gets accepted it will land in the PR for the irc bridge.

### Rationale

Reduce code and complexity, make dev person happi.

### Juju Events Changes

n/a

### Module Changes

moved Ops pieces (checks and config manipulation) to pebble, still WIP

### Library Changes

n/a

### Checklist


src/pebble.py
```diff
@@ -121,36 +224,36 @@ def change_config(charm_state: CharmState, container: ops.model.Container) -> No
     """
     try:
         synapse.execute_migrate_config(container=container, charm_state=charm_state)
-        synapse.enable_metrics(container=container)
-        synapse.enable_forgotten_room_retention(container=container)
-        synapse.enable_serve_server_wellknown(container=container)
+        current_yaml = _get_synapse_config(container)
```
nrobinaubertin, line 227: I wonder if `current_yaml` is a good name. Maybe `current_synapse_config` ?

src/pebble.py
```diff
@@ -121,36 +224,36 @@ def change_config(charm_state: CharmState, container: ops.model.Container) -> No
...
-            synapse.enable_trusted_key_servers(container=container, charm_state=charm_state)
+            synapse.enable_trusted_key_servers(current_yaml, charm_state=charm_state)
         if charm_state.synapse_config.ip_range_whitelist:
-            synapse.enable_ip_range_whitelist(container=container, charm_state=charm_state)
+            synapse.enable_ip_range_whitelist(current_yaml, charm_state=charm_state)
         if charm_state.datasource:
             logger.info("Synapse Stats Exporter enabled.")
             replan_stats_exporter(container=container, charm_state=charm_state)
-        synapse.validate_config(container=container)
+        synapse.validate_config(container)
```
amandahla, line 255: The yaml should be updated in the container before running validate_config

amandahla: Or could be refactored to run in a temporary file, it would work too.

amandahla:
Nice refactor, thanks!

src/pebble.py
merkata: Good catch! I moved it after pushing (for now).

src/pebble.py
merkata: It is better, renamed. Thanks!

PR #266, opened 2024-03-26
Set worker name with unit ID

amandahla:
### Overview

This PR set the worker name to use the Unit ID as part of the name.

### Rationale

Horizonal scaling requires that each worker has a unique name.

### Checklist


src/charm.py
```diff
@@ -124,6 +124,24 @@ def is_main(self) -> bool:
...
+            unit_name: unit name. E.g.: synapse/0
+
+        Returns:
+            Unit number. E.g.: 0
+        """
+        if not unit_name:
+            unit_name = self.unit.name
+        match = re.search(r"[-\/](\d+)", unit_name)
+        # A Juju unit name is s always named on the
+        # pattern <application>/<unit ID>, where <application> is the name
```
yhaliaw, line 140: Would using regex a bit overkill if the pattern is this simple?

yanksyoon (approved):

src/charm.py
```diff
@@ -124,6 +124,24 @@ def is_main(self) -> bool:
...
+
+        Args:
+            unit_name: unit name. E.g.: synapse/0
+
+        Returns:
+            Unit number. E.g.: 0
+        """
+        if not unit_name:
+            unit_name = self.unit.name
+        match = re.search(r"[-\/](\d+)", unit_name)
```
cbartz, line 138: The regex also matches `-0` in `synapse-0` (I think you replace / with - in another part of the code). You might want to add this to the docstring (otherwise it might be a bit confusing for the reader of the code).

src/charm.py
amandahla: Actually the comment is outdated. This pattern is for two strings:
synapse/0 and synapse-0.synapse.endpoints
Maybe my mind is too shell-oriented :laughing: 
How would you do without regex?

src/charm.py
amandahla: Indeed, but I'm using the first group, not the match :)

This is a nice tool that I usually test regular expressions:
https://regex101.com/r/YHoPN0/1

I'll add it to docstring.

amandahla: Actually...From your comment and @yhaliaw I'll remove the regex and replace for something like this. Is more readable?

```python
    index = unit_name.rfind('/') # synapse/0 pattern
    if index == -1:
        index = unit_name.rfind('-') # synapse-0.synapse-endpoints pattern
   return unit_name[index + 1:]
```

amandahla: Replaced by split + rfind :)

src/charm.py
amandahla: Replaced by split + rfind
Please, if you want to comment something more regarding this I'll leave the https://github.com/canonical/synapse-operator/pull/266#discussion_r1540590003 open so you and Bartz can see it.

tests/unit/test_charm_scaling.py
```diff
@@ -173,6 +174,85 @@ def test_scaling_instance_map_configured(harness: Harness) -> None:
...
+    with open(config_path, encoding="utf-8") as config_file:
+        content = yaml.safe_load(config_file)
+        assert "instance_map" not in content
+
+
+def test_scaling_stream_writers_configured(harness: Harness) -> None:
+    """
+    arrange: charm deployed, integrated with Redis, two more units in peer relation
+        and set as leader.
+    act: emit config-changed event..
```
cbartz, line 205:
```suggestion
    act: emit config-changed event.
```

cbartz (approved):
LGTM, only have one nitpick

PR #268, opened 2024-03-27
Update NGINX configuration to work with horizontal scaling

amandahla:
Note: Integration test will be done in further story.

### Overview

In Synapse, the main unit and workers handle different endpoints. This PR prepares NGINX to handle it while scaling by using template files that the charm will update with the proper main unit address.

Also, this PR changes Mjolnir to be enabled only in the main unit and redirecting abuse reports properly to it.

#### Flow
- Synapse NGINX Pebble ready event is emitted.
- The charm will call pebble.replan_nginx() with the main unit address.
- The replan_nginx will copy the files main_location.conf.template and abuse_report_location.conf.template to their original places and run the command `sed` to replace the string "main-unit" to the main unit address received.
- Synapse NGINX Pebble layer is replanned.

If the main unit is changed, the relation-changed event is emitted. The same process will happen there.

#### Why all this?
Any unit can correctly forward requests for the main unit if this is the case. Especially considering that only the main unit will run Mjolnir.
Otherwise, the unit itself will handle the request.

### Rationale

Redirect requests as expected while horizontal scaling Synapse.

### Juju Events Changes

NGINX pebble ready event.
_on_relation_changed event.

### Checklist

Charmhub doc will be updated once horizontal scaling is in place.


nginx_rock/etc/main_location.conf.template
```diff
@@ -0,0 +1,8 @@
+proxy_read_timeout 300;
+proxy_pass http://main-unit:8008;
```
merkata, line 2: where is this `main-unit` set, how is it resolved?

nginx_rock/etc/nginx.conf
```diff
@@ -52,27 +52,280 @@ http {
...
     location  / {
-      proxy_read_timeout 300;
-      proxy_pass http://127.0.0.1:8008;
-      proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
-      proxy_set_header X-Forwarded-Proto $http_x_forwarded_proto;
-      proxy_set_header Host $http_host;
-      proxy_set_header X-Real-IP $remote_addr;
-      client_max_body_size 50M;
-      proxy_http_version 1.1;
+      include main_location.conf;
```
merkata, line 328: I'm wondering are these endpoints documented for the workers and if you can combine locations so it is not as long? Is there a reference for this config somewhere?

src/charm.py
```diff
@@ -352,6 +353,37 @@ def _on_relation_changed(self, _: ops.HookEvent, charm_state: CharmState) -> Non
...
+    def _generate_nginx_config(self) -> None:
+        """Generate NGINX configuration based on templates."""
+        container = self.unit.get_container(synapse.SYNAPSE_NGINX_CONTAINER_NAME)
+        container.exec(
+            [
+                "cp",
+                "/etc/nginx/main_location.conf.template",
+                "/etc/nginx/main_location.conf",
+            ],
+        ).wait()
```
merkata, line 365: That's a new one, I think previously we used some internal _exec function (that wraps container.exec) and I'm seeing a wait() for the first time.

src/charm.py
```diff
@@ -361,6 +393,7 @@ def _on_synapse_nginx_pebble_ready(self, _: ops.HookEvent) -> None:
             return
         logger.debug("synapse_nginx_pebble_ready replanning nginx")
         pebble.replan_nginx(container)
+        self._generate_nginx_config()
```
merkata, line 396: Two things - are the lines not in the proper order or I'm missing something - shouldn't it be _generate_nginx_config first and then replan_nginx? Second is aren't these functions better placed in pebble.py?

merkata (commented):
looks good, left some comments

nginx_rock/etc/main_location.conf.template
amandahla: I replace it in here:
https://github.com/canonical/synapse-operator/pull/268/files#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307cR377
This template will be used for generating a new main_location.conf

amandahla: I added a comment now :)

src/charm.py
amandahla: Well, this is an experiment ;-)
I thought about not catching the error if something happens here.

src/charm.py
amandahla: Moved!

src/charm.py
amandahla: Based:
https://discourse.charmhub.io/t/handling-temporary-errors-in-charms/13247

tests/unit/test_charm.py
```diff
@@ -108,12 +108,13 @@ def test_replan_nginx_container_down(harness: Harness) -> None:
...
     arrange: charm deployed.
-    act: start the Synapse charm and set Synapse container to be ready.
+    act: emit config-changed event.
     assert: Synapse charm waits for server_name to be set.
     """
     harness = Harness(SynapseCharm)
+    harness.begin()
 
-    harness.begin_with_initial_hooks()
+    harness.charm.on.config_changed.emit()
```
javierdelapuente, line 117: why this change?

amandahla: This test is not using the harness fixture so calling "begin_with_initial_hooks" would generate: "ops.pebble.APIError: execution handler not found, please register one using Harness.handle_exec"

So I changed to only call the config changed event.

merkata (approved):

src/synapse/workload.py
```diff
@@ -823,3 +823,41 @@ def get_environment(charm_state: CharmState) -> typing.Dict[str, str]:
...
+
+    1. Copy template files as configuration files to be used.
+    2. Run sed command to replace string main-unit in configuration files.
+    3. Reload NGINX.
+
+    Args:
+        container: Container of the charm.
+        main_unit_address: Main unit address to be used in configuration.
+    """
+    container.exec(
```
yanksyoon, line 839: Out of curiosity, may I ask why .template was used instead of jinja? I'm asking because it might be more Pythonic to replace strings than doing sed operations :D

templates/worker.yaml
yanksyoon, line 1: I'm wondering if this file might suit better in a different directory, since it doesn't quite look like a template but rather a configuration. (I might have misunderstood the purpose of this file though!)

tests/unit/conftest.py
```diff
@@ -174,6 +174,19 @@ def start_cmd_handler(argv: list[str]) -> synapse.ExecResult:
         executable="/usr/bin/python3",
         handler=lambda _: synapse.ExecResult(0, "", ""),
     )
+    synapse_nginx_container: ops.Container = harness.model.unit.get_container(
+        synapse.SYNAPSE_NGINX_CONTAINER_NAME
+    )
+    harness.register_command_handler(  # type: ignore # pylint: disable=no-member
```
yanksyoon, line 180: Out of curiosity, may I ask why the pylint disable is applied here? :D

yanksyoon (approved):
LGTM, minor nitpicky comments regarding file structure! Thank you!

src/synapse/workload.py
amandahla: Excellent suggestion! I admit that sed was my first idea so I didn't consider Jinja. I created a story to do that as a refactor: ISD-1745
Thanks!

templates/worker.yaml
amandahla: Good catch. My first approach was to get this file and add it to the unit only if is not a main unit. But then I realized that I only needed to change the command. I think I could even add it to the Rock itself.

tests/unit/conftest.py
amandahla: Sorry, I didnt add the comment but the reason is to ignore this error "error: "Harness[SynapseCharm]" has no attribute "register_command_handler"  [attr-defined]"

templates/worker.yaml
amandahla: It will be addressed in https://github.com/canonical/synapse-operator/issues/275

tests/unit/conftest.py
amandahla: It will be addressed in https://github.com/canonical/synapse-operator/issues/275

PR #290, opened 2024-04-08
fix(irc): Use irc port in app registration

merkata:
### Overview

Changes the port in the app registration file - it has to be the one used by the app and not the one used by Synapse. Also create the file conditionally, we don't want to overwrite it when an event hits and it is already present.

### Module Changes

src/synapse/workload.py change to create the app registration conditionally and use the correct port.

### Library Changes

n/a

### Checklist

- [ x] The changes are compliant with [ISD054 - Manging Charm Complexity](https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619)
- [ x] The documentation is generated using `src-docs`
- [ x] The documentation for charmhub is updated.
- [ x] The PR is tagged with appropriate label (`urgent`, `trivial`, `complex`)

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -16,3 +16,4 @@ jobs:
       modules: '["test_charm", "test_nginx", "test_s3", "test_redis"]'
       self-hosted-runner: true
       self-hosted-runner-label: "edge"
+      tmate-debug: true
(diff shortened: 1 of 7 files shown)


varshigupta12 (dismissed):

arturo-seijas (dismissed):

arturo-seijas (approved):

.github/workflows/integration_test.yaml
```diff
@@ -16,3 +16,4 @@ jobs:
       modules: '["test_charm", "test_nginx", "test_s3", "test_redis"]'
       self-hosted-runner: true
       self-hosted-runner-label: "edge"
+      tmate-debug: true
```
nrobinaubertin, line 19: Shouldn't this be set back to false now ?

amandahla (approved):
LGTM

.github/workflows/integration_test.yaml
merkata: I can remove it in a new PR, I honestly feel this will be needed for a while though :)

PR #296, opened 2024-04-10
Enable libjemalloc

amandahla:
### Overview

The [documentation](https://matrix-org.github.io/synapse/latest/usage/administration/admin_faq.html#help-synapse-is-slow-and-eats-all-my-ramcpu) recommends installing libjemalloc as a improvement of memory resource usage.

We are not checking the file existence since this is not done in any other part of the charm and we own the Rock so we can guarantee that the file is there.

### Rationale

Improve performance.

### Checklist

No need to update charmhub documentation.

### DIFF
--- a/src/synapse/workload.py
+++ b/src/synapse/workload.py
@@ -762,6 +762,7 @@ def get_environment(charm_state: CharmState) -> typing.Dict[str, str]:
         # TLS disabled so the listener is HTTP. HTTPS will be handled by Traefik.
         # TODO verify support to HTTPS backend before changing this  # pylint: disable=fixme
         "SYNAPSE_NO_TLS": str(True),
+        "LD_PRELOAD": "/usr/lib/x86_64-linux-gnu/libjemalloc.so.2",
     }
     datasource = charm_state.datasource
     if datasource is not None:
--- a/src/synapse/workload.py
+++ b/src/synapse/workload.py
@@ -15,7 +15,7 @@ from ops.pebble import ExecError, PathError
 
 from charm_state import CharmState
 
-from .api import SYNAPSE_PORT, SYNAPSE_URL
+from .api import SYNAPSE_URL
 
 SYNAPSE_CONFIG_DIR = "/data"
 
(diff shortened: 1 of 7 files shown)


nrobinaubertin (dismissed):

arturo-seijas (dismissed):

src/synapse/workload.py
```diff
@@ -758,6 +758,7 @@ def get_environment(charm_state: CharmState) -> typing.Dict[str, str]:
         # TLS disabled so the listener is HTTP. HTTPS will be handled by Traefik.
         # TODO verify support to HTTPS backend before changing this  # pylint: disable=fixme
         "SYNAPSE_NO_TLS": str(True),
+        "LD_PRELOAD": "/usr/lib/x86_64-linux-gnu/libjemalloc.so.2",
```
gregory-schiano, line 761: I'm wondering if we shouldn't validate that the file exists before adding it to preload ? WDYT ?

arturo-seijas (dismissed):

src/pebble.py
```diff
@@ -62,6 +62,29 @@ def check_synapse_alive() -> ops.pebble.CheckDict:
...
+        container: Synapse container.
+        pebble_layer: Pebble layer.
+
+    Returns:
+        New pebble layer validated.
+    """
+    if not container.exists(synapse.LIBJEMALLOC_PATH):
+        logger.debug(
+            "%s not found, removing LD_PRELOAD environment variable", synapse.LIBJEMALLOC_PATH
+        )
```
mthaddon, line 83: Do we really want this as a debug level message, rather than (for instance) warning? It seems like it's a pretty big red flag if this isn't installed since we're installing it in the rock, so we'd always expect it to be there?

amandahla: Thanks Niels, it's a good point. As discussed, this check was removed since is not done in any other part of the charm and we own the Rock.

javierdelapuente (approved):

cbartz (approved):
LGTM

PR #299, opened 2024-04-12
Publish feature branch

amandahla:
### Overview

While we don't define how to handle publishing charm in different channel based on spec ISD114, this workflow allows publishing to a temporary branch that can be used to deploy the charm and test it.

### Rationale

Publish and test a new feature without affecting edge and stable versions.

### Checklist

No need to update doc.


arturo-seijas (approved):

amandahla:
> Out of curiosity, would this be for testing on staging? (With github runners we currently scp the files manually... and we can't quite use the edge channels due to deployment promotion reasons) LGTM!

The idea was to testing in staging features that could break the current version but while talking with @javierdelapuente a better approach is to use flags to enable/disable features so we won't have too much trouble reviewing features branches.
Closing this in favor of that :+1:

PR #315, opened 2024-04-23
Add publish_rooms_allowlist config (fix #162)

### LINKED ISSUE #162: Add room_list_publication_rules
### Steps to reproduce

Publish a room on the ubuntu.com homeserver. The server allows this.

### Outcome

Only users on the whitelist are allowed to publish rooms on the ubuntu.com homeserver.
Please add this list of users to room_list_publication_rules in the homeserver.yaml to enable this limitation:

```

room_list_publication_rules:
 - user_id: "@n.buechner:ubuntu.com"
   alias: "*"
   room_id: "*"
   action: allow
 - user_id: "@ilvipero:ubuntu.com"
   alias: "*"
   room_id: "*"
   action: allow
 - user_id: "@merlijn-sebrechts:ubuntu.com"
   alias: "*"
   room_id: "*"
   action: allow
 - user_id: "@arraybolt3:ubuntu.com"
   alias: "*"
   room_id: "*"
   action: allow
 - user_id: "@seth-arnold:ubuntu.com"
   alias: "*"
   room_id: "*"
   action: allow
 - user_id: "@ulfnic:ubuntu.com"
   alias: "*"
   room_id: "*"
   action: allow
 - user_id: "@aaronprisk:ubuntu.com"
   alias: "*"
   room_id: "*"
   action: allow
 - user_id: "@kewisch:ubuntu.com"
   alias: "*"
   room_id: "*"
   action: allow
 - user_id: "@ravage:xentonix.net"
   alias: "*"
   room_id: "*"
   action: allow
 - user_id: "@arraybolt3:matrix.org"
   alias: "*"
   room_id: "*"
 - user_id: "@kohutek:ubuntu.com"
   alias: "*"
   room_id: "*"
 - user_id: "@hggdh2:ubuntu.com"
   alias: "*"
   room_id: "*"
 - user_id: "@gschiano:ubuntu.com"
   alias: "*"
   room_id: "*"
   action: allow
 - user_id: "*"
   alias: "*"
   room_id: "*"
   action: deny

```

### Client information

_No response_

amandahla:
### Overview

The publish_rooms_allowlist configuration allows the user to specify a list of user that will have permission to publish rooms. Once is set, all other users will be forbidden to do that.

Note: why allowlist instead of whitelist? This is a charm configuration so is not related to any existing Synapse configuration that already has the non-inclusive word "whitelist".

Note 1: The "@" symbol is added in the charm because while running an action, Juju considers the symbol "@" in the input as a  reference to local file.

### Rationale

Fix #162

### Checklist

No need to change documentation.


src/charm_state.py
```diff
@@ -173,6 +174,7 @@ class SynapseConfig(BaseModel):  # pylint: disable=too-few-public-methods
     federation_domain_whitelist: str | None = Field(None)
     ip_range_whitelist: str | None = Field(None, regex=r"^[\.:,/\d]+\d+(?:,[:,\d]+)*$")
     public_baseurl: str | None = Field(None)
+    publish_rooms_allowlist: str | None = Field(None, regex=r"(@\w+:\w+\.\w+,?)+")
```
arturo-seijas, line 177: Have yo considered treating this attr as a list here and parsing before setting it here?

amandahla: Excellent suggestion, thanks. Changed! Please, just check if what I did is the same as that you have in mind, I just added a validator to convert the item to list (and check the ids with the regex).

arturo-seijas: I think that's actually better then what I had in mind :)

arturo-seijas (approved):

src/charm_state.py
```diff
@@ -224,6 +227,30 @@ def to_yes_or_no(cls, value: str) -> str:
...
+            The string converted to list.
+
+        Raises:
+            ValidationError: if user_id is not as expected.
+        """
+        if value is None:
+            return []
+        allow_list = [user_id.strip() for user_id in value.split(",")]
+        for user_id in allow_list:
+            if not re.match(r"@\w+:\w+\.\w+", user_id):
```
yanksyoon, line 248: I'm wondering if this regex could be a constant and have a comment explaining what it does help - although this seems simple enough :)
FYI, I've had issues with `w+` pattern with it not allowing dashes, just checking in if you want to allow dashes in user ids :D
Thanks!

src/synapse/workload.py
```diff
@@ -750,6 +750,35 @@ def enable_redis(current_yaml: dict, charm_state: CharmState) -> None:
...
+        for user in publish_rooms_allowlist:
+            rule = {"user_id": user, "alias": "*", "room_id": "*", "action": "allow"}
+            room_list_publication_rules.append(rule)
+
+    if len(room_list_publication_rules) == 0:
+        raise WorkloadError("publish_rooms_allowlist has unexpected value. Please, verify it.")
+
+    last_rule = {"user_id": "*", "alias": "*", "room_id": "*", "action": "deny"}
+    room_list_publication_rules.append(last_rule)
+    current_yaml["room_list_publication_rules"] = room_list_publication_rules
```
yanksyoon, line 779: Minor nitpick: would it be better we keep the inputs immutable and perhaps create a new dict?

yanksyoon (dismissed):
LGTM! 🚀

src/charm_state.py
amandahla: Excellent catch, this helped a lot. After your review, I checked on the spec that actually "-" (and other symbols) are allowed.
Also, instead of creating a constant, I applied the validator for the two configuration items that use it.

src/synapse/workload.py
amandahla: We have many methods that receive the configuration content as input, change it, and then this content is pushed back to the container.
Do you see a downside to doing that in place instead of creating and returning a new dict with the configuration updated?

javierdelapuente (dismissed):

src/synapse/workload.py
yanksyoon: I see - then it might be too much of a change to implement.
I think immutability helps later when the codebase becomes complex - since it will become harder to keep track of where the configurations get changed.

yanksyoon (dismissed):

javierdelapuente (approved):

amandahla:
Losing approvals to make coverage happy :(

config.yaml
```diff
@@ -70,6 +70,12 @@ options:
       Configures whether to report statistics.
     default: false
     type: boolean
+  publish_rooms_allowlist:
+      type: string
+      description: |
+        Comma separated list of users allowed to publish rooms. Once is set, no
```
yanksyoon, line 76:
```suggestion
        Comma separated list of users allowed to publish rooms. Once set, no
```
Minor suggestion :)

yanksyoon (dismissed):
LGTM, thanks for the changes so far!

merkata (approved):

src/synapse/workload.py
```diff
@@ -517,7 +517,7 @@ def _get_irc_bridge_config(charm_state: CharmState, db_connect_string: str) -> t
     config["database"]["connectionString"] = db_connect_string
     if charm_state.synapse_config.irc_bridge_admins:
         config["ircService"]["permissions"] = {}
-        for admin in (a.strip() for a in charm_state.synapse_config.irc_bridge_admins.split(",")):
+        for admin in charm_state.synapse_config.irc_bridge_admins:
             config["ircService"]["permissions"][f"@{admin}"] = "admin"
```
javierdelapuente, line 521: Wasn't the @ already added? (I mean in `[f"@{admin}"] `). Maybe a test could help...

merkata: it is removed as juju interprets it as a pointer to a file, so we add it in the code and not via the action

javierdelapuente: but I believe the `@` was already added in the function userids_to_list. I believe there could be two `@` symbols

amandahla: excellent catch. Fixed.

merkata (approved):

nrobinaubertin (approved):

PR #328, opened 2024-04-29
Refactor NGINX configuration

codethulu:
### Overview

Refactor of `generate_nginx_config` method, to use Jinja2 instead of relying on `.template` files.

### Rationale

Previously, we would configure nginx using `.template` files and executing `sed` in the container. By using Jinja2, we can create a more pythonic and scalable way to edit the config files.

### Juju Events Changes

N/A
### Module Changes

N/A
### Library Changes

Jinja2, currently pinned at version 3.1.2
### Checklist


buse_report_location.conf
```diff
@@ -0,0 +1,11 @@
...
+# https://github.com/matrix-org/mjolnir/blob/main/test/nginx.conf
+# Abuse reports should be sent to Mjolnir.
+# The r0 endpoint is deprecated but still used by many clients.
+# As of this writing, the v3 endpoint is the up-to-date version.
+# Alias the regexps, to ensure that they're not rewritten.
+set $room_id $2;
+set $event_id $3;
+# main-unit will be replace via charm by the main unit address
+# E.g.: synapse-0.model.endpoints
+proxy_pass http://synapse-0.synapse-endpoints:9999/api/1/report/$room_id/$event_id;
```
cbartz, line 11:
```suggestion
proxy_pass http://synapse-0.synapse-endpoints:9999/api/1/report/$room_id/$event_id;

```
final newline is missing

src/synapse/workload.py
```diff
@@ -859,39 +860,28 @@ def get_environment(charm_state: CharmState) -> typing.Dict[str, str]:
...
+        ("main_location.conf.j2", "main_location.conf"),
+        ("abuse_report_location.conf.j2", "buse_report_location.conf"),
+    ]
+
+    for template_name, output_file in templates:
+        template = env.get_template(template_name)
+        output = template.render(main_unit_address=main_unit_address)
+
+        with open(output_file, "w", encoding="utf-8") as f:
+            f.write(output)
```
cbartz, line 887: If you use `Path` objects, you can just do `p.write_text(output, encoding="utf-8")`

cbartz, line 887: I think you are writing these files to the charm container rather than the workload container. You probably need to use something like https://ops.readthedocs.io/en/latest/#ops.Container.push to push the content into the workload container.

buse_report_location.conf
yanksyoon, line 1: I think you meant to name the file `abuse_report_location.conf` :D

yanksyoon (commented):
Minor comments :D

src/synapse/workload.py
codethulu: ah forgot about this, have done so now.

buse_report_location.conf
codethulu: whoops 😅

codethulu: fixed

buse_report_location.conf
```diff
@@ -0,0 +1,11 @@
+# Extracted from
```
amandahla, line 1: Why are these files here?

src/synapse/workload.py
```diff
@@ -870,28 +871,18 @@ def generate_nginx_config(container: ops.Container, main_unit_address: str) -> N
...
-    ).wait()
-    container.exec(
-        [
-            "sed",
-            "-i",
-            f"s/main-unit/{main_unit_address}/g",
-            "/etc/nginx/abuse_report_location.conf",
-        ],
-    ).wait()
+    file_loader = FileSystemLoader(Path("./nginx_rock/etc"), followlinks=True)
```
amandahla, line 874: Since the template files are not used in the rock, wouldn't be better to move to somewhere else?

See the PostgreSQL charm as an example:
https://github.com/canonical/postgresql-k8s-operator/blob/167ce3ae7495b51c741af193863d58e9c1f8ef08/src/patroni.py#L340

codethulu: I actually did have something similar, decided to keep the templates where they were originally. i can move them under a `/templates` directory now

src/synapse/workload.py
```diff
@@ -870,28 +871,18 @@ def generate_nginx_config(container: ops.Container, main_unit_address: str) -> N
...
+
+    # List of templates and their corresponding output files
+    templates = [
+        ("main_location.conf.j2", Path("main_location.conf")),
+        ("abuse_report_location.conf.j2", Path("abuse_report_location.conf")),
+    ]
+
+    for template_name, output_file in templates:
+        template = env.get_template(template_name)
+        output = template.render(main_unit_address=main_unit_address)
```
amandahla, line 885: Is possible that render fails?

src/synapse/workload.py
```diff
@@ -870,28 +871,18 @@ def generate_nginx_config(container: ops.Container, main_unit_address: str) -> N
...
-    container.exec(
-        [
-            "sed",
-            "-i",
-            f"s/main-unit/{main_unit_address}/g",
-            "/etc/nginx/abuse_report_location.conf",
-        ],
-    ).wait()
+    file_loader = FileSystemLoader(Path("./nginx_rock/etc"), followlinks=True)
+    env = Environment(loader=file_loader, autoescape=True)
```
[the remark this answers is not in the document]
codethulu: this is recommended boilerplate for jinja2 to load and render templates, i believe it may also be necessary

src/synapse/workload.py
```diff
@@ -870,28 +871,18 @@ def generate_nginx_config(container: ops.Container, main_unit_address: str) -> N
...
+    # List of templates and their corresponding output files
+    templates = [
+        ("main_location.conf.j2", Path("main_location.conf")),
+        ("abuse_report_location.conf.j2", Path("abuse_report_location.conf")),
+    ]
+
+    for template_name, output_file in templates:
+        template = env.get_template(template_name)
+        output = template.render(main_unit_address=main_unit_address)
+        output_file.write_text(output, encoding="utf-8")
```
amandahla, line 886: I think I'm missing something. Where output_file is used?

src/synapse/workload.py
amandahla: I'm not familiar with Jinja but I believe an exception could happen if main_unit_address is empty or an unexpected type, perhaps?
It seems an edge case but it would be interesting to check just in case.

src/synapse/workload.py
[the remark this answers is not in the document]
amandahla: Ok, the postgresql charm is using only the Template class but I don't have a strong opinion about it.

src/synapse/workload.py
codethulu: have done some investigating, and the method that generates the `main_unit_address`, `get_main_unit_address` will always return a `str` object. 

if this `str` is empty, then all cases of `main_unit_address` in the template file will be replaced by the empty string, which is expected behaviour, if not ideal. This should not happen though, because `get_main_unit_address()` builds a string using variables that should always be set.

I don't think that we need to catch any render exceptions, if an exception occurs it means something has gone horribly wrong elsewhere, and the error should have been caught somewhere more appropriate

codethulu: would appreciate your thoughts!

nginx_rock/etc/main_location.conf
```diff
@@ -1 +1 @@
-return 404;
+return 404;
```
Thanhphan1147, line 1: Should a newline be added here?

nginx_rock/etc/abuse_report_location.conf
```diff
@@ -1 +1 @@
-return 404;
+return 404;
```
Thanhphan1147, line 1: Same for the one below, should a newline be added here?

nginx_rock/etc/main_location.conf
codethulu: whoops, yes. accidentally overwrote these files and tried to revert, missed new line adn the linter did not pick it up. fixed now

src/synapse/workload.py
codethulu: Asked @nrobinaubertin  for input - I don't think that by catching any render errors will be useful for us - we cannot provide any more additional context than Jinja will already have. We are simply passing through a `str`, that even if `""` should still result in a successful render.

merkata (approved):

PR #337, opened 2024-05-02
Fix start.py and Grafana dashboard

amandahla:
### Overview

This PR fixes two issues:
- Remove chown from original start.py script, preventing the refresh of taking too long because /data is a big directory and chown takes a long time to run.
- Fix CPU usage Grafana dashboard. Maybe is related to some prometheus_client upgrade in Synapse but after the refresh the graph was different.

### Rationale

Fix issues.

### Checklist

No need of changing charmhub doc


synapse_rock/start_custom.py
cbartz, line 1: https://github.com/element-hq/synapse/blob/develop/docker/start.py could be updated in the future, what are the plans to keep it up to date?

synapse_rock/rockcraft.yaml
```diff
@@ -129,4 +126,18 @@ parts:
...
-            login_ubuntu.py: /usr/local/attributemaps/login_ubuntu.py
\ No newline at end of file
+            login_ubuntu.py: /usr/local/attributemaps/login_ubuntu.py
+    synapse-start:
+        plugin: dump
+        after:
+          - synapse
+        source: .
+        # start.py is extracted from
+        # https://github.com/matrix-org/synapse/blob/develop/docker/start.py
```
cbartz, line 136: You may want to point to https://github.com/element-hq/synapse/blob/develop/docker/start.py as the other repo is a public archive.

src/grafana_dashboards/synapse.json
cbartz, line 1: Does the dashboard look different after the change? It is difficult to tell from the diff.

synapse_rock/rockcraft.yaml
```diff
@@ -129,4 +126,18 @@ parts:
...
+        # Our version does not use chown command that, depending of the /data
+        # size, could take a long time to run.
+        organize:
+            start_custom.py: start.py
```
javierdelapuente, line 140: Would it be better to "patch" the original one, so we can see when it is changed upstream?

mthaddon:
What's the implication of not doing the chown? Could that cause problems for us later?

src/grafana_dashboards/synapse.json
amandahla: The main difference is migrating from graph (deprecated) to timeseries.

synapse_rock/rockcraft.yaml
amandahla: Do you mean patching in the rockcraft itself? Using sed commands or something similar?

amandahla:
> What's the implication of not doing the chown? Could that cause problems for us later?

I thought that charm storage was mounted with 777 permission but that's not the case. So I'll change the command from 'chown' to 'find . -maxdepth 1 -exec chown {ownership} "{}" \;'
When Synapse starts for the first time creates just this:
```
homeserver.db  homeserver.db-shm  homeserver.db-wal  homeserver.yaml  log.config  lost+found  media  test.macaroon.key  test.registration.key  test.signing.key  worker.yaml
```

So changing the command from chown to find will impact only in the case the user was changed and media directory content still has the old permission.

synapse_rock/rockcraft.yaml
amandahla: Removed ;)

src/grafana_dashboards/synapse.json
amandahla: I'll mark as resolved, feel free to re-open it.

synapse_rock/rockcraft.yaml
amandahla: But then I would need to git clone synapse repository right? Right now the start.py is obtained from what pip installs.

synapse_rock/start_custom.py
amandahla: Javier suggested using a patch so now if for some reason it fails we will know during the rock build.

merkata (approved):
Nice!

javierdelapuente (approved):

javierdelapuente (approved):

merkata (approved):

merkata (dismissed):

synapse_rock/patches/start.patch
```diff
@@ -0,0 +1,18 @@
...
++#!/usr/bin/python3
+ 
+ import codecs
+ import glob
+@@ -136,7 +136,7 @@
+ 
+     if ownership is not None:
+         log(f"Setting ownership on /data to {ownership}")
+-        subprocess.run(["chown", "-R", ownership, "/data"], check=True)
++        subprocess.run(["find", "/data", "-maxdepth", "1", "-exec", "chown",ownership, "{}", ";"], check=True)
```
codethulu, line 14: should we have space after "chown"? small nitpick i know...

codethulu (approved):

amandahla:
"tests/integration/test_s3.py:104:65: E1135: Value 'magic.from_buffer(objbuf)' doesn't support membership test (unsupported-membership-test) 

[22](https://github.com/canonical/synapse-operator/actions/runs/9081358604/job/24955027561?pr=337#step:6:23)"
???

codethulu (approved):

PR #341, opened 2024-05-04
Verify if configuration has changed before restarting Synapse

amandahla:
### Overview

Currently, most events trigger a restart even if the configuration has not changed. This will prevent that by checking the configuration before the restart.

### Rationale

Prevent constant restarts like we saw in Production handling database relation changed in second unit.

### Juju Events Changes

Config change event and integration related. Synapse will only restart if the generated configuration is different from the existing one in the workload.

### Checklist

No need to update CH doc.

### DIFF
--- a/src/pebble.py
+++ b/src/pebble.py
@@ -327,10 +327,11 @@ def change_config(  # noqa: C901 pylint: disable=too-many-branches,too-many-stat
             synapse.add_app_service_config_field(current_synapse_config)
             replan_irc_bridge(container=container)
         config_has_changed = DeepDiff(
-            existing_synapse_config, current_synapse_config, ignore_order=True
+            existing_synapse_config, current_synapse_config, ignore_order=True, ignore_string_case=True
         )
         if config_has_changed:
             logging.info("Configuration has changed, Synapse will be restarted.")
+            logging.debug("The change is: %s", config_has_changed)
             # Push worker configuration
             _push_synapse_config(
                 container,
--- a/src/pebble.py
+++ b/src/pebble.py
@@ -327,7 +327,10 @@ def change_config(  # noqa: C901 pylint: disable=too-many-branches,too-many-stat
             synapse.add_app_service_config_field(current_synapse_config)
             replan_irc_bridge(container=container)
         config_has_changed = DeepDiff(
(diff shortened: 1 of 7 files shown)


src/pebble.py
```diff
@@ -276,6 +277,7 @@ def change_config(  # noqa: C901 pylint: disable=too-many-branches,too-many-stat
     """
     try:
         synapse.execute_migrate_config(container=container, charm_state=charm_state)
+        existing_synapse_config = _get_synapse_config(container)
         current_synapse_config = _get_synapse_config(container)
```
javierdelapuente, line 281: isn't `existing_synapse_config` equal to `current_synapse_config`?, they are just one line apart

javierdelapuente (commented):
Maybe a unit test could be added for the change?

src/pebble.py
mthaddon, line 281: I'm not sure I understand how these two values would ever be different - they're running the same things?

mthaddon: Ok, I see how this works, you're passing current_synapse_config to enable_* methods below, which may modify it.

src/pebble.py
```diff
@@ -324,16 +326,27 @@ def change_config(  # noqa: C901 pylint: disable=too-many-branches,too-many-stat
...
+                container,
+                get_worker_config(unit_number),
+                config_path=synapse.SYNAPSE_WORKER_CONFIG_PATH,
+            )
+            # Push main configuration
+            _push_synapse_config(container, current_synapse_config)
+            synapse.validate_config(container=container)
+            restart_synapse(container=container, charm_state=charm_state, is_main=is_main)
+        else:
+            logging.info("Configuration has not changed, no action.")
```
mthaddon, line 349: I think it'd be good to have a test that confirms this behaviour. Confirm that if a config-changed hook fires but configuration doesn't change the service isn't restarted, and vice versa.

javierdelapuente (approved):
LGTM! tests for the next PR

src/pebble.py
amandahla: Nice, I'll add one in further PR.

amandahla:
I'll add a unit test in further PR.

PR #349, opened 2024-05-13
Copy signing keys from main to workers

amandahla:
### Attention
To be merged after IRC PRs.

### Overview

Copy signing keys from main unit to worker since is required to be the same for prevent problems while sending signed events.

Also, check if Redis configuration has changed before restarting Synapse. Redis charm sends the same custom event for relation created and joined.

### Rationale

Every unit will sign events with the same key.

### Checklist

No doc is changed in CH.


src/charm.py
```diff
@@ -361,6 +374,53 @@ def set_main_unit(self, unit: str) -> None:
...
+
+    def get_signing_key(self) -> typing.Optional[str]:
+        """Get signing key from secret.
+
+        Returns:
+            Signing key as string or None if not found.
+        """
+        peer_relation = self.model.relations[synapse.SYNAPSE_PEER_RELATION_NAME]
+        if not peer_relation:
+            logger.error(
```
arturo-seijas, line 406: I wouldn't log this if only a single instance is deployed. Also, shouldn't an exception be raised when there're no peer relations but there're multiple units?

src/charm.py
```diff
@@ -361,6 +374,53 @@ def set_main_unit(self, unit: str) -> None:
...
+        secret_id = peer_relation[0].data[self.app].get("secret-signing-id")
+        if secret_id:
+            try:
+                secret = self.model.get_secret(id=secret_id)
+                logging.debug(secret.get_content().get("secret-signing-key"))
+                return secret.get_content().get("secret-signing-key")
+            except (ops.model.SecretNotFoundError, ValueError, TypeError) as exc:
+                logger.exception("Failed to get secret id %s: %s", secret_id, str(exc))
+                del peer_relation[0].data[self.app]["secret-signing-id"]
+                return None
```
arturo-seijas, line 421: This `return` doesn't do anything as far as I can see

src/charm.py
```diff
@@ -361,6 +374,53 @@ def set_main_unit(self, unit: str) -> None:
...
+        return None
```
arturo-seijas, line 422: Same here

src/charm.py
arturo-seijas, line 406: I wouldn't log this if only a single instance is deployed. Also, shouldn't an exception be raised when there're no peer relations but there're multiple units?

src/charm.py
arturo-seijas, line 421: This `return` doesn't do anything as far as I can see

src/charm.py
arturo-seijas, line 422: Same here

src/charm.py
```diff
@@ -200,10 +200,23 @@ def change_config(self, charm_state: CharmState) -> None:
             return
         self.model.unit.status = ops.MaintenanceStatus("Configuring Synapse")
         try:
+            signing_key_path = f"/data/{charm_state.synapse_config.server_name}.signing.key"
```
arturo-seijas, line 203: Is this needed for a single unit deployment?

src/charm.py
amandahla: A synapse_peers_relation_created event is emitted after the start and the relation synapse-peers is created even for 1 unit. This check is more for a really exceptional case (considering peers) and also to make lint happy.

Currently, we return None if a relation is not found in the charm but this could be a subject for a refactoring and also a point to the Contributing guide.

src/charm.py
amandahla: Removed

src/charm.py
amandahla: If I remove this one I get a "error: Missing return statement".

src/charm.py
amandahla: No, if there is just one unit it wont get the signing keys from a secret but it will be necessary to copy the existing signing key to a secret in case of a scaling.

src/charm.py
arturo-seijas: Gotcha

arturo-seijas: It actually is

src/charm.py
arturo-seijas: Weird, in any case, you don't need to return `None` here and in l410, right?

amandahla: The only return that I could remove without pylint complaining about it is the one inside the exception :|

arturo-seijas: What if you remove the `None`while keeping the `return`?

amandahla: src/charm.py:409: error: Return value expected  [return-value]
src/charm.py:420: error: Return value expected  [return-value]

arturo-seijas: For some reason, I though the method signature return value was None, sorry about that :/

arturo-seijas (approved):

src/database_observer.py
```diff
@@ -58,15 +61,32 @@ def _change_config(self, charm_state: CharmState) -> None:
         Args:
             charm_state: Instance of CharmState
         """
+        if self._charm.get_main_unit() is None and self._charm.unit.is_leader():
```
javierdelapuente, line 64: As far as I see, the method database_observer._change_config is the same as the charm.change_config. Why not calling the charm method instead?  (maybe add that method to the base class of the charm, so no casting is needed)

amandahla: I believe this will be better handled/discussed in ISD-1820. What do you think?

amandahla: But...this is related to the last commit that I pushed regarding is_main check. I'll wait for the integration tests to run and see the behavior but I think it would be  easy to do your suggestion in all observers and just call the charm.change_config method.

amandahla: Proposal: Merge this one (after reviews) and then work on the reconcile/restarts in the ISD-1820.

amandahla: Reconcile: https://github.com/canonical/synapse-operator/pull/378

javierdelapuente (approved):

arturo-seijas (approved):

PR #375, opened 2024-06-03
Skip Mjolnir configuration if not main unit

amandahla:
### Overview

Small fix for skip Mjolnir configuration if not main unit.

### Rationale

Mjolnir should be started only in the main unit.

### Checklist

No need to change CH docs.


src/mjolnir.py
```diff
@@ -87,9 +87,12 @@ def _on_collect_status(  # noqa: C901
             logger.debug(
                 "Peer relation found, checking if is main unit before configuring Mjolnir"
             )
-            main_unit_id = peer_relation[0].data[self._charm.app].get("main_unit_id")
+            main_unit_id = (
+                peer_relation[0].data[self._charm.app].get("main_unit_id", self._charm.unit.name)
+            )
             if not self._charm.unit.name == main_unit_id:
                 logger.info("This is not the main unit, skipping Mjolnir configuration")
+                return
```
arturo-seijas, line 95: Do you think an unit test might be worth for this?

amandahla: Done

src/mjolnir.py
```diff
@@ -87,9 +87,12 @@ def _on_collect_status(  # noqa: C901
             logger.debug(
                 "Peer relation found, checking if is main unit before configuring Mjolnir"
             )
-            main_unit_id = peer_relation[0].data[self._charm.app].get("main_unit_id")
+            main_unit_id = (
+                peer_relation[0].data[self._charm.app].get("main_unit_id", self._charm.unit.name)
```
cbartz, line 91: Why do we use the default value `self._charm.unit.name` in `.get("main_unit_id", self._charm.unit.name)`, i.e. consider this unit as the main unit if it is not set?

cbartz (commented):
Left a question

src/mjolnir.py
amandahla: If for some reason the main unit is not found/not set, we assume the the unit is the main.
But actually I came up with this to make existing unit tests pass. Harness does not create a peer relation for a charm while starting it with begin(), so you need to create one if you want to test peer relation behavior or start harness with begin_with_initial_hooks.
See: https://github.com/canonical/operator/issues/547

cbartz: Ok, and you don't see any cases where this could apply to two units in parallel (so that Mjolnir is started in two units if both can't find the main unit)? Maybe add a short comment explaining this at the top of the line.

cbartz (approved):

src/mjolnir.py
amandahla: Sure, to not lose the approvals, I'll add a comment on this PR https://github.com/canonical/synapse-operator/pull/378

amandahla: Done [here](https://github.com/canonical/synapse-operator/pull/378/commits/23c8723d25f8e5e4b57dc8b5dd7860cf847b9344#diff-646a0a89a8d0a57307ba77bbcaab1399f19f8d62d31241a61e37b0ed3fbe63fb)

PR #378, opened 2024-06-04
Reconcile Synapse configuration avoiding restarts from different sources

amandahla:
### Overview

Synapse does not provide a hot reload so for every configuration change, a restart is needed.

Because of this, especially while working on horizontal scaling, we saw the Synapse application being restarted multiple times.

You can see all the steps to improve this in the commit order.
TL;DR All observers and charm events call a single method named reconcile that will be responsible for all changes and restart.

### Rationale

Improve stability.

### Checklist


src/pebble.py
```diff
@@ -125,16 +125,16 @@ def check_irc_bridge_ready() -> ops.pebble.CheckDict:
...
+    """Restart Synapse NGINX service and regenerate configuration.
 
     Args:
         container: Charm container.
         main_unit_address: Main unit address to be used in configuration.
     """
     container.add_layer("synapse-nginx", _nginx_pebble_layer(), combine=True)
     synapse.generate_nginx_config(container=container, main_unit_address=main_unit_address)
-    container.replan()
+    container.restart(synapse.SYNAPSE_NGINX_SERVICE_NAME)
```
arturo-seijas, line 137: Won't this trigger more restarts and cause more downtime that a replan?

tests/unit/test_charm.py
```diff
@@ -170,52 +169,6 @@ def test_saml_integration_container_down(saml_configured: Harness) -> None:
...
-    enable_saml_mock = MagicMock()
-    monkeypatch.setattr(synapse, "enable_saml", enable_saml_mock)
-
-    charm_state = harness.charm.build_charm_state()
-    pebble.enable_saml(charm_state, container=container)
-
-    enable_saml_mock.assert_called_once()
-
-
-def test_saml_integration_pebble_error(
```
arturo-seijas, line 197: Why isn't this scenario tested anymore?

tests/unit/test_charm.py
```diff
@@ -233,47 +186,26 @@ def test_smtp_integration_container_down(smtp_configured: Harness) -> None:
...
     harness.begin()
 
     relation = harness.charm.framework.model.get_relation("smtp", 0)
     harness.charm._smtp.smtp.on.smtp_data_available.emit(relation)
 
     enable_smtp_mock.assert_called_once()
-    assert isinstance(harness.model.unit.status, ops.ActiveStatus)
-
-
-def test_smtp_relation_pebble_error(smtp_configured: Harness, monkeypatch: pytest.MonkeyPatch):
```
arturo-seijas, line 260: Why isn't this scenario tested anymore?

tests/unit/test_charm.py
```diff
@@ -450,58 +382,38 @@ def test_nginx_replan_with_synapse_service_not_existing(
...
+    )
     harness.begin()
 
     harness.charm.on.redis_relation_updated.emit()
 
     enable_redis_mock.assert_called_once()
-    assert isinstance(harness.model.unit.status, ops.ActiveStatus)
-
-
-def test_redis_relation_pebble_error(redis_configured: Harness, monkeypatch: pytest.MonkeyPatch):
```
arturo-seijas, line 489: Why isn't this scenario tested anymore?

src/charm.py
```diff
@@ -185,7 +185,7 @@ def instance_map(self) -> typing.Optional[typing.Dict]:
         logger.debug("instance_map is: %s", str(instance_map))
         return instance_map
 
-    def change_config(self, charm_state: CharmState) -> None:
+    def reconcile(self, charm_state: CharmState) -> None:
         """Change configuration.
```
merkata, line 189: I would expand the description here that this is the main entry for changes that require a restart, good to have it documented and serve as guidance (and reminder) for future additions/changes to the code base.

src/charm.py
```diff
@@ -442,15 +449,16 @@ def _on_relation_changed(self, _: ops.HookEvent, charm_state: CharmState) -> Non
...
+            self.reconcile(charm_state)
         # Reload NGINX configuration with new main address
         nginx_container = self.unit.get_container(synapse.SYNAPSE_NGINX_CONTAINER_NAME)
         if not nginx_container.can_connect():
             logger.warning(
                 "Relation changed received but NGINX container is not available for reloading."
             )
             return
-        pebble.replan_nginx(nginx_container, self.get_main_unit_address())
+        pebble.restart_nginx(nginx_container, self.get_main_unit_address())
```
merkata, line 461: We do a restart now instead of a replan, can you say why?

src/synapse/workload.py
```diff
@@ -400,7 +400,10 @@ def enable_stream_writers(current_yaml: dict, charm_state: CharmState) -> None:
     persisters = []
     if charm_state.instance_map_config is not None:
         persisters = [key for key in charm_state.instance_map_config.keys() if key != "main"]
-    current_yaml["stream_writers"] = {"events": persisters}
+    if persisters:
```
merkata, line 403: Nit: just for consistency sake, can the check be against `None`?

merkata (dismissed):
Looks great!
`+245 -669` :clap:

src/pebble.py
amandahla: Using the replan doesn't get changes in the configuration file and when scaling, the file changes. I noticed that while deploying Synapse with two units and one of the units didn't have the right configuration.
It would be nice to have something like container.reload or we can work in a further PR to reload NGINX using exec.
WDYT?

src/charm.py
amandahla: We can discuss in the same thread here: https://github.com/canonical/synapse-operator/pull/378#discussion_r1629410433 :)

tests/unit/test_charm.py
amandahla: The saml observer called directly the pebble.enable_saml. Since this is not done anymore, this test was removed.

src/pebble.py
arturo-seijas: Oh, makes sense. I think that if this is good enough I wouldn't complicate it further. But those are definitely good options

tests/unit/test_charm.py
amandahla: Same thing than SAML :)
The smtp also called directly the pebble.enable_smtp, not anymore.

tests/unit/test_charm.py
arturo-seijas: And is there a pebble error test scenario for `enable_saml`?

tests/unit/test_charm.py
amandahla: Same as other observers, they are not interacting with pebble.enable_<observer> anymore. Only reconcile.

tests/unit/test_charm.py
amandahla: Good catch, you have a good eye for missing tests!
We don't have any to check how the charm behaves if reconcile raises a PebbleServiceError. I'll add one.

src/charm.py
amandahla: sure,done.

tests/unit/test_charm.py
amandahla: Done: test_saml_enabled_reconcile_pebble_error, test_smtp_enabled_reconcile_pebble_error and test_redis_enabled_reconcile_pebble_error

arturo-seijas (approved):

src/charm.py
```diff
@@ -390,9 +395,12 @@ def set_signing_key(self, signing_key: str) -> None:
         if signing_key == self.get_signing_key():
             logger.info("Received signing key but there is no change, skipping")
             return
-        logger.debug("Adding signing key to secret: %s", signing_key)
-        secret = self.app.add_secret({"secret-signing-key": signing_key})
-        peer_relation[0].data[self.app].update({"secret-signing-id": typing.cast(str, secret.id)})
+        if self.unit.is_leader():
```
javierdelapuente, line 398: I understand it is quite bad if we are here and the unit is not a leader

javierdelapuente (approved):
Nice PR!. That code in those observers was not very useful...

src/charm.py
amandahla: Only leader can write data on the application peer relation data bag so it would fail anyway :|
The pitfall here is the main unit is not the leader and the secret is not there (kind of edge case but anyway, to be considered).

amandahla: I'll create a story for that, just in case to not forget about it.

src/charm.py
amandahla: I'll close this one considering the approvals (and my anxiety to test it). Please, feel free to reach out for further discussion regarding that.

PR #388, opened 2024-06-10
Reconcile Synapse configuration after every peer relation-changed event

amandahla:
### Overview

We are relying on relation.units to set the instance_map configuration. This property is only properly set with all units after the relation-joined event.

Since the charm already observers relation-changed event and this event is emitted after the relation-joined, this PR only remove the "if" that was preventing the configuration in the main unit to be reconciled.

Also:
- Sort stream_writers since this is needed for replication works.

### Rationale

Fix how instance_map is set.

### Checklist

CH horizontal scaling documentation will be created in different PR/story.


arturo-seijas (dismissed):

src/charm.py
```diff
@@ -449,10 +449,10 @@ def _on_relation_changed(self, _: ops.HookEvent, charm_state: CharmState) -> Non
         Args:
             charm_state: The charm state.
         """
-        # the main unit has changed so workers must be restarted
-        if self.get_main_unit() != self.unit.name:
-            logger.debug("_on_relation_changed emitting reconcile")
-            self.reconcile(charm_state)
+        # At this point, instance_map will configured as expected since relation.units
+        # is set after relation joined. The relation changed event is always emitted after.
```
cbartz, line 453: I think `relation-changed` can also occur before `relation-joined` , so I wonder if the assumption in this comment is correct.

See also https://juju.is/docs/sdk/charm-lifecycle :
>A X-relation-joined event is always followed up (immediately after) by a X-relation-changed event. But any number of [*-relation-changed](https://juju.is/docs/sdk/relations#heading--relation-events) events can be fired at any time during operation, and they need not be preceded by a [*-relation-joined](https://juju.is/docs/sdk/relations#heading--relation-events) event.

cbartz (commented):
A comment on the order of events.

src/charm.py
amandahla: The comment is not clear but my idea was to justify why this is being done in the relation-changed handler instead of creating one for the relation-joined. I'll rephrase it.

cbartz (approved):

merkata (approved):

PR #404, opened 2024-06-20
Add alive_check configuration

amandahla:
**Update:** I changed the check from ready to alive because the alive one behaves as expected by restarting the pod once is failing.
From pebble documentation, I assumed that:
- Setting only the ready check would mean that if fails, Pebble would show alive as failed too.
- Once ready check fails, a restart action action would be triggered.
https://github.com/canonical/pebble#layer-specification

Thanks @javierdelapuente for raising this.
Pebble issue: https://github.com/canonical/pebble/issues/439

---

### Overview

Since there is no way of delaying health checks before Synapse start, we need an easy way of changing timeout/threshold/retry.

Reference:
https://github.com/canonical/pebble/issues/145

Note:
The test_synapse_pebble_layer_change is skipped because it seems that the checks are not replaced.
Real environments: :heavy_check_mark: 
Harness: https://github.com/canonical/operator/issues/1112

### Rationale

This way we can easily change checks if Synapse is being impacted (constantly restarting)

### Checklist

No CH doc.


src/pebble.py
```diff
@@ -38,31 +41,25 @@ def __init__(self, msg: str):
         self.msg = msg
 
 
-def check_synapse_ready() -> ops.pebble.CheckDict:
+def check_synapse_ready(charm_state: CharmState) -> ops.pebble.CheckDict:
```
Thanhphan1147, line 44: do you think this would need the entire `charm_state` instance as parameter? can we only pass the `ready` config?

Thanhphan1147 (dismissed):
LGTM, only a minor comment.

src/charm_state.py
```diff
@@ -266,6 +268,35 @@ def userids_to_list(cls, value: str) -> typing.List[str]:
...
+
+        Returns:
+            The pebble check.
+
+        Raises:
+            ValidationError: if ready_check is invalid.
+        """
+        # expected
+        # period;threshold;timeout
+        config_values = value.split(",")
```
dimaqq, line 287: Are the values delimited by semicolons or commas?

src/charm_state.py
```diff
@@ -266,6 +268,35 @@ def userids_to_list(cls, value: str) -> typing.List[str]:
...
+        if len(config_values) < 3:
+            raise ValidationError(f"Invalid ready_check, less than 3 values: {value}", cls)
```
dimaqq, line 289: What are the semantics for having more than three?

cbartz, line 289:
```suggestion
        if len(config_values) != 3:
            raise ValidationError(f"Invalid ready_check, less or more than 3 values: {value}", cls)
```

cbartz (commented):
Looks good, minor comments

config.yaml
```diff
@@ -72,6 +72,11 @@ options:
       The public-facing base URL that clients use to access this Homeserver.
       Defaults to https://<server_name>/. Only used if there is integration with
       SAML integrator charm.
+  ready_check:
+    type: string
+    description: Comma separated list of period,threshold and timeout for Synapse
+      pebble ready check.
+    default: "2m,5,20s"
```
yanksyoon, line 79: Just out of curiosity, what are the time units that can be specified (s/m/h)? I'm wondering if it might help to add them to the description as well :D

src/charm_state.py
```diff
@@ -266,6 +268,35 @@ def userids_to_list(cls, value: str) -> typing.List[str]:
...
+        try:
+            period = config_values[0]
+            threshold = int(config_values[1])
+            timeout = config_values[2]
+            return {"period": period, "threshold": threshold, "timeout": timeout}
```
yanksyoon, line 294: Just out of curiosity, i'm wondering if it would help to check the time units here as well (`values[-1] in ("s", "m", "h")`) :)
One minor thing I also like to do is to `str.trim()` them just incase the user does `config1, config2, config3` (spaces inbetween for aesthetics)

yanksyoon (approved):
LGTM, minor suggestions to for aesthetics :D Thank you!

src/charm_state.py
amandahla: Commas, fixed.

src/charm_state.py
amandahla: Also fixed. Changed to check for != 3
Thanks @cbartz :)

config.yaml
amandahla: It's not specified but it should follow the same rules defined by pebble layer. I'll add a link
https://github.com/canonical/pebble/#layer-specification

src/charm_state.py
amandahla: Curious, I thought that I added...I'll add it now, thanks!

src/pebble.py
```diff
@@ -38,31 +41,25 @@ def __init__(self, msg: str):
...
-    Returns:
-        Dict: check object converted to its dict representation.
-    """
-    check = Check(synapse.CHECK_ALIVE_NAME)
-    check.override = "replace"
-    check.level = "alive"
-    check.tcp = {"port": synapse.SYNAPSE_PORT}
+    ready_check = charm_state.synapse_config.ready_check
+    if ready_check:
+        # This will tolerate failure for 8 minutes before restarting Synapse
```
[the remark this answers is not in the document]
amandahla: Yes, good catch, it's ~10minutes before restarting!

src/pebble.py
amandahla: The charm is not using this pattern yet so I follow the existing one, at least until we have a refactor in place to change that.
Also, I could check if is None or not in check_synapse_ready

src/charm_state.py
```diff
@@ -266,6 +268,43 @@ def userids_to_list(cls, value: str) -> typing.List[str]:
...
+            The pebble check.
+
+        Raises:
+            ValidationError: if ready_check is invalid.
+        """
+        # expected
+        # period,threshold,timeout
+        config_values = value.split(",")
+        if len(config_values) != 3:
+            raise ValidationError(f"Invalid ready_check, less or more than 3 values: {value}", cls)
```
dimaqq, line 289: 🤔 a trade-off to consider:

a simpler way to write this would be

`foo, bar, baz = value.split(",")`

Which raises a ValueError on wrong number of commas, and pydantic.ValidationError has ValueError as a base class.

Ofc, that would not provide a neat, user-readable exception...

yanksyoon (approved):

config.yaml
```diff
@@ -72,6 +72,12 @@ options:
       The public-facing base URL that clients use to access this Homeserver.
       Defaults to https://<server_name>/. Only used if there is integration with
       SAML integrator charm.
+  ready_check:
```
mthaddon, line 75: I think we should be careful about adding a config option with this name. To me this is a temporary workaround and something we'd want to remove at some point. Having a "experimental" prefix or something like that, as well as a note about it in the description would make this clearer.

src/charm_state.py
amandahla: Thanks, I'll leave it in favor of the user-readable exception. This configuration is a workaround until https://github.com/canonical/pebble/issues/145 is in place

config.yaml
amandahla: Good catch, renamed.

cbartz (dismissed):

src/charm_state.py
```diff
@@ -266,6 +268,45 @@ def userids_to_list(cls, value: str) -> typing.List[str]:
...
+            )
+        try:
+            period = config_values[0].strip().lower()
+            if period[-1] not in ("s", "m", "h"):
+                raise ValidationError(
+                    f"Invalid experimental_alive_check, period should end in s/m/h: {value}", cls
+                )
+            threshold = int(config_values[1].strip())
+            timeout = config_values[2].strip().lower()
+            if timeout[-1] not in ("s", "m", "h"):
```
dimaqq, line 300: Is there a default for the units, e.g. `15` meaning `15m` or `15s`?

cc @benhoyt as I can't make it out from documentation.

dimaqq (approved):
LGTM code-wise.

dimaqq:
Some high level thoughts: 

Does the heath check interval really need configuration? As in, who is in the better position to understand synapse/mjolnir behaviour, the charm author or the operator? Who would know what values to apply best?

The code to parse this custom format somehow bothers me... because the stringified `10a,3,30s`  format is like invented for this charm only. Would it be easier to supply each value separately and rely on defaults? Or would it be more pragmatic to parse this string with a regexp?

Given that the feature is marked experimental, I figure it's fine.

tests/unit/test_charm.py
```diff
@@ -51,6 +53,29 @@ def test_synapse_pebble_layer(harness: Harness) -> None:
...
+    harness.begin_with_initial_hooks()
+    pebble_plan = harness.get_container_pebble_plan(synapse.SYNAPSE_CONTAINER_NAME).to_dict()
+    assert pebble_plan["checks"]["synapse-ready"]["period"] == "2m"
+    assert pebble_plan["checks"]["synapse-ready"]["threshold"] == 5
+    assert pebble_plan["checks"]["synapse-ready"]["timeout"] == "20s"
+
+    harness.update_config({"experimental_alive_check": "1m,3,30s"})
+
+    pebble_plan = harness.get_container_pebble_plan(synapse.SYNAPSE_CONTAINER_NAME).to_dict()
+    assert pebble_plan["checks"]["synapse-ready"]["period"] == "1m"
```
javierdelapuente, line 74: should not this one be `synapse-alive`?

javierdelapuente: I see the test is skipped for now

src/charm_state.py
benhoyt: No, there's no default -- they require a unit suffix for clarity. The format of these intervals is Go's `time.Duration` format, documented briefly here: https://pkg.go.dev/time#ParseDuration

Yeah, we should clear that up in our docs. Docs are in flux right now with Tiexin moving them to a docs site, but feel free to open an issue.

tests/unit/test_charm.py
amandahla: At least until this one is in place :)
https://github.com/canonical/operator/pull/1268

javierdelapuente (approved):
LGTM (pending the updated test with the ops library is updated)

yanksyoon (approved):

PR #440, opened 2024-07-29
fix: set all units to use the same Prometheus exporter port

amandahla:
### Overview

I wasn't in my right mind when I set the main and workers to have different Prometheus exporter ports...

Since there is no conflict when using the same (a unit can be exclusively a main or a worker), this PR removes all logic regarding that difference and also adds an integration test to guarantee that all units have metrics enabled and listening on the same port (9000).

### Rationale

Collect metrics from all units.

### Checklist


src/synapse/workload.py
```diff
@@ -849,7 +848,7 @@ def generate_worker_config(unit_number: str) -> dict:
             {
                 "type": "metrics",
                 "bind_addresses": ["::"],
-                "port": PROMETHEUS_WORKER_TARGET_PORT,
+                "port": SYNAPSE_EXPORTER_PORT,
```
yanksyoon, line 851: Does this need to be cast as `int` too?

yanksyoon (approved):
LGTM!

src/synapse/workload.py
amandahla: good catch, thanks! fixed :)

yanksyoon (approved):

arturo-seijas (approved):

PR #476, opened 2024-08-19
feat: add federation_sender_instances

amandahla:
### Overview

Synapse configurations allows you to handle outbound federation requests in a different process using a generic_worker.

This PR does that by:
- Changing main replication listener from 8034 to 8035 port
- Starting a generic worker together with the main unit that will be responsible exclusively  for outbound federation requests
- The generic worker will listen to the 8034 port (that's why the replication port change)

### Rationale

Although will be running in the same pod, extracting this process from main will at least and hopefully improve startup time in case of users joining big federated rooms.

Note: This PR does not consider scaling federation sender workers since they have same design as the stream_writers, requiring restart once they change. By adding it together with main, we avoid this complexity for now.
See [Synapse documentation](https://matrix-org.github.io/synapse/latest/usage/configuration/config_documentation.html#federation_sender_instances) for more details about it.

### Checklist

CH documentation will be updated later.


nrobinaubertin (dismissed):
LGTM

gregory-schiano (approved):
That's ok for now, but we'll have to think about the next step of workers in the future

javierdelapuente (approved):

PR #544, opened 2024-10-04
docs: change requirements in tutorial

amandahla:
### Overview

Change requirements in tutorial.

### Rationale

Better description.

### Checklist


docs/tutorial/getting-started.md
```diff
@@ -11,10 +11,10 @@ Through the process, you'll verify the workload state, and log in to your
 Synapse instance via Element Desktop application.
 
 ## Requirements
-- Juju 3 installed.
-- Juju controller created.
 
-For more information about how to install Juju, see [Get started with Juju](https://juju.is/docs/olm/get-started-with-juju).
+* A working station, e.g., a laptop, with amd64 architecture.
+* Juju 3 installed and bootstrapped to a MicroK8s controller. You can accomplish
+this process by following this guide: [Set up / Tear down your test environment](https://juju.is/docs/juju/set-up--tear-down-your-test-environment)
```
erinecon, line 17: If the reader uses a Multipass VM, will they still be able to visit http://tutorial-synapse.juju.local/ from a browser? Or will other parts of the tutorial need to be updated?

amandahla: No, in that case the user would need to point to the Multipass VM instance IP. Also we would need to add steps mentioning how to setup Multipass and that is a requirement for the tutorial.
Similar to:
https://github.com/canonical/postgresql-k8s-operator/blob/main/docs/tutorial/t-set-up.md

erinecon: Multipass is set up if the reader follows the instructions in [Set up / Tear down your test environment](https://juju.is/docs/juju/set-up--tear-down-your-test-environment). I don't think we need to include steps about how to set up Multipass because that guide already contains the instructions.

I do think the requirements could be clearer, i.e., instead of "Juju 3 installed and bootstrapped to a MicroK8s controller" the line could read "A Multipass VM with Juju 3 installed and bootstrapped to a MicroK8s controller."

But if we don't want to force the reader to use a Multipass VM, then we could update the requirement to say "Juju 3 installed and bootstrapped to a MicroK8s controller. You can accomplish this process by using a Multipass VM as outlined in this guide: [Set up / Tear down your test environment](https://juju.is/docs/juju/set-up--tear-down-your-test-environment)"

amandahla: Nice, I added that and a warning about replacing IP since the steps now consider a local installation.

erinecon: Cool, I think that will work :D
One nitpick in the warning you added: I think it might be good to include information about how the user can check the VM IP. Something like, "To get the IP address of the Multipass instance run `multipass info my-juju-vm`."

amandahla: Good point, added.

erinecon (approved):

alithethird (approved):

amandahla:
This PR will get new mjolnir snap version, let's see.

PR #550, opened 2024-10-08
Feat/add plugin provider

merkata:
### Overview

Adds a basic matrix authentication provider for usage of plugins.

### Rationale

Plugins are decoupled from matrix and need to authenticate and register over a library.

### Juju Events Changes

Observes "matrix-auth" relation events.

### Module Changes

Adds a plugins.py.

### Library Changes

Adds a matrix_auth library.

### Checklist


gregory-schiano:
The overall structure should be reviewed:
- We're missing a listener on the relation departed event, this departed event should be listened by the library which will emit a custom event
- Instead of listening to the relation joined event, I'd rather add some logic in the charm reconcile method, that checks if the conditions are met to inject the shared secret in the relation databag, if so the information is injected in all the "Matrix Auth" relation available (through a helper method in the ObserverClass)
- Instead of manipulating the homeserver.yaml in the MatrixAuthObserver, it should follow the same pattern as Redis, Media, ... meaning provide a from_relation_data_config, which is used to inject the required information in the charm_state
- A listener on the custom departed event will trigger a cleanup of the file
- Then Inside pebble & workload_config we can add the required method to add the homeserver attributes we need
- To call the pebble / workload, the Observer hooks will call the charm reconcile like Redis is doing

PR #595, opened 2024-11-18
ISD-2736 Make public_baseurl required and render it in homeserver config

Thanhphan1147:
Synapse is currently configured to generate and serve the .well-known/matrix/client endpoint 

However, this endpoint is currently broken as synapse does not add the public_baseurl property to the homeserver configuration by default. This causes this error message when calling the well-known endpoint: 
```
root@synapse-0:/# curl localhost:8080/.well-known/matrix/client`
{"errcode":"M_NOT_FOUND","error":".well-known not available"}root@synapse-0:/#
```
With the public_baseurl attribute we can reach the endpoint (related issue: 

 ```
root@synapse-0:/# curl localhost:8080/.well-known/matrix/client
{"m.homeserver":{"base_url":"http://synapse.internal/"}}root@synapse-0:/# 
```

The public_baseurl is currently a config option that the operator can set which will add the attribute to the homeserver config. As this attribute is critical for MAS since its value is reflected to the baseurl attribute of MAS, as well as some attributes in the experimental msc3861 configuration. 

This PR makes it a required attribute in charmstate, as well as render it everytime the homeserver config is refreshed.

### Checklist

### CODE REVIEW

File: src/charm_state.py [hunk 1]
```diff
@@ -413,9 +414,21 @@ def from_charm(  # pylint: disable=too-many-arguments,too-many-positional-argume
             CharmConfigInvalidError: if the charm configuration is invalid.
         """
         try:
+            # Compute public_baseurl if not configured
+            config = dict(charm.config.items())
```
weiiwang01, line 418:
```suggestion
            config = dict(charm.config)
```

The `dict` function automatically accepts mappings, so it doesn't require `.items()`. https://docs.python.org/3/library/stdtypes.html#dict

weiiwang01 (approved):

amandahla (approved):
Approving, but we still need to review how to properly set a charm configuration where one of the default values depends on a value obtained via integration (ingress).

Thanhphan1147:
Thanks @amandahla, I'll merge this PR and create a ticket to track the refactoring of the code.

amandahla:
Related:
https://github.com/canonical/synapse-operator/pull/602#issuecomment-2498000253

src/charm_state.py
Thanhphan1147: thanks, I'll fix this in a follow-up PR

PR #603, opened 2024-11-25
ISD-2758 Install mas cli in synapse rock

Thanhphan1147:
This PR builds `mas-cli` and its assets ( css, templates, translations ) in the synapse rock. The PR simply builds the binary and doesn't start any MAS-related workflow. The PR also add some validations to integration tests to ensure that the binary and its assets has been correctly installed.

Note: The changes from L18 to L189 are re-indentation, no code change happened there.

### Checklist

### DIFF
--- a/synapse_rock/rockcraft.yaml
+++ b/synapse_rock/rockcraft.yaml
@@ -208,6 +208,7 @@ parts:
       curl -L -o opa https://openpolicyagent.org/downloads/v0.70.0/opa_linux_amd64_static
       chmod 755 ./opa; mv opa /usr/local/bin
       (cd policies; make)
+      mkdir -p $CRAFT_PART_INSTALL/mas
       cp policies/policy.wasm $CRAFT_PART_INSTALL/mas/policy.wasm
 
       craftctl default
--- a/synapse_rock/rockcraft.yaml
+++ b/synapse_rock/rockcraft.yaml
@@ -112,7 +112,7 @@ parts:
     plugin: nil
     source: https://github.com/element-hq/synapse/
     source-type: git
-    source-tag: v1.118.0
+    source-tag: v1.120.0
     build-environment:
       - RUST_VERSION: "1.76.0"
       - POETRY_VERSION: "1.7.1"
--- a/synapse_rock/rockcraft.yaml
+++ b/synapse_rock/rockcraft.yaml
@@ -195,12 +195,15 @@ parts:
       # because we've already run craftctl default
       mkdir -p $CRAFT_PART_INSTALL/mas/share/assets
       cp -r frontend/dist/* $CRAFT_PART_INSTALL/mas/share/assets/
-      cp frontend/dist/manifest.json $CRAFT_PART_INSTALL/mas/manifest.json
+      cp frontend/dist/manifest.json $CRAFT_PART_INSTALL/mas/share/manifest.json
     stage:
       - mas/*
   mas-cli:
     plugin: rust
     rust-channel: stable
+    rust-cargo-parameters:
+      - "--features dist"
+      - "--no-default-features"
     source: https://github.com/element-hq/matrix-authentication-service
     source-type: git
     source-tag: v0.12.0
@@ -208,15 +211,15 @@ parts:
       curl -L -o opa https://openpolicyagent.org/downloads/v0.70.0/opa_linux_amd64_static
       chmod 755 ./opa; mv opa /usr/local/bin
       (cd policies; make)
-      mkdir -p $CRAFT_PART_INSTALL/mas
-      cp policies/policy.wasm $CRAFT_PART_INSTALL/mas/policy.wasm
+      mkdir -p $CRAFT_PART_INSTALL/mas/share
+      cp policies/policy.wasm $CRAFT_PART_INSTALL/mas/share/policy.wasm
 
       craftctl default
     organize:
       "mas-cli": usr/bin/mas-cli
     stage:
       - usr/bin/mas-cli
-      - mas/policy.wasm
+      - mas/share/policy.wasm
   mas-templates:
     plugin: dump
     source: https://github.com/element-hq/matrix-authentication-service
--- a/synapse_rock/rockcraft.yaml
+++ b/synapse_rock/rockcraft.yaml
@@ -12,6 +12,10 @@ build-base: ubuntu@22.04
 license: Apache-2.0
 platforms:
   amd64:
+environment: &mas-source
+  source: https://github.com/element-hq/matrix-authentication-service
+  source-type: git
+  source-tag: v0.12.0
 parts:
   scripts:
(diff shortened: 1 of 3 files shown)


synapse_rock/rockcraft.yaml
```diff
@@ -10,176 +10,219 @@ build-base: ubuntu@24.04
...
+    plugin: dump
+    source: attributemaps
+    organize:
+      login_ubuntu.py: /usr/local/attributemaps/login_ubuntu.py
+  mas-templates:
+    plugin: dump
+    <<: *mas-source
+    source-subdir: templates
+    organize:
+      "*": mas/share/templates/
```
amandahla, line 195: Out of curiosity: Is there a reason for dumping files in different parts?

synapse_rock/rockcraft.yaml
```diff
@@ -10,176 +10,219 @@ build-base: ubuntu@24.04
...
+  mas-translations:
+    plugin: dump
+    <<: *mas-source
+    source-subdir: translations
+    organize:
+      "*": mas/share/translations/
+  mas-assets:
+    plugin: nil
```
amandahla, line 203: Would be possible to use npm plugin?

synapse_rock/rockcraft.yaml
```diff
@@ -10,176 +10,219 @@ build-base: ubuntu@24.04
...
+      - mas/*
+  mas-cli:
+    plugin: rust
+    rust-channel: stable
+    rust-cargo-parameters:
+      - "--features dist"
+      - "--no-default-features"
+    <<: *mas-source
+    override-build: |
+      curl -L -o opa https://openpolicyagent.org/downloads/v0.70.0/opa_linux_amd64_static
```
amandahla, line 226: They are using a different version in the upstream. Could this make a difference during execution?
https://github.com/element-hq/matrix-authentication-service/blob/main/Dockerfile#L13

synapse_rock/rockcraft.yaml
```diff
@@ -10,176 +10,219 @@ build-base: ubuntu@24.04
...
+    <<: *mas-source
+    source-subdir: translations
+    organize:
+      "*": mas/share/translations/
+  mas-assets:
+    plugin: nil
+    <<: *mas-source
+    source-subdir: frontend
+    build-environment:
+      - NODE_URI: "https://nodejs.org/dist/v22.11.0/node-v22.11.0-linux-x64.tar.gz"
```
amandahla, line 207: Same question here about a different version
https://github.com/element-hq/matrix-authentication-service/blob/24be677a0669c212bc5dec008968e5af79165248/Dockerfile#L12

synapse_rock/rockcraft.yaml
amandahla: (same for rust)

synapse_rock/rockcraft.yaml
Thanhphan1147: The main reason is I want to properly organize each component of MAS and give each of them their own part, this can also come in handy later if we want to do some patching of the translations or html templates

synapse_rock/rockcraft.yaml
Thanhphan1147: I tried with npm plugin first but I ran into a problem that it doesn't install the dev dependencies, so I switched to installing npm manually

synapse_rock/rockcraft.yaml
Thanhphan1147: hmm, I'm not sure if we need to match the version exactly since they're build time dependencies rather than runtime

amandahla: I think would be safer using same versions to avoid surprises in the future

synapse_rock/rockcraft.yaml
```diff
@@ -10,176 +10,219 @@ build-base: ubuntu@24.04
...
+      - zlib1g-dev
+    stage-packages:
+      - bash
+      - coreutils
+      - curl
+      - gosu
+      - gpg
+      - gpg-agent
+      - libffi-dev
+      - libicu70
```
amandahla: this looks different than the one in main

Thanhphan1147: Fixed, nice catch, this PR was made before your PR to switch base landed so that part was not up-to-date. Thanks!

synapse_rock/rockcraft.yaml
Thanhphan1147: Makes sense, but if you look at the [installation guide](https://element-hq.github.io/matrix-authentication-service/setup/installation.html#building-from-the-source) for MAS, it doesn't actually pin the exact version of OPA or npm, while the rust toolchain should be on `stable` which is currently being used in the rock already.

Thanhphan1147: So I think we don't need to match what's in the dockerfile, we're building with `--feature dist` instead of `--feature docker` anyways.

synapse_rock/rockcraft.yaml
Thanhphan1147: Thanks, I replied to that comment

synapse_rock/rockcraft.yaml
amandahla: I see, makes sense.

amandahla (approved):

javierdelapuente (approved):

Thanhphan1147:
I've disable the review dismissal on commits as I merged main into this branch to properly generate the diff

tests/integration/test_charm.py
```diff
@@ -47,6 +47,25 @@ async def test_synapse_is_up(
...
+        "  -f /mas/share/policy.wasm   -a"
+        "  -f /mas/share/manifest.json -a"
+        "  -d /mas/share/templates     -a"
+        "  -d /mas/share/translations ] && echo ok"
+    )
+    action = await unit.run("/bin/bash -c " f"'{pebble_exec_cmd} {check_assets_cmd}'")
+    await action.wait()
+    assert action.results["return-code"] == 0, "mas assets folder not found."
+    assert "ok" in action.results["stdout"]
+
```
Thanhphan1147, line 68: This will be updated once MAS is started with pebble.

synapse_rock/rockcraft.yaml
```diff
@@ -183,3 +187,51 @@ parts:
...
+      - mas/*
+  mas-cli:
+    plugin: rust
+    rust-channel: stable
+    rust-cargo-parameters:
+      - "--features dist"
+      - "--no-default-features"
+    <<: *mas-source
+    override-build: |
+      curl -L -o opa https://openpolicyagent.org/downloads/v0.70.0/opa_linux_amd64_static
```
Thanhphan1147, line 226: I'll make a follow-up PR to build the opa from source / verify the binary checksum

PR #612, opened 2024-11-28
ISD-2788 Add mas database integration

Thanhphan1147:
Add mas-database integration to provide postgresql to MAS. The charm will block until this integration is established.

This PR also introduce the concept of charm state with multiple components, each can be instantiated separately. 

Currently the charm only create a second instance of `DatabaseObserver` and a new charm state component `MASConfiguration` that stores the datasource.

This PR also reworks how charm state is created and validated. I removed the `inject_charm_state` decorator and replace it with a `validate_charm_state` decorator that only does exception handling and setting the unit to the correct state.

Initializing the charm state components is now done explicitly in each hook methods, and the exceptions raised from that is caught and handled by the decorator. Otherwise it will flow naturally and make the charm go into an error state.

### Checklist


src/charm.py
```diff
@@ -276,13 +280,12 @@ def _set_workload_version(self) -> None:
...
             logger.debug("Cannot set workload version at this time: %s", exc)
 
-    @inject_charm_state
-    def _on_config_changed(self, _: ops.HookEvent, charm_state: CharmState) -> None:
-        """Handle changed configuration.
+    @validate_charm_state
+    def _on_config_changed(self, _: ops.HookEvent) -> None:
+        """Handle changed configuration."""
+        charm_state = self.build_charm_state()
+        MASConfiguration.validate(self)
```
javierdelapuente, line 290: I like the idea of just validating instead of creating the charm state. Also splitting the charm state can be a good thing. However, now, we have this everywhere:
```
        charm_state = self.build_charm_state()
        MASConfiguration.validate(self)
....
        reconcile(charm_state)
```

How far are we from building the charm state in the reconcile and having most of the handlers empty? Is it realistic?

Thanhphan1147: I can create the state in `reconcile` and have the decorator only on that method, I think I'd need to update the prototype of the `wrapper` method from `(instance: C, event: E)` to something that matches `reconcile`. 

But from my experience this can be a problem in the future if we want to defer some events (in that case we'd need to pass the event to `reconcile`) or if we want to have multiple separate `reconcile` flows and run only a few of them for some hooks rather than having to run the big `reconcile` method. (this also fits nicely with the separated charm state components as we only initialize the states that the hook needs and use them to only reconcile a specific part of the charm rather than the whole thing).

I also like initializing the states in the hooks like this because it gives us a bit of structure to the hook and show me clearly which are the components they need.

Obviously for now every hook initializes the same state so it's a bit redundant but I think if we keep trying to split the big charm state and the `reconcile` method initializing the states at the beginning of the hooks like this will become necessary.

What do you think?

src/charm.py
```diff
@@ -292,14 +295,16 @@ def _on_config_changed(self, _: ops.HookEvent, charm_state: CharmState) -> None:
...
+    @validate_charm_state
+    def _on_relation_departed(self, event: RelationDepartedEvent) -> None:
         """Handle Synapse peer relation departed event.
 
         Args:
             event: relation departed event.
-            charm_state: The charm state.
         """
+        charm_state = self.build_charm_state()
+        MASConfiguration.validate(self)
```
amandahla, line 309: Couldnt this be added to the self.build_charm_state()?

amandahla:
Many things in one PR :eyes: Is it expected to have charm configurations related to MAS?

src/charm.py
Thanhphan1147: I was thinking of spliting the charm state into multiple components so that they can be initialized separately, and ideally in the future the hooks can choose what components they need and only initialize those components instead of the whole charm state. So I wouldn't want to put them together into a wrapper method.

And also in this PR I didn't use the initialized MAS charm state so I use the `validate` method here, but in the next PR I changed this to `mas_configuration = MASConfiguration.from_charm(self)`

Thanhphan1147:
> Many things in one PR 👀 Is it expected to have charm configurations related to MAS?

@amandahla I have plans to move some charm config like `enable_password_config` to the MAS charm state since it only concern MAS. In the future we might also look at adding config to allow users to modify the oidc subject claim so that if they switch oidc providers the local users can still be detected.

gregory-schiano (approved):
LGTM

amandahla (approved):

src/charm.py
Thanhphan1147: I'm marking this as resolved, we can continue the discussion internally if needed!

PR #620, opened 2024-12-05
ISD-2817 Add mas to pebble plan

Thanhphan1147:
This PR adds a pebble layer that starts the Matrix Authentication Service:

- Adds a method to generate synapse's `msc3861` config to activate MAS
- Update nginx config template to route the correct traffic to MAS
- Rework the `register-user` action and add a new `verify-user-email` action based on the mas-cli
- Update tests

❗This PR also removes support for mjolnir. As it is currently incompatible with MAS.

Mjolnir relies on synapse's `/_synapse/admin/v1/users/:user_id/admin` API for most of its commands (ref: https://github.com/matrix-org/mjolnir/blob/a7f49b816022112a78ea35a2e30a06092e77c2eb/src/Mjolnir.ts#L528) which has been disabled in MAS as per this PR : https://github.com/matrix-org/synapse/pull/15582

```
The following admin APIs are disabled:

/users/:user_id/admin
/users/:user_id/login
/registration_tokens
/registration_tokens/new
/registration_tokens/:token
/reset_password
/register
```

### Checklist


src/auth/mas.py
```diff
@@ -2,20 +2,155 @@
...
+                "startup": "enabled",
+                "command": f"{MAS_EXECUTABLE_PATH} server -c {MAS_CONFIGURATION_PATH}",
+                "working-dir": MAS_WORKING_DIR,
+            }
+        },
+    }
+)
+
+
+class MASConfigInvalidError(Exception):
```
amandahla, line 45: I wonder if we really need different exceptions for ops.pebble.ExecError. Are they handled differently?

src/auth/mas.py
```diff
@@ -2,20 +2,155 @@
...
+    process = container.exec(command=command, working_dir=MAS_WORKING_DIR)
+    process.wait()
+
+
+def register_user(
+    container: ops.model.Container,
+    username: str,
+    is_admin: bool = False,
+) -> str:
+    """Register a new user with MAS.
```
amandahla, line 88: Out of curiosity: registering this way does not require an e-mail, right?

src/auth/mas.py
```diff
@@ -2,20 +2,155 @@
...
+
+    return password
+
+
+def verify_user_email(
+    container: ops.model.Container,
+    username: str,
+    email: str,
+) -> None:
+    """Verify a user email with mas-cli.
```
amandahla, line 130: Should we use this action only if Synapse has no SMTP integration?

src/auth/mas.py
```diff
@@ -58,3 +193,44 @@ def generate_mas_config(
...
+    """
+    mas_context = mas_configuration.mas_context
+
+    mas_prefix = mas_configuration.mas_prefix
+    # We explicitly set the oauth2 endpoints using MAS local address
+    # This is to avoid problems with TLS self-signed certificates
+    # when the charm is behind an https ingress
+    mas_local_address = f"http://localhost:8081{mas_prefix}"
+    # MAS public address is used when redirecting the client to MAS for login
+    mas_public_address = f"{synapse_configuration.public_baseurl}{mas_prefix}"
```
amandahla, line 246: at this point, should we validate the URL?

src/pebble.py
```diff
@@ -299,9 +315,12 @@ def _environment_has_changed(
 
 
 # The complexity of this method will be reviewed.
-def reconcile(  # noqa: C901 pylint: disable=too-many-branches,too-many-statements
+# pylint: disable=too-many-branches,too-many-statements
+# pylint: disable=too-many-arguments,too-many-positional-arguments
+def reconcile(  # noqa: C901
```
amandahla, line 294: I'm afraid of how big this reconcile is getting :-P

src/pebble.py
```diff
@@ -401,6 +419,11 @@ def reconcile(  # noqa: C901 pylint: disable=too-many-branches,too-many-statemen
             ignore_order=True,
             ignore_string_case=True,
         )
+
+        restart_mas(container, rendered_mas_configuration)
```
amandahla: Considering that MAS is now a requirement for Synapse (right?), I think this could be done as the first step, before anything related to Synapse.

amandahla:
Will this change impact here?
https://github.com/canonical/synapse-operator/blob/83c9b1f627ae75145850e836d8764b68baee10ca/src/admin_access_token.py#L105

src/pebble.py
gregory-schiano: We discussed with Thanh about this, this method definitely requires some refactoring

requirements.txt
```diff
@@ -7,5 +7,5 @@ jsonschema ==4.23.0
 ops ==2.17.1
 psycopg2-binary ==2.9.10
 pydantic ==2.10.3
-python-ulid ==3.0.0
 requests ==2.32.3
+python-ulid ==3.0.0
```
gregory-schiano: The ordering should be fixed

Thanhphan1147: I sorted them in alphabetical order

src/auth/mas.py
Thanhphan1147: Yes, the user will be prompted to add an email when they log-in with the client. And there's also the `verify-user-email` action to verify the email in case smtp is not configured

src/auth/mas.py
Thanhphan1147: yes, are you suggesting that we add a logic here to do nothing if SMTP is related?

Thanhphan1147: I think even with SMTP we might still want to use this action in special cases where the mail server might not be working properly

src/auth/mas.py
Thanhphan1147: The baseurl should already be verified at this point when we initialize the charm state, so I don't think we need to re-validate it here as mas_prefix is only a constant

src/auth/mas.py
amandahla: I agree

src/auth/mas.py
```diff
@@ -2,32 +2,227 @@
...
+    try:
+        process = container.exec(command=command, working_dir=MAS_WORKING_DIR)
+        process.wait_output()
+    except ops.pebble.ExecError as exc:
+        logger.error("Error verifying the user email: %s", exc.stderr)
+        raise MASVerifyUserEmailFailedError("Error verifying the user email.") from exc
+
+
+def generate_admin_access_token(container: ops.model.Container, username: str) -> str:
+    """Verify a user email with mas-cli.
```
amandahla: Minor: is this description correct?

templates/mas_config.yaml.j2
```diff
@@ -46,3 +47,18 @@ templates:
   translations_path: /mas/share/translations
 policy:
   wasm_module: /mas/share/policy.wasm
+{% if smtp_configuration is not none %}
+email:
+  from: '"The almighty auth service" <auth@example.com>'
```
amandahla: Could this be replaced by something like MAS [homeserver value] ?

tests/integration/conftest.py
```diff
@@ -250,30 +264,41 @@ def user_username_fixture() -> typing.Generator[str, None, None]:
...
     Returns:
         The access token
     """
-    synapse_ip = (await get_unit_ips(synapse_app.name))[0]
-    return get_access_token(synapse_ip, user_username, user_password)
+    username, _ = user
+    pebble_exec_cmd = "PEBBLE_SOCKET=/charm/containers/synapse/pebble.socket pebble exec --"
+    generate_token_cmd = (
+        f"{pebble_exec_cmd} mas-cli -c {MAS_CONFIGURATION_PATH} manage issue-compatibility-token "
+        f"--yes-i-want-to-grant-synapse-admin-privileges {username}"
```
amandahla, line 291: cool parameter :smile:

src/auth/mas.py
Thanhphan1147: Not differently, do you think it would be better to not wrap ops.pebble.ExecError?

tests/integration/test_charm.py
```diff
@@ -318,40 +241,13 @@ async def test_anonymize_user(
     )
     await action_register_user.wait()
     assert action_register_user.status == "completed"
```
amandahla, line 243: Once MAS is set, is there another way of checking/calling the Synapse API? This would allow us to validate whether the action succeeded.

tests/integration/conftest.py
Thanhphan1147: And without a shorthand for some reason

tests/integration/test_charm.py
Thanhphan1147: We have the dedicated MAS admin api for that. But I don't think we can reuse the synapse API client as this API uses OIDC spec for authentication / authorization. I think we can leave it like this for now since doing that would be quite an involved task.

src/auth/mas.py
Thanhphan1147: Actually I'll remove this since it's not currently being used ( it was added to support mjolnir but we removed mjolnir )

templates/mas_config.yaml.j2
Thanhphan1147: I updated the email template

src/auth/mas.py
```diff
@@ -2,32 +2,192 @@
...
 
-# pylint: disable=too-few-public-methods
+def validate_mas_config(container: ops.model.Container) -> None:
+    """Validate current MAS configuration.
+
+    Args:
+        container: Synapse container.
+    """
+    command = [MAS_EXECUTABLE_PATH, "config", "check", "-c", MAS_CONFIGURATION_PATH]
+    process = container.exec(command=command, working_dir=MAS_WORKING_DIR)
```
gregory-schiano, line 68: Should wrap the pebble exec error into a business error, same comment for `sync_mas_config`

src/auth/mas.py
```diff
@@ -2,32 +2,192 @@
...
+        MAS_EXECUTABLE_PATH,
+        "-c",
+        MAS_CONFIGURATION_PATH,
+        "manage",
+        "lock-user",
+        "--deactivate",
+        username,
+    ]
+
+    process = container.exec(command=command, working_dir=MAS_WORKING_DIR, combine_stderr=True)
```
gregory-schiano, line 175: Should wrap pebble exec error

amandahla:
Considering a refresh/upgrade scenario where Mjolnir was enabled, should the charm delete the previously created management room (also the rate limit configuration or anything related to mjolnir)?
https://github.com/canonical/synapse-operator/blob/435a99db35cb2727b58c231ed87690c7c9cc736f/src/mjolnir.py#L214

Thanhphan1147:
> Considering a refresh/upgrade scenario where Mjolnir was enabled, should the charm delete the previously created management room (also the rate limit configuration or anything related to mjolnir)?
> 
> https://github.com/canonical/synapse-operator/blob/435a99db35cb2727b58c231ed87690c7c9cc736f/src/mjolnir.py#L214

@amandahla We don't support automatic upgrade from older revisions to avoid the complexity related to migration of SAML users to OIDC in MAS (which involves modifying the `external-id` in synapse db, aligning the local subject with the upstream OIDC subject claim, having OIDC configured before user migration...). So any upgrades we do will be manual and there we just need to keep in mind to remove mjolnir-related user and rooms.

docs/explanation/moderation-and-spam-control.md
```diff
@@ -60,22 +60,4 @@ For details and implementation, visit the module’s repository: [Synapse Invite
...
-- Room directory changes and alias transfers: Adjust room visibility and manage
-aliases efficiently.
-- Account deactivation: Disable abusive or compromised accounts.
-- Room shutdown: Close problematic rooms completely.
-
-### More information
-
-For more details and implementation guidance, refer to the [Mjolnir GitHub repository](https://github.com/matrix-org/mjolnir).
-
+With the arrival of MAS mjolnir has been temporary disabled.
```
amandahla, line 63: Regarding what we discussed about refresh, can you add a note here in the doc about manually removing what was created?

tests/unit/test_action.py
```diff
@@ -0,0 +1,192 @@
+# Copyright 2025 Canonical Ltd.
```
javierdelapuente, line 1: should not we stop using harness and use scenario instead for new tests?

Thanhphan1147: That would mean rewriting the majority of fixtures in `conftest` in scenario I believe, which could be quite involved. My thought process is that I'd keep this PR simple and (ideally) move everything to scenario in a separate PR. what do you think?

docs/explanation/moderation-and-spam-control.md
Thanhphan1147: I'm thinking of adding a dedicated doc page for manual upgrade and I'll add a link to that page here once that's been done, what do you think?

src/auth/mas.py
Thanhphan1147: I updated the code to raise a custom exception here

src/auth/mas.py
Thanhphan1147: I updated the code to raise a custom exception here

docs/explanation/moderation-and-spam-control.md
amandahla: Nice :)

amandahla (approved):

javierdelapuente (approved):

src/auth/mas.py
Thanhphan1147: After discussing it with Greg I decided that it would be better to wrap ExecError with a custom business error

src/pebble.py
Thanhphan1147: Good point, I moved this to the beginning of the method.

PR #621, opened 2024-12-06
ISD-2761 Add oauth lib and requirer

Thanhphan1147:
Add the oidc interface to configure `upstream_providers` in MAS.

### Checklist


requirements.txt
```diff
@@ -7,5 +7,5 @@ jsonschema ==4.23.0
 ops ==2.17.1
 psycopg2-binary ==2.9.10
 pydantic ==2.10.3
-python-ulid ==3.0.0
 requests ==2.32.3
+python-ulid ==3.0.0
```
gregory-schiano: Keep alphabetic ordering

src/charm.py
```diff
@@ -99,10 +107,15 @@ def __init__(self, *args: typing.Any) -> None:
...
         self.framework.observe(self.on.register_user_action, self._on_register_user_action)
+        self.framework.observe(self.on.verify_user_email_action, self._on_verify_user_email_action)
+
         self.framework.observe(
             self.on.promote_user_admin_action, self._on_promote_user_admin_action
         )
         self.framework.observe(self.on.anonymize_user_action, self._on_anonymize_user_action)
+        self.framework.observe(self._oauth.on.oauth_info_changed, self._on_config_changed)
+        self.framework.observe(self._oauth.on.oauth_info_removed, self._on_config_changed)
+        self.framework.observe(self._oauth.on.invalid_client_config, self._on_config_changed)
```
gregory-schiano, line 113: I'd rename the hook method to something more "generic"

src/pebble.py
```diff
@@ -401,6 +419,11 @@ def reconcile(  # noqa: C901 pylint: disable=too-many-branches,too-many-statemen
             ignore_order=True,
             ignore_string_case=True,
         )
+
+        restart_mas(container, rendered_mas_configuration)
```
gregory-schiano: To be set before restart_syname with a comment explaning it must be done before synapse

src/pebble.py
```diff
@@ -401,6 +419,11 @@ def reconcile(  # noqa: C901 pylint: disable=too-many-branches,too-many-statemen
...
+        # Activate MAS on synapse
+        synapse.configure_mas(current_synapse_config, synapse_msc3861_configuration)
```
gregory-schiano: Move it before the DeepDiff to allow synapse to restart when the config has changed

javierdelapuente:
could you update the pull request description? Thanks!

src/auth/mas.py
```diff
@@ -119,9 +134,36 @@ def register_user(
...
+    command = [
+        MAS_EXECUTABLE_PATH,
+        "-c",
+        MAS_CONFIGURATION_PATH,
+        "manage",
+        "provision-all-users",
+    ]
+
+    try:
+        process = container.exec(command=command, working_dir=MAS_WORKING_DIR)
```
amandahla: Would it make sense to add a timeout here? Could this command possibly take too much time?

templates/mas_config.yaml.j2
```diff
@@ -35,18 +35,42 @@ passwords:
...
   wasm_module: /mas/share/policy.wasm
+{% if oauth_provider_info is not none %}
+upstream_oauth2:
+  providers:
+    - id: {{ upstream_oidc_provider_id }}
+      human_name: Single Sign-On
+      issuer: {{ oauth_provider_info.issuer_url }}
+      client_id: {{ oauth_provider_info.client_id }}
+      client_secret: {{ oauth_provider_info.client_secret }}
+      scope: "openid profile email"
```
amandahla: Should this be set by the constant MAS_OIDC_SCOPE (if it changes)?

src/pebble.py
Thanhphan1147: moved!

templates/mas_config.yaml.j2
Thanhphan1147: Thanks, I updated it to use the constant

src/pebble.py
Thanhphan1147: thanks, I moved this to before the diff

src/auth/mas.py
Thanhphan1147: This command simply schedule a background job and exit immediately iirc, so I don't think we need to add a timeout here.

Thanhphan1147:
Hi @gregory-schiano / @amandahla, can you take a look and review the PR when you have the time? thanks!

src/charm.py
Thanhphan1147: I think we have a lot of events hooked to `_on_config_changed` for now, I'll probably change this in a follow-up PR

src/auth/mas.py
```diff
@@ -146,9 +154,36 @@ def register_user(
...
+
+    Raises:
+        MASProvisionUserFailedError: when triggering the provisioning jobs fails.
+    """
+    command = [
+        MAS_EXECUTABLE_PATH,
+        "-c",
+        MAS_CONFIGURATION_PATH,
+        "manage",
+        "provision-all-users",
```
amandahla: Does this job have any impact on MAS?
Also, should this be run after every restart too?

src/charm.py
```diff
@@ -561,6 +578,16 @@ def _on_anonymize_user_action(self, event: ActionEvent) -> None:
...
         )
 
+    def _on_oauth_relation_changed(self, _: RelationChangedEvent) -> None:
+        """Handle oauth_relation_changed event."""
+        charm_state = self.build_charm_state()
+        mas_configuration = MASConfiguration.from_charm(self)
+        oauth_client_config = generate_oauth_client_config(
+            mas_configuration, charm_state.synapse_config
+        )
+        self._oauth.update_client_config(oauth_client_config)
```
amandahla, line 588: Wouldnt this trigger another relation-changed?

Thanhphan1147: I put it here because the oauth client config that we generate is mostly constant, in which case it won't fire a relation changed event.

The only part that can change is the `redirect_uri` which changes if we update the `public_baseurl` config, in which case we'd need to fire a relation changed event again so that the provider updates its oauth client on that side.

src/auth/mas.py
Thanhphan1147: I'm also not sure if this command is necessary, will confirm with MAS team and let you know.

Thanhphan1147: I got an update from MAS team and basically it ensures that MAS users are synced with the synapse DB, although we won't need to run it every time a new user is added as MAS will take care of that. So I removed the command wrapper for now, and potentially add it back as a charm action if necessary.

amandahla (approved):

src/charm.py
```diff
@@ -90,6 +93,7 @@ def __init__(self, *args: typing.Any) -> None:
             relation_name=INGRESS_INTEGRATION_NAME,
             port=synapse.SYNAPSE_NGINX_PORT,
         )
+        self._oauth = OAuthRequirer(self)
```
javierdelapuente, line 96: I would rather pass the `relation_name=OAUTH_INTEGRATION_NAME`, or instead use the `DEFAULT_RELATION_NAME` in the library

javierdelapuente (approved):
LGTM.

Would it be interesting to have an integration test with oauth?

src/charm.py
Thanhphan1147: makes sense, I'll update it in a follow-up PR.

PR #623, opened 2024-12-06
Update matrix-auth library

amandahla:
### Overview

In a CMR, a provider cannot read from a secret created by the required.

Related:
https://bugs.launchpad.net/juju/+bug/2060222

This PR then changes the library to encrypt the content that the requirer needs to share with the provider.

Flow
1. The provider creates a key and stores it in a Juju secret.
2. The requirer reads this key and use it to encrypt content. The content is shared via relation data.
3. The provider uses the key to decrypt the content.

Case:
Appservices (IRC Bridge) needs to share a registration file with Synapse and Synapse needs to register this file as a configuration in the workload.

### Rationale

Allow easy and safe two-way data sharing.

### Checklist


src/matrix_auth_observer.py
```diff
@@ -87,6 +87,7 @@ class RegistrationSecret(NamedTuple):
 
         for relation in list(self._charm.model.relations["matrix-auth"]):
             requirer_data = MatrixAuthRequirerData.from_relation(self.model, relation=relation)
+            print(requirer_data)
```
gregory-schiano: Should be removed

lib/charms/synapse/v1/matrix_auth.py
```diff
@@ -0,0 +1,552 @@
...
+        """Encrypt a string using AES-256-ECB.
+
+        Args:
+            key: aes key in bytes.
+            plaintext: text to encrypt.
+
+        Returns:
+            encrypted text.
+        """
+        if not plaintext:
```
nrobinaubertin: We maybe should encrypt everything, even empty strings

lib/charms/synapse/v1/matrix_auth.py
```diff
@@ -0,0 +1,552 @@
...
+            return ""
+        # Initialize the cipher with AES-256 and ECB mode
+        cipher = Cipher(algorithms.AES(key), modes.ECB())
```
nrobinaubertin: Is ECB really the mode that we're looking for ? I wonder if we should use CBC instead

lib/charms/synapse/v1/matrix_auth.py
```diff
@@ -0,0 +1,552 @@
...
+        encryptor = cipher.encryptor()
+
+        # Pad the plaintext to match the block size (16 bytes)
+        padder = PKCS7(algorithms.AES.block_size).padder()
+        padded_data = padder.update(plaintext.get_secret_value().encode()) + padder.finalize()
```
nrobinaubertin: Isn't there something in the cryptography library that we can use without running into crypto-footguns ? I have the feeling that's the kind of magic we don't want to mess with.

nrobinaubertin: Something like:
```
algorithms.use_this_AES_when_you_dont_know_what_youre_doing(text_to_encrypt, key)
```

amandahla: Thanks for raising this, I replaced it  with Fernet as suggested and it looks much better now.

nrobinaubertin (approved):

lib/charms/synapse/v1/matrix_auth.py
```diff
@@ -240,25 +239,14 @@ def encrypt_string(cls, key: bytes, plaintext: str) -> str:
...
-        padder = PKCS7(algorithms.AES.block_size).padder()
-        padded_data = padder.update(plaintext.get_secret_value().encode()) + padder.finalize()
-
-        # Encrypt the padded plaintext
-        ciphertext = encryptor.update(padded_data) + encryptor.finalize()
-
-        # Return ciphertext as Base64 for easy handling
-        return base64.b64encode(ciphertext).decode()
+        plaintext = cast(SecretStr, plaintext)
+        f = Fernet(key)
```
gregory-schiano: The variable name should probably be more explicit like `encryptor`, `decryptor`

lib/charms/synapse/v1/matrix_auth.py
```diff
@@ -0,0 +1,530 @@
...
+logger = logging.getLogger(__name__)
+
+#### Constants ####
+APP_REGISTRATION_LABEL = "app-registration"
+APP_REGISTRATION_CONTENT_LABEL = "app-registration-content"
+DEFAULT_RELATION_NAME = "matrix-auth"
+SHARED_SECRET_LABEL = "shared-secret"
+SHARED_SECRET_CONTENT_LABEL = "shared-secret-content"
+AES_KEY_SECRET_LABEL = "aes-key-secret"
+AES_KEY_SECRET_CONTENT_LABEL = "aes-key-content"
```
gregory-schiano: We should probably rename all variables, names related to the encryption to avoid being "implementation" specific, and therefore not specifying "AES" but more "ENCRYPTION_KEY...."

gregory-schiano (commented):
Few nitpick comments

lib/charms/synapse/v1/matrix_auth.py
Thanhphan1147, line 1: should we add the reference in the charm-libs section of `charmcraft.yaml` like [here](https://github.com/canonical/synapse-operator/blob/79124d1fa040e57a44f56d3c7931574b38525489/charmcraft.yaml#L15)?

lib/charms/synapse/v1/matrix_auth.py
```diff
@@ -0,0 +1,530 @@
...
+# to 0 if you are raising the major API version
+LIBPATCH = 3
+
+# pylint: disable=wrong-import-position
+import json
+import logging
+from typing import Dict, List, Optional, Tuple, cast
+import secrets
+import base64
+from cryptography.fernet import Fernet
```
Thanhphan1147, line 78: Just out of curiosity, are there any specific reasons why we use Fernet?

Thanhphan1147 (approved):
LGTM, I only have a few comments

lib/charms/synapse/v1/matrix_auth.py
```diff
@@ -0,0 +1,530 @@
...
+
+import ops
+from pydantic import BaseModel, Field, SecretStr
+
+logger = logging.getLogger(__name__)
+
+#### Constants ####
```
javierdelapuente: Not sure this comment brings value

lib/charms/synapse/v1/matrix_auth.py
amandahla: @nrobinaubertin can explain better than me ;)

lib/charms/synapse/v1/matrix_auth.py
amandahla: TBH I'm unsure if this attribute applies to libraries provided by the charm.  The documentation is not clear about it.

lib/charms/synapse/v1/matrix_auth.py
gregory-schiano: Initially we wanted to use AES, but this requires attention to how it's implemented to make sure it's used securely. Fernet is an opinionated and secure way to use AES, so basically doing what we wanted to do, in a better way than we would do (or in a similar way). In the end it simplifies the code in the charm to avoid the complexity of generating the IV, ensure the padding is right, ...

gregory-schiano:
I made a diff between v0 and v1, it's easier to review:
```
gschiano@charm-dev:/src/Canonical/synapse-operator/lib/charms/synapse$ diff v0/matrix_auth.py v1/matrix_auth.py
66c66
< LIBAPI = 0
---
> LIBAPI = 1
75a76,78
> import secrets
> import base64
> from cryptography.fernet import Fernet
87a91,92
> ENCRYPTION_KEY_SECRET_LABEL = "encryption-key-secret"
> ENCRYPTION_KEY_SECRET_CONTENT_LABEL = "encryption-key-content"
102a108
>     encryption_key_secret_id: Optional[SecretStr] = Field(default=None)
126a133,153
>     def set_encryption_key_secret_id(self, model: ops.Model, relation: ops.Relation) -> None:
>         """Store the encryption key to encrypt/decrypt appservice registrations.
>
>         Args:
>             model: the Juju model
>             relation: relation to grant access to the secrets to.
>         """
>         key = Fernet.generate_key()
>         encryption_key = key.decode('utf-8')
>         try:
>             secret = model.get_secret(label=ENCRYPTION_KEY_SECRET_LABEL)
>             secret.set_content({ENCRYPTION_KEY_SECRET_CONTENT_LABEL: encryption_key})
>             # secret.id is not None at this point
>             self.encryption_key_secret_id = cast(str, secret.id)
>         except ops.SecretNotFoundError:
>             secret = relation.app.add_secret(
>                 {ENCRYPTION_KEY_SECRET_CONTENT_LABEL: encryption_key}, label=ENCRYPTION_KEY_SECRET_LABEL
>             )
>             secret.grant(relation)
>             self.encryption_key_secret_id = cast(str, secret.id)
>
161a189
>         self.set_encryption_key_secret_id(model, relation)
199d226
<         registration_id: the registration Juju secret ID.
203d229
<     registration_secret_id: Optional[SecretStr] = Field(default=None)
205,206c231,233
<     def set_registration_id(self, model: ops.Model, relation: ops.Relation) -> None:
<         """Store the app registration as a Juju secret.
---
>     @classmethod
>     def encrypt_string(cls, key: bytes, plaintext: SecretStr) -> str:
>         """Encrypt a string using Fernet.
209,210c236,240
<             model: the Juju model
<             relation: relation to grant access to the secrets to.
---
>             key: encryption key in bytes.
>             plaintext: text to encrypt.
>
>         Returns:
>             encrypted text.
212,226c242,245
<         # password is always defined since pydantic guarantees it
<         password = cast(SecretStr, self.registration)
<         # pylint doesn't like get_secret_value
<         secret_value = password.get_secret_value()  # pylint: disable=no-member
<         try:
<             secret = model.get_secret(label=APP_REGISTRATION_LABEL)
<             secret.set_content({APP_REGISTRATION_CONTENT_LABEL: secret_value})
<             # secret.id is not None at this point
<             self.registration_secret_id = cast(str, secret.id)
<         except ops.SecretNotFoundError:
<             secret = relation.app.add_secret(
<                 {APP_REGISTRATION_CONTENT_LABEL: secret_value}, label=APP_REGISTRATION_LABEL
<             )
<             secret.grant(relation)
<             self.registration_secret_id = cast(str, secret.id)
---
>         plaintext = cast(SecretStr, plaintext)
>         encryptor = Fernet(key)
>         ciphertext = encryptor.encrypt(plaintext.get_secret_value().encode('utf-8'))
>         return ciphertext.decode()
229,232c248,266
<     def get_registration(
<         cls, model: ops.Model, registration_secret_id: Optional[str]
<     ) -> Optional[SecretStr]:
<         """Retrieve the registration corresponding to the registration_secret_id.
---
>     def decrypt_string(cls, key: bytes, ciphertext: str) -> str:
>         """Decrypt a string using Fernet.
>
>         Args:
>             key: encryption key in bytes.
>             ciphertext: encrypted text.
>
>         Returns:
>             decrypted text.
>         """
>         decryptor = Fernet(key)
>         plaintext = decryptor.decrypt(ciphertext.encode('utf-8'))
>         return plaintext.decode()
>
>     @classmethod
>     def get_encryption_key_secret(
>         cls, model: ops.Model, encryption_key_secret_id: Optional[str]
>     ) -> Optional[bytes]:
>         """Retrieve the encryption key secret corresponding to the encryption_key_secret_id.
236c270
<             registration_secret_id: the secret ID for the registration.
---
>             encryption_key_secret_id: the secret ID for the encryption key secret.
239c273
<             the registration or None if not found.
---
>             the encryption key secret  as bytes or None if not found.
241,242d274
<         if not registration_secret_id:
<             return None
244,246c276,282
<             secret = model.get_secret(id=registration_secret_id)
<             password = secret.get_content().get(APP_REGISTRATION_CONTENT_LABEL)
<             if not password:
---
>             if not encryption_key_secret_id:
>                 # then its the provider and we can get using label
>                 secret = model.get_secret(label=ENCRYPTION_KEY_SECRET_LABEL)
>             else:
>                 secret = model.get_secret(id=encryption_key_secret_id)
>             encryption_key = secret.get_content().get(ENCRYPTION_KEY_SECRET_CONTENT_LABEL)
>             if not encryption_key:
248c284
<             return SecretStr(password)
---
>             return encryption_key.encode('utf-8')
260a297,299
>
>         Raises:
>             ValueError if encryption key not found.
262,263c301,309
<         self.set_registration_id(model, relation)
<         dumped_model = self.model_dump(exclude_unset=True)
---
>         # get encryption key
>         app = cast(ops.Application, relation.app)
>         relation_data = relation.data[app]
>         encryption_key_secret_id = relation_data.get("encryption_key_secret_id")
>         encryption_key = MatrixAuthRequirerData.get_encryption_key_secret(model, encryption_key_secret_id)
>         if not encryption_key:
>             raise ValueError("Invalid relation data: encryption_key_secret_id not found")
>         # encrypt content
>         content = MatrixAuthRequirerData.encrypt_string(key=encryption_key, plaintext=self.registration)
265c311
<             "registration_secret_id": dumped_model["registration_secret_id"],
---
>             "registration_secret": content,
282a329
>         # get encryption key
285,286c332,340
<         registration_secret_id = relation_data.get("registration_secret_id")
<         registration = MatrixAuthRequirerData.get_registration(model, registration_secret_id)
---
>         encryption_key_secret_id = relation_data.get("encryption_key_secret_id")
>         encryption_key = MatrixAuthRequirerData.get_encryption_key_secret(model, encryption_key_secret_id)
>         if not encryption_key:
>             logger.warning("Invalid relation data: encryption_key_secret_id not found")
>             return None
>         # decrypt content
>         registration_secret = relation_data.get("registration_secret")
>         if not registration_secret:
>             return MatrixAuthRequirerData()
288c342
<             registration=registration,
---
>             registration=MatrixAuthRequirerData.decrypt_string(key=encryption_key, ciphertext=registration_secret),
```

tests/integration/test_matrix_auth.py
```diff
@@ -0,0 +1,89 @@
...
+    arrange: deploy the Synapse charm, create offer, deploy any-charm as consumer
+        in a different model and consume offer.
+    act: integrate them via matrix-auth offer.
+    assert: Synapse set the registration file received via matrix-auth.
+    """
+    await model.wait_for_idle(idle_period=10, status=ACTIVE_STATUS_NAME)
+    # This workaround was extracted from prometheus-k8s charm.
+    # Without it, the offer creation fails.
+    # https://github.com/canonical/prometheus-k8s-operator/blob/5779ecc749ee1582c6be20030a83472d024cd24f/tests/integration/test_remote_write_with_zinc.py#L103
+    controller = Controller()
```
javierdelapuente, line 42: I think I did something similar in here using [OpsTest](https://github.com/canonical/paas-charm/blob/async-workers/tests/integration/conftest.py#L19) instead (in my case with cross controller relations), which does the same as this code inside.

javierdelapuente (approved):

lib/charms/synapse/v1/matrix_auth.py
amandahla: Removed.

tests/integration/test_matrix_auth.py
amandahla: Nice!

lib/charms/synapse/v1/matrix_auth.py
amandahla: I'll mark this one as resolved but we can discuss it again if we found out how to handle local libs with this charm-libs section

PR #653, opened 2025-01-14
update docs and wrap pebble.ExecError with custom error

Thanhphan1147:
This PR adds missing commits from https://github.com/canonical/synapse-operator/pull/620

- Add missing custom exception wrapping `pebble.ExecError` for `mas-cli` commands 
- Update docs about mjolnir
### Checklist

### DIFF
--- a/src/auth/mas.py
+++ b/src/auth/mas.py
@@ -16,7 +16,6 @@ from state.mas import MASConfiguration
 logger = logging.getLogger()
 
 MAS_TEMPLATE_FILE_NAME = "mas_config.yaml.j2"
-
 MAS_SERVICE_NAME = "synapse-mas"
 MAS_EXECUTABLE_PATH = "/usr/bin/mas-cli"
 MAS_WORKING_DIR = "/mas"
@@ -46,12 +45,20 @@ class MASConfigInvalidError(Exception):
     """Exception raised when validation of the MAS config failed."""
 
 
+class MASConfigSyncError(Exception):
+    """Exception raised when synchronisation of the MAS config failed."""
+
+
 class MASRegisterUserFailedError(Exception):
     """Exception raised when validation of the MAS config failed."""
 
 
 class MASVerifyUserEmailFailedError(Exception):
     """Exception raised when validation of the MAS config failed."""
+
+
+class MASDeactivateUserFailedError(Exception):
+    """Exception raised when deactivation of a MAS user failed."""
 
 
 class MASGenerateAdminAccessTokenError(Exception):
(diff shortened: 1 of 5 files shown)


src/auth/mas.py
```diff
@@ -63,21 +70,37 @@ def validate_mas_config(container: ops.model.Container) -> None:
...
     command = [MAS_EXECUTABLE_PATH, "config", "sync", "--prune", "-c", MAS_CONFIGURATION_PATH]
-    process = container.exec(command=command, working_dir=MAS_WORKING_DIR)
-    process.wait()
+
+    try:
+        process = container.exec(command=command, working_dir=MAS_WORKING_DIR)
+        process.wait()
+    except ops.pebble.ExecError as exc:
+        logger.exception("Error syncing MAS config with the database.")
+        raise MASConfigSyncError("Error validating MAS configuration.") from exc
```
javierdelapuente: Will this exception be catched somewhere? Otherwise I think the unit will go into error state (and currently synapse does not go into error state, but into blocked state I think)

Thanhphan1147: My reasoning is that if this command fails then there's something wrong with the postgresql database which should put the charm into an error state as there's usually no specific action for the operator to take

amandahla: Should the exception description be the same as the logger.exception?

src/auth/mas.py
```diff
@@ -116,7 +139,7 @@ def register_user(
         process = container.exec(command=command, working_dir=MAS_WORKING_DIR)
         process.wait_output()
     except ops.pebble.ExecError as exc:
-        logger.error("Error registering new user: %s", exc.stderr)
+        logger.exception("Error registering new user.")
         raise MASRegisterUserFailedError("Error validating MAS configuration.") from exc
```
amandahla: Same here regarding expcetion description and logger.exception

src/auth/mas.py
javierdelapuente: The issue with getting into error state is that it is problematic.

For example, as you are not using the latest version of the redis library, if you get into error state and the redis url changes (because for example the redis pod was recreated), the charm will not get notified, and you may not get out of that issue (as has happened in the discourse charm). Updating the redis library will fix that, but other issues may arise, as not being able to upgrade the charm in that state...

Thanhphan1147: thanks, I updated it

src/auth/mas.py
Thanhphan1147: Updated!

src/auth/mas.py
Thanhphan1147: Ah I see, I added a catch for all mas-related exception to ensure that we don't go into error state

Thanhphan1147:
Hi @amandahla / @javierdelapuente, could I ask you to take a look again and review this PR? thanks!

javierdelapuente (approved):

amandahla (approved):

PR #664, opened 2025-01-22
Force getting last secret revision in matrix-auth ISD-2913

amandahla:
### Overview

While testing IRC Bridge + Synapse, I noticed that IRC Bridge was not getting the latest secret version, throwing an error while encrypting/decrypting content.

This PR changes the behavior to always get the latest content instead.

Reference:
https://ops.readthedocs.io/en/latest/howto/manage-secrets.html#start-tracking-a-different-secret-revision

### Library Changes

Changes matrix-auth library.

### Checklist

### DIFF
--- a/lib/charms/synapse/v1/matrix_auth.py
+++ b/lib/charms/synapse/v1/matrix_auth.py
@@ -67,7 +67,7 @@ LIBAPI = 1
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 0
+LIBPATCH = 4
 
 # pylint: disable=wrong-import-position
 import json
@@ -195,7 +195,7 @@ class MatrixAuthProviderData(BaseModel):
             return None
         try:
             secret = model.get_secret(id=shared_secret_id)
-            password = secret.get_content().get(SHARED_SECRET_CONTENT_LABEL)
+            password = secret.get_content(refresh=True).get(SHARED_SECRET_CONTENT_LABEL)
             if not password:
                 return None
             return SecretStr(password)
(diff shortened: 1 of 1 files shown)


lib/charms/synapse/v1/matrix_auth.py
```diff
@@ -195,7 +195,7 @@ def get_shared_secret(
             return None
         try:
             secret = model.get_secret(id=shared_secret_id)
-            password = secret.get_content().get(SHARED_SECRET_CONTENT_LABEL)
+            password = secret.get_content(refresh=True).get(SHARED_SECRET_CONTENT_LABEL)
```
javierdelapuente, line 198: wouldn't it be better to subscribe to `secret-changed` and refresh the secret there?

amandahla: I think this way it's easier to guarantee that at any moment the requirer needs it, the latest secret will be fetched.

amandahla: Example: [here](https://github.com/canonical/irc-bridge-operator/blob/96c160c23b412cadc49664e97534e26ea79d747a/src/charm.py#L118) in IRC every reconcile gets data from the relation

javierdelapuente: Not sure if I understood.

If the secret is changed and irc does not listen to the secrets-changes, does it call the reconcile? Wouldn't it be possible to register to the `secrets-changed` in the MatrixAuthRequires, refresth the secret there  and notify the charm? That way also there would only be one access to the juju controller as the refresh=true implies a call to the juju controller.

amandahla: - Once Synapse has a matrix-auth integration in place (Maubot/IRC Bridge) verifies if needs to update the relation data (provider) every reconcile.
- Only the provider creates secrets, one for the registration_shared_secret and other for a encryption key to be used by the requirer.
- The library observes relation-change for Requirer and Provider and emits matrix_auth_request_received/matrix_auth_request_processed handled by charms.

So observing a secrets-change would duplicate the behavior alread handled by the previous flow.

javierdelapuente (approved):

Thanhphan1147 (approved):

PR #670, opened 2025-01-31
Prevent updating matrix-auth data more than once

amandahla:
### Overview

This PR changes the way that matrix-auth library updates relation data by preventing to do it more than once. This way, relation-change events are not emitted for no reason and the encryption key secret is not changed.

### Rationale

Prevent changes in relation data.

### Checklist

### DIFF
--- a/lib/charms/synapse/v1/matrix_auth.py
+++ b/lib/charms/synapse/v1/matrix_auth.py
@@ -452,11 +452,9 @@ class MatrixAuthProvides(ops.Object):
                 updated.
         """
         try:
-            logger.warning(relation)
             MatrixAuthProviderData.from_relation(self.model, relation=relation)
             logger.warning("Matrix Provider relation data is already set, skipping")
-        except ValueError as e:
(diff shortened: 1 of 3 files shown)


javierdelapuente:
Out of curiosity, why was it emitted several times?

amandahla:
> Out of curiosity, why was it emitted several times?

The update_matrix_auth_integration is called during the reconcile [here](https://github.com/canonical/synapse-operator/blob/e7645690619f9274d0d5a707866c8287b5b7e629/src/charm.py#L253).

There was a method in the charm for checking if homeserver or registration changed but by mistake was checking for the key "shared_secret" which doesn't exist so it was updating every time.

Since these values are not supposed to be changed during a Synapse instance lifetime I moved the check to the library. The relation data is only updated if there is an ValueError (relation empty or invalid) now.

javierdelapuente (approved):

nrobinaubertin (approved):

PR #757, opened 2025-04-08
ISD-3222 remove custom issuer metadata

Thanhphan1147:
In the past we added custom issuer metadata to synapse configuration to avoid problems when deploying synapse with a self-signed certificate when behind a https proxy and also resolving DNS. However some clients like element X iOS requires a more complete version of the issuer metadata and it's much easier to simply fetch the complete metadata rather than try to manually input everything.

If synapse is behind an HTTPS proxy  and is using a self-signed certificate then the CA would need to be added to the pod, as well as ensure that the hostname is properly resolved from within the pod.

### Checklist


amandahla:
This means Synapse pod should have access to the metadata endpoint, right?

amandahla (approved):

Thanhphan1147:
> This means Synapse pod should have access to the metadata endpoint, right?

Yes, which should always be the case since MAS is on the same pod as synapse. However as the authentication flow is "browser-based", when you have ingress with HTTPS in front of MAS (synapse.com for example), synapse does a "roundtrip" and tries to reach MAS using `synapse.com/auth/` which would go outside and back into the ingress before getting routed to the same pod, instead of doing a request to `localhost` directly which is why there's complications with DNS and routing in that case.

gregory-schiano (approved):

PR #779, opened 2025-04-29
ISD-2819 restart synapse and MAS if MAS config has changed

Thanhphan1147:
Check for drifts using the existing method for synapse config and restart both Synapse and MAS if the config has changed.

### Checklist

- [x] The PR is tagged with appropriate label (`urgent`, `trivial`, `senior-review-required`)


src/pebble.py
```diff
@@ -402,7 +430,7 @@ def reconcile(  # noqa: C901
             ignore_string_case=True,
         )
 
-        if config_has_changed:
+        if config_has_changed or mas_config_has_changed:
```
amandahla, line 436: Does Synapse need to be restarted when MAS is restarted, or only when the MAS configuration changes?

Thanhphan1147: Synapse should only be restarted when MAS configuration changes, so even if MAS is restarted manually we don't need to restart synapse.

src/pebble.py
```diff
@@ -219,6 +219,28 @@ def _get_synapse_config(container: ops.model.Container) -> dict:
...
+        dict: MAS configuration.
+
+    Raises:
+        PebbleServiceError: if something goes wrong while interacting with Pebble.
+    """
+    try:
+        config = container.pull(MAS_CONFIGURATION_PATH).read()
+        return yaml.safe_load(config)
+    except ops.pebble.PathError as exc:
+        # If the MAS config has not been created, we return an empty dict to trigger a replan
```
amandahla, line 238: When could this happen?

amandahla (approved):

src/pebble.py
Thanhphan1147: The 2 most likely cases are: 
When the charm first starts, or when the config is manually deleted for whatever reason

PR #870, opened 2025-06-30
Change main unit to always be the 0 (and more)

amandahla:
### Overview

Changes in this PR:
- Adjusted how the main unit starts by consistently using "synapse/0".
- Revised how the instance_map config is generated based on the planned units number.
- Updated Mjolnir configuration: it no longer observes the collect status event and will only block the charm if moderator rooms are not found. Users should disable Mjolnir, create the room, and then re-enable it.
- Set the Synapse version using a constant instead of retrieving it via API due to frequent failures and reliability issues. Change Renovate configuration to properly update this constant.
- Introduced new tests using ops.testing/scenario.
- Charm will now enter an error status instead of being blocked if Pebble fails to start services.
- Refactored code: moved signing key handlers and helpers to dedicated files.
- Remove checks from the Pebble health endpoint (the checks are still there but no longer reflects on the result of the /health endpoint used by Kubernetes)

### Rationale

Enhances scalability management in Synapse by addressing complexity.

### Juju Events Changes

- Collect status and leader events are not observed anymore. 

### Module Changes

- Created files for signing key and helpers.

### Checklist

- [X] The PR is tagged with appropriate label (`urgent`, `trivial`, `senior-review-required`)

This track doesn't have a changelog.


.github/workflows/integration_test.yaml
```diff
@@ -14,3 +14,5 @@ jobs:
       juju-channel: 3.4/stable
       channel: 1.28-strict/stable
       modules: '["test_charm", "test_nginx", "test_s3", "test_scaling", "test_matrix_auth"]'
+      tmate-debug: true
```
nrobinaubertin: to be removed after debugging

src/signing_key.py
```diff
@@ -0,0 +1,117 @@
...
+        logger.error(
+            "Failed to get signing key: no peer relation %s found",
+            synapse.SYNAPSE_PEER_RELATION_NAME,
+        )
+        return content
+    secret_id = peer_relation.data[charm.app].get("secret-signing-id")
+    if secret_id:
+        try:
+            secret = charm.model.get_secret(id=secret_id)
+            logging.debug(secret.get_content().get("secret-signing-key"))
```
srbouffard: Not sure of the contents of the secret but is keeping this debug line a potential risk if ever the prod application is placed in debug logging?

src/signing_key.py
```diff
@@ -0,0 +1,117 @@
...
+        return
+    signing_key = ""
+    with container.pull(signing_key_path(charm_state)) as f:
+        signing_key = f.read()
+        signing_key = signing_key.rstrip()
+    if signing_key == get_signing_key_secret_content(peer_relation, charm):
+        logger.info("Received signing key but there is no change, skipping")
+        return
+    if charm.unit.is_leader():
+        logger.debug("Adding signing key to secret: %s", signing_key)
```
srbouffard: Similar question to previous comment regarding the debug line

srbouffard (approved):
Minor comments, LGTM!

src/helpers.py
```diff
@@ -0,0 +1,165 @@
...
+        charm: charm instance.
+        unit_id: number as 0 in synapse/0.
+
+    Returns:
+        unit address as unit-0.synapse-endpoints.
+    """
+    return f"{charm.app.name}-{unit_id}.{charm.app.name}-endpoints"
+
+
+def get_unit_number(charm: ops.CharmBase) -> str:
```
nrobinaubertin: This function is only used once in the code. I know that it is a matter of preference but I wouldn't have written it. 
When appropriate, 
```
unit_number = charm.unit.name.split("/")[1]
```
does the job for me

src/signing_key.py
amandahla: Right now maybe not because to get access to logs, one should have access to environment meaning that could read the file as well but considering a Loki scenario, it might be good removing. Good catch

src/signing_key.py
amandahla: Same, removed :)

.github/workflows/integration_test.yaml
amandahla: Done!

src/helpers.py
amandahla: I agree. I just left it like this because I wanted to check with @dimaqq  if it would make sense to have this helper function in ops, since [other charms](https://github.com/search?q=int%28self.unit.name.split%28%22%2F%22%29%5B1%5D%29+org%3Acanonical&type=code) use it as well.

DeeKay3 (approved):
LGTM, thanks!

src/helpers.py
amandahla: Marking as resolved for now but feel free to re-open it.

nrobinaubertin (approved):

src/charm.py
```diff
@@ -35,8 +34,8 @@
 
 logger = logging.getLogger(__name__)
 
-MAIN_UNIT_ID = "main_unit_id"
 INGRESS_INTEGRATION_NAME = "ingress"
+SYNAPSE_VERSION = "1.133.0"
```
gregory-schiano: Add a comment to specify that renovate udates this constant

src/charm.py
```diff
@@ -58,6 +57,7 @@ def __init__(self, *args: typing.Any) -> None:
             args: class arguments.
         """
         super().__init__(*args)
+        self.unit.set_workload_version(SYNAPSE_VERSION)
```
gregory-schiano: This can raise a a `TypeError` I'd move it to the event hook method

src/charm.py
```diff
@@ -84,7 +84,6 @@ def __init__(self, *args: typing.Any) -> None:
         self._mjolnir = Mjolnir(self, token_service=self.token_service)
         self.framework.observe(self.on.config_changed, self._on_config_changed)
         self.framework.observe(self.on.upgrade_charm, self._on_config_changed)
-        self.framework.observe(self.on.leader_elected, self._on_leader_elected)
```
gregory-schiano: We can now hook to the reconcile method directly (our conventions has been relaxed)

src/charm.py
```diff
@@ -305,180 +142,70 @@ def _on_relation_departed(self, event: RelationDepartedEvent, charm_state: Charm
...
-                return secret.get_content().get("secret-signing-key")
-            except (ops.model.SecretNotFoundError, ValueError, TypeError) as exc:
-                logger.exception("Failed to get secret id %s: %s", secret_id, str(exc))
-                del peer_relation[0].data[self.app]["secret-signing-id"]
-        return None
+    def _set_unit_status(self) -> None:
+        """Set unit status depending on Synapse and NGINX state."""
+        if isinstance(self.unit.status, ops.BlockedStatus):
+            # Preserve error state set elsewhere
+            return
```
gregory-schiano: This is added because as a side effect some methods in reconcile will update the status. As a good practice the status should only be set from the `charm.py` file.
As far as I see it's the `mjolnir.enable` method that could update the status. I'd refactor this method to throw business exceptions, catch them from the reconcile method and set the status accordingly. This will avoid this guard.

src/charm.py
```diff
@@ -305,180 +142,70 @@ def _on_relation_departed(self, event: RelationDepartedEvent, charm_state: Charm
...
-        if secret_id:
-            try:
-                secret = self.model.get_secret(id=secret_id)
-                logging.debug(secret.get_content().get("secret-signing-key"))
-                return secret.get_content().get("secret-signing-key")
-            except (ops.model.SecretNotFoundError, ValueError, TypeError) as exc:
-                logger.exception("Failed to get secret id %s: %s", secret_id, str(exc))
-                del peer_relation[0].data[self.app]["secret-signing-id"]
-        return None
+    def _set_unit_status(self) -> None:
```
gregory-schiano: I think that this method is not necessary anymore, `reconcile` is pretty short and already setting some status. This method is only called from reconcile, and contains some overlapping code (`can_connect` for example)

src/charm.py
```diff
@@ -305,180 +142,70 @@ def _on_relation_departed(self, event: RelationDepartedEvent, charm_state: Charm
...
 
-        This event handler will reconcile Synapse configuration after the following
-        scenarios:
-        - When the charm is deployed so the leader will be the main unit.
-        - When the leader, for any reason, has changed so the leader unit will be the main.
-        Once the peer data (main_unit_id) is changed, other units will emit reconcile and be
-        properly configured.
+        if not self._is_service_running(container, synapse.SYNAPSE_SERVICE_NAME):
+            self.unit.status = ops.MaintenanceStatus("Waiting for Synapse")
+            return
```
gregory-schiano: We rely on having a next event, but if there's no next event, we could stay in `Maintenance` for ever.
Usually what we do is calling synchronously a health endpoint of the service (or the one of pebble checks) repeating it with some backoff. In case of failure, error out (as the service can't start).

src/charm.py
```diff
@@ -305,180 +142,70 @@ def _on_relation_departed(self, event: RelationDepartedEvent, charm_state: Charm
...
-
-        if signing_key == self.get_signing_key():
-            logger.info("Received signing key but there is no change, skipping")
+        if helpers.is_redis_required(self, charm_state):
+            self.unit.status = ops.BlockedStatus("Redis integration is required.")
             return
+        self.model.unit.status = ops.MaintenanceStatus("Configuring Synapse")
+        helpers.configure_and_start_services(self, charm_state, container)
+        if helpers.is_mjolnir_enabled(self, charm_state):
+            self._mjolnir.enable(charm_state)
```
gregory-schiano: Shouldn't we check if mjolnir is started (similarly to the nginx / synapse check) ?

src/helpers.py
```diff
@@ -0,0 +1,165 @@
...
+        charm: charm instance.
+        charm_state: charm state.
+
+    Return:
+        True if is main and config is enabled.
+    """
+    return is_main(charm) and charm_state.synapse_config.enable_mjolnir
+
+
+def planned_units(charm: ops.CharmBase) -> int:
```
gregory-schiano: I think we could remove that on-liner as it's only used 3 times

src/helpers.py
```diff
@@ -0,0 +1,165 @@
...
+import signing_key
+import synapse
+from charm_state import CharmState
+
+logger = logging.getLogger(__name__)
+
+MAIN_UNIT_ID = 0
+
+
+def is_redis_required(charm: ops.CharmBase, charm_state: CharmState) -> bool:
```
gregory-schiano: Can be part of the charm state and computed when we create it

src/helpers.py
```diff
@@ -0,0 +1,165 @@
...
+        charm: charm instance.
+        charm_state: charm state.
+
+    Return:
+        True if more than 1 unit is found.
+    """
+    return charm_state.redis_config is None and planned_units(charm) > 1
+
+
+def is_mjolnir_enabled(charm: ops.CharmBase, charm_state: CharmState) -> bool:
```
gregory-schiano: This one too could be in the charm state and computed when it's created

src/helpers.py
```diff
@@ -0,0 +1,165 @@
...
+    Args:
+        charm: charm instance.
+
+    Returns:
+        Number of planned units.
+    """
+    return charm.app.planned_units()
+
+
+def peer_relation(charm: ops.CharmBase) -> typing.Optional[ops.Relation]:
```
gregory-schiano: Could be a property of the charm

src/helpers.py
```diff
@@ -0,0 +1,165 @@
...
+    Returns:
+        Synapse peer relation.
+    """
+    peer_relations = charm.model.relations[synapse.SYNAPSE_PEER_RELATION_NAME]
+    if not peer_relations:
+        return None
+    return peer_relations[0]
+
+
+def is_main(charm: ops.CharmBase) -> bool:
```
gregory-schiano: Could be a property of the charm

src/helpers.py
```diff
@@ -0,0 +1,165 @@
...
+    Args:
+        charm: charm instance.
+
+    Returns:
+        bool: true if is the main unit.
+    """
+    return f"/{MAIN_UNIT_ID}" in charm.unit.name
+
+
+def get_unit_address(charm: ops.CharmBase, unit_id: int) -> str:
```
gregory-schiano: Set it as a private method for this module

src/helpers.py
gregory-schiano: Only used once, could be inlined

src/signing_key.py
```diff
@@ -0,0 +1,115 @@
...
+def signing_key_path(charm_state: CharmState) -> str:
+    """Get signing key path.
+
+    Args:
+        charm_state: charm state.
+
+    Returns:
+        signing key path as string.
+    """
+    return f"/data/{charm_state.synapse_config.server_name}.signing.key"
```
gregory-schiano: I feel like `/data` should be a constant

src/charm.py
```diff
@@ -305,180 +142,70 @@ def _on_relation_departed(self, event: RelationDepartedEvent, charm_state: Charm
...
         Args:
-            signing_key: signing key as string.
+            charm_state: Instance of CharmState
         """
-        peer_relation = self.model.relations[synapse.SYNAPSE_PEER_RELATION_NAME]
-        if not peer_relation:
-            logger.error(
-                "Failed to set signing key: no peer relation %s found",
-                synapse.SYNAPSE_PEER_RELATION_NAME,
-            )
```
gregory-schiano, line 400: I'd keep this check and block the charm is there's no peer relation, as it's part of the pre-check of the charm from the reconcile perspective
The remaining part of the code will then be able to assume the relation exists as this is guarded early on

src/signing_key.py
```diff
@@ -0,0 +1,115 @@
...
+
+    Args:
+        peer_relation: synapse peer relation.
+        charm: charm instance.
+
+    Returns:
+        Content as string.
+    """
+    content = None
+    if not peer_relation:
```
gregory-schiano: If we check for the relation in reconcile, we can keep the check here to be defensive, but since we can assume the relation is there, we therefore can throw an exception (and remove the optional for the relation)

src/signing_key.py
```diff
@@ -0,0 +1,115 @@
...
+) -> None:
+    """Create secret with signing key content.
+
+    Args:
+        peer_relation: synapse peer relation.
+        charm: charm instance.
+        charm_state: charm state.
+        container: container.
+    """
+    if not peer_relation:
```
gregory-schiano: Either done here and in `write_to_container` or done in `get_signing_key_secret_content` but not necessary for both

src/signing_key.py
```diff
@@ -0,0 +1,115 @@
...
+        return
+    signing_key = ""
+    with container.pull(signing_key_path(charm_state)) as f:
+        signing_key = f.read()
+        signing_key = signing_key.rstrip()
+    if signing_key == get_signing_key_secret_content(peer_relation, charm):
+        logger.info("Received signing key but there is no change, skipping")
+        return
+    if charm.unit.is_leader():
+        secret = charm.app.add_secret({"secret-signing-key": signing_key})
```
gregory-schiano: What happens if the secret already exists but is different from the one on disk ?

gregory-schiano: To make the comment more explicit, this will create a new secret even if the secret already existed. We should update the previous with the key on disk (which is the source of truth)

src/signing_key.py
```diff
@@ -0,0 +1,115 @@
...
+        )
+        return
+    signing_key = ""
+    with container.pull(signing_key_path(charm_state)) as f:
+        signing_key = f.read()
+        signing_key = signing_key.rstrip()
+    if signing_key == get_signing_key_secret_content(peer_relation, charm):
+        logger.info("Received signing key but there is no change, skipping")
+        return
+    if charm.unit.is_leader():
```
gregory-schiano: Thie can be done at the beginning of the method with: 
```python
if not charm.unit.is_leader():
   return
```

src/charm.py
gregory-schiano: After further code review, the peer relation is used to store a secret_id, we now can use labels, so the peer relation might not be necessary anymore.
If that's removed beware of the upgrade path (coming from existing peer relation to non existing relation situation)

gregory-schiano: For example we could do a 2-step upgrade:
- Add the label (not removing the peer relation)
- Remove the peer relation (and secret can be retrieved using the label)

src/charm.py
```diff
@@ -113,372 +126,145 @@ def build_charm_state(self) -> CharmState:
...
-        self.set_main_unit(self.unit.name)
-        logger.debug("_on_leader_elected emitting reconcile")
-        self.reconcile(charm_state)
-
-    @inject_charm_state
-    def _on_relation_changed(self, _: ops.HookEvent, charm_state: CharmState) -> None:
-        """Handle Synapse peer relation changed event.
+        script = textwrap.dedent(
+            f"""\
+            while :; do
```
dimaqq: super-minor: while this is totally valid, it could be more readable for mere mortals as:


```suggestion
            while true; do
```

src/charm.py
```diff
@@ -113,372 +126,145 @@ def build_charm_state(self) -> CharmState:
...
-        peer_relation = self.model.relations[synapse.SYNAPSE_PEER_RELATION_NAME]
-        if not peer_relation:
-            logger.error(
-                "Failed to set signing key: no peer relation %s found",
-                synapse.SYNAPSE_PEER_RELATION_NAME,
-            )
+        if not self._is_service_running(container, synapse.SYNAPSE_SERVICE_NAME):
+            self.unit.status = ops.MaintenanceStatus("Waiting for Synapse")
+            self._install_callback_script(synapse.SYNAPSE_SERVICE_NAME)
+            container.start(f"{synapse.SYNAPSE_SERVICE_NAME}-charm-callback")
```
dimaqq: **question:** what is the rationale for this change?

`container.start()` kicks off the service and waits for it to start synchronously. 

Why is that not enough, or what does this PR do on top of that?

src/charm.py
amandahla: As we discussed, in this PR I added the label to the secret so the next step will be removing the peer relation

src/signing_key.py
amandahla: Moved to the configure_and_start_services

amandahla:
I'll disable style checker for now since changes in the documentation is out of the scope of this PR.

@erinecon FYI
Reference:
https://github.com/canonical/operator-workflows/pull/740

src/charm.py
amandahla: Excellent catch, thanks for raising this.
You made me realize that this isn’t needed.
The main reason for this change was to ensure the correct status is set when the service isn’t ready. In production, Synapse takes some time to start (probably longer than the checks) which could lead to Kubernetes restarting the container while the charm is also restarting after a config-changed event (for example).
So, I removed the entire collect-status observer and also removed the Pebble check names (as mentioned in the last AMA).

src/charm.py
```diff
@@ -112,101 +124,168 @@ def build_charm_state(self) -> CharmState:
...
-        if isinstance(self.unit.status, ops.BlockedStatus):
-            # Preserve error state set elsewhere
-            return
+        if charm_state.mjolnir_enabled:
+            try:
+                self._mjolnir.enable(charm_state)
+            except MjolnirEnableError as e:
+                self.model.unit.status = ops.MaintenanceStatus(str(e))
+            except MjolnirModeratorsNotFoundError as e:
+                self.unit.status = ops.BlockedStatus(str(e))
```
gregory-schiano, line 201: I like how it looks like now 👍

src/charm.py
```diff
@@ -112,101 +124,168 @@ def build_charm_state(self) -> CharmState:
...
-            charm_state: The charm state.
+            event: event triggers reconcile.
+            charm_state: charm state.
         """
+        if isinstance(event, ops.RelationDepartedEvent) and event.departing_unit == self.unit:
+            return
         self.reconcile(charm_state)
 
-    def reconcile(self, charm_state: CharmState) -> None:
+    # Too complex for now.
```
gregory-schiano: is it ?

src/charm.py
```diff
@@ -112,101 +124,168 @@ def build_charm_state(self) -> CharmState:
...
-        if not self._is_service_running(container, synapse.SYNAPSE_NGINX_SERVICE_NAME):
-            self.unit.status = ops.MaintenanceStatus("Waiting for NGINX")
-            return
+        try:
+            service = container.get_service(synapse.MJOLNIR_SERVICE_NAME)
+            if service and not service.is_running():
+                self.unit.status = ops.MaintenanceStatus("Waiting for Mjolnir")
+        except ops.ModelError:
+            # ModelError is raised if service not found
+            logger.info("mjolnir not found, skipping")
```
gregory-schiano, line 303: We have the info of whether mjolnir is expected or not from charm_state shouldn't we use that instead of using an exception as execution flow ?

src/database_observer.py
gregory-schiano, line 1: For later but I'm wondering if we couldn't remove the observer and include the hoook method logic into the reconcile method.
This will probably require the `db_client.prepare()` method to be idempotent but will move towards pure holistic aproach

src/media_observer.py
gregory-schiano, line 1: Same comment as the one of the DB observer, I'm wondering how much of this could be included into the reconcile to go to pure holistic

src/signing_key.py
```diff
@@ -73,18 +72,20 @@ def write_to_container(
...
+    secret = get_signing_key_secret(peer_relation, charm)
+    if secret:
+        content = secret.get_content().get(SIGNING_KEY_SECRET_CONTENT_ID)
+        if content:
+            container.push(
+                signing_key_path(charm_state),
+                content,
+                make_dirs=True,
+                encoding="utf-8",
+            )
```
gregory-schiano: I'd rather have a more defensive approach:
```
if not secret:
   ...
content = secret.get_content()...
if not content:
   ...
container.push(...)
```

src/signing_key.py
```diff
@@ -97,19 +98,35 @@ def write_to_secret(
...
+    if existing_secret:
+        # the existing secret already has label but has different content
+        # so we must only update it
+        try:
+            existing_secret.set_content(new_content)
+            existing_secret.set_info(label=SIGNING_KEY_SECRET_LABEL)
+            logger.info("Signing key secret was updated, id: %d", existing_secret.id)
+            return
+        except (ops.model.SecretNotFoundError, ValueError, TypeError) as exc:
+            logger.exception("Failed to get secret id %s: %s", existing_secret, str(exc))
```
gregory-schiano, line 172: Should we keep proceeding in such conditions. It looks like if we reach that state there's something going wrong on Juju side which could have broader impact. I'd perhaps let the exception flow in such case.

src/charm.py
amandahla: Since the charm is getting the service status, I considered using the ops.ModelError exception for simply logging if mjolnir is not found. This would cover not-leader units and Mjolnir not enabled. WDYT?

src/database_observer.py
amandahla: I agree, this would be a nice refactor.

amandahla: Created: ISD-4095

src/media_observer.py
amandahla: Created: ISD-4095

src/charm.py
amandahla: not anymore :)

src/signing_key.py
amandahla: For this change, I added an Exception in case there is no secret or no content for writing to the container.
During the reconcile, this will cause only a MaintenanceStatus change.
During the check for setting unit status, I added a method for check if the secret content and the file in the container are the same. If not, will change to MaintenanceStatus too.

src/signing_key.py
amandahla: In that case, a new secret is created. If add secret fails, then the charm will enter error status.

amandahla:
Lint fails due to https://github.com/canonical/operator-workflows/pull/790

gregory-schiano (approved):

src/charm.py
gregory-schiano: I still think that such kind of information should come from the CharmState instead of relying on a side effect. But it's not a blocking issue so I approved the PR

src/signing_key.py
gregory-schiano: Ok, so if there's anything wrong we'll endup in error state at the time to trying to "recreate" the secret, fine.

PR #927, opened 2025-08-07
Add improvements to script upgrade_room

amandahla:
### Overview

This PR adds improvements to upgrade_room
- Timer
- Check version before upgrading
- Dry run shows current version
- Upgrade only room in homeserver

### Checklist

### DIFF
--- a/scripts/upgrade_rooms.py
+++ b/scripts/upgrade_rooms.py
@@ -328,9 +328,7 @@ def main() -> None:  # noqa: C901
         logger.info("%s - upgrading using room_id", args.room_id)
         current_room_version = get_room_version(admin_access_token, server_url, args.room_id)
         if current_room_version == version:
-            logger.warning(
-                "%s - room already is version %d, no action", args.room_id, version
-            )
+            logger.warning("%s - room already is version %d, no action", args.room_id, version)
             return
         upgrade_room(admin_access_token, server_url, args.room_id, version)
         return
--- a/scripts/upgrade_rooms.py
+++ b/scripts/upgrade_rooms.py
@@ -357,7 +357,7 @@ def main() -> None:  # noqa: C901
             while True:
                 user_input = (
                     input(
-                        f"{room_id} - failed, action required. What do you want to do? [r]etry / [s]kip / [f]inish: "  # noqa: E501
+                        f"{room_id} - failed, action required. What do you want to do? [r]etry / [s]kip / [f]inish: "  # noqa: E501 W505
                     )
                     .strip()
                     .lower()
--- a/scripts/upgrade_rooms.py
+++ b/scripts/upgrade_rooms.py
@@ -340,8 +340,8 @@ def main() -> None:  # noqa: C901
         room_id = room.get("room_id", "")
         try:
             current_room_version = get_room_version(admin_access_token, server_url, room_id)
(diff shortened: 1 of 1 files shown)


scripts/upgrade_rooms.py
```diff
@@ -282,21 +309,76 @@ def main() -> None:
...
+            except SynapseServerError:
+                error = "Failed"
+            print(f"{name};{room_id};{num_joined};{room_version};{error}")
         return
 
     if args.room_id:
-        logger.info("room-id is set, upgrading %s", args.room_id)
+        logger.info("%s - upgrading using room_id", args.room_id)
+        current_room_version = get_room_version(admin_access_token, server_url, args.room_id)
+        if current_room_version == args.version:
```
DeeKay3: Minor: `version = args.version` is defined in line 302.
```suggestion
        if current_room_version == version:
```

scripts/upgrade_rooms.py
```diff
@@ -282,21 +309,76 @@ def main() -> None:
...
+            logger.warning(
+                "%s - room already is version %d, no action", args.room_id, args.version
```
DeeKay3: ditto
```suggestion
                "%s - room already is version %d, no action", args.room_id, version
```

scripts/upgrade_rooms.py
```diff
@@ -282,21 +309,76 @@ def main() -> None:
...
         return
 
+    logger.info("all public rooms will be upgraded")
     for room in public_rooms:
         room_id = room.get("room_id", "")
-        upgrade_room(admin_access_token, server_url, args.room_id, version)
+        try:
+            current_room_version = get_room_version(admin_access_token, server_url, room_id)
+            if current_room_version == args.version:
+                logger.warning("%s - room already is version %s, no action", room_id, args.version)
```
DeeKay3: ditto
```suggestion
            if current_room_version == version:
                logger.warning("%s - room already is version %s, no action", room_id, version)
```

DeeKay3 (approved):
lgtm, minor proposal

scripts/upgrade_rooms.py
```diff
@@ -126,7 +129,7 @@ def get_current_user(admin_access_token: str, server_url: str) -> str:
 
 
 def get_public_rooms(admin_access_token: str, server_url: str, limit: int = 30) -> list:
-    """Get public rooms.
+    """Get the public rooms list, ordered from smallest to largest.
```
srbouffard, line 132: Given the `publicRooms` API returns rooms from biggest to smallest, when the API call hits the room limit, then some of the remaining smaller rooms will be truncated right? Do we expect to hit the room limit at times?  

If so, then it might be a good idea to pull in the full paginated list before flipping it to smallest -> largest so make sure the output can consistently return the smallest rooms.

amandahla: Yes, I'm not exactly proud of this implementation since considers that the operator will increase the limit to get all the public rooms. There is no logic with pagination + get rooms from smallest to largest.

srbouffard: Ok, I think that we can leave it as is given we are the ones to run the script and we know what to look out for

srbouffard (approved):

PR #932, opened 2025-08-07
[DNR DNM] CI for juju 3.6, ck8s 1.34-classic, track-1

nrobinaubertin:
### Checklist


amandahla:
Considering what we discussed about using latest versions for CI and also considering that we should use k8s instead of microk8s, could you update this PR and the one for 2/main branch with:
```
with:
      channel: 1.34-classic/stable
      juju-channel: 3/stable
      self-hosted-runner: true
      self-hosted-runner-label: "xlarge"
      use-canonical-k8s: true
      provider: 'k8s'
```

So we can make it everything right: use k8s, latest versions and also use self-hosted-runners.

wdyt

nrobinaubertin:
> So we can make it everything right: use k8s, latest versions and also use self-hosted-runners.

I made the change but decided to pin the version of juju to the minor: I feel like this is more correct if we want to test for a particular version of it (rather than just following the major)

PR #992, opened 2025-09-10
Fix rock conflict resolution

amandahla:
### Checklist

### DIFF
--- a/synapse_rock/rockcraft.yaml
+++ b/synapse_rock/rockcraft.yaml
@@ -164,8 +164,14 @@ parts:
             chmod 755 $CRAFT_PART_INSTALL/usr/local/attributemaps
             rm -rf $CRAFT_PART_INSTALL/aws/dist/awscli/examples/
         overlay-packages:
-            - ca-certificates
             - libjemalloc2
+    add-ca-certificates:
+      plugin: nil
+      build-packages:
+        - ca-certificates
+      override-build: |-
+        mkdir -p $CRAFT_PART_INSTALL/etc/ssl/certs/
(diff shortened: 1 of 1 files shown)


synapse_rock/rockcraft.yaml
```diff
@@ -164,8 +164,14 @@ parts:
...
         overlay-packages:
-            - ca-certificates
             - libjemalloc2
+    add-ca-certificates:
+      plugin: nil
+      build-packages:
+        - ca-certificates
+      override-build: |-
+        mkdir -p $CRAFT_PART_INSTALL/etc/ssl/certs/
+        cp /etc/ssl/certs/ca-certificates.crt $CRAFT_PART_INSTALL/etc/ssl/certs/ca-certificates.crt
```
beliaev-maksim, line 174: can you please explain how this has fixed the issue with path error in `usr/lib` ?

```
Part 'superset' and the overlay of part 'overlay-pkgs' list the following files, but with different contents or permissions:
    usr/lib/ssl/cert.pem
    usr/lib/ssl/certs
    usr/lib/ssl/openssl.cnf
    usr/lib/ssl/private
```

looks very counter intuitive

amandahla: Since I only needed the ca-certificates.crt, I could remove the ca-certificates from the overlay step, adding it to a new part and installed it only in the build step. So no conflicts anymore and I have the file that I need in the rock.
Another way would be adding to stage list in synapse part as Tiago suggested. Same as the solution that was applied to the Discourse rock.

beliaev-maksim: thanks

PR #1061, opened 2025-11-21
Run Allure workflow only on schedule

swetha1654:
Run Allure workflow only on schedule

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -41,7 +41,7 @@ jobs:
       channel: 1.28-strict/stable
       modules: '["test_charm", "test_nginx", "test_s3", "test_scaling", "test_matrix_auth"]'
   allure-report:
-    if: ${{ (success() || failure()) && github.event_name == 'schedule' }}
+    if: ${{ !cancelled() && github.event_name == 'schedule' }}
     needs:
     - integration-tests
     uses: canonical/operator-workflows/.github/workflows/allure_report.yaml@main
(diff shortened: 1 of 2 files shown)


DeeKay3 (approved):

.github/workflows/integration_test.yaml
```diff
@@ -2,43 +2,46 @@ name: Integration tests
...
-            if [[ "$file" != terraform/* ]]; then
-              echo "run_tests=true" >> "$GITHUB_OUTPUT"
-              break
-            fi
-          done
+    - name: Set output if non-terraform files changed
+      id: set-output
+      run: |
+        echo "run_tests=false" >> "$GITHUB_OUTPUT"
+        for file in ${{ steps.changed-files.outputs.all_changed_files }}; do
```
amandahla, line 23: Does the schedule trigger provide "steps.changed-files.outputs.all_changed_files" ?

.github/workflows/scheduled_integration_test.yaml
```diff
@@ -13,7 +13,7 @@ jobs:
       extra-arguments: -x --localstack-address 172.17.0.1
       pre-run-script: localstack-installation.sh
       trivy-image-config: "trivy.yaml"
-      juju-channel: 3.4/stable
+      juju-channel: 3.6/stable
```
amandahla, line 16: I splitted the workflow in 2 to prevent running integration tests if the PR only changes the terraform but now I see that this could be done in only 1 workflow by adding a "if" to the "check-changes". wdyt

swetha1654:
@amandahla closing this PR since there is already another workflow file running scheduled integation tests and updating the Allure reports

PR #1098, opened 2025-12-16
Refactor integration tests

amandahla:
#### What this PR does

This PR refactors the integration test suite to improve performance, maintainability, and organization:

- Creates centralized dependency management with a new `dependencies.py` module 
- Optimizes test performance by converting fixtures to constants and reducing timeout values

Notes:
- Why repeated fixtures for deploying Synapse? Isolating and speeding up module tests.
- Why keep the test_matrix_auth.py? If it is decided to use IRC Bridge, we have a test ready to be used.

Next steps:
- Set Renovate to update the revisions
- Run the entire S3 module only if there is a change in Synapse workload, boto version, or S3 charm

#### Why we need it

**Performance Issues:**
- Integration tests were slow due to inefficient fixture scoping and excessive timeouts
- S3-related fixtures were loaded for all tests even when not needed
- Hardcoded charm versions made dependency management error-prone

**Maintainability Problems:**
- No single source of truth for charm versions and channels
- Renovate couldn't automatically track and update charm dependencies  
- Test infrastructure was scattered across multiple files without clear organization

#### Review focus

- Review the `CharmDependency` dataclass design and integration with existing fixtures
- Ensure S3-specific fixtures in `test_s3.py` are properly isolated and don't affect other tests
- Integration tests removed are not needed.

#### Checklist

- [x] I followed the [contributing guide](https://github.com/canonical/is-charms-contributing-guide)
- [ ] I added or updated the documentation (if applicable)
- [ ] I updated `docs/changelog.md` with user-relevant changes
- [ ] I added a [change artifact](../docs/release-notes/template/docs/release-notes/template/_change-artifact-template.yaml) for user-relevant changes in `docs/release-notes/artifacts`. If no change artifact is necessary, I tagged the PR with the label `no-release-note`.
- [x] I added or updated tests as needed (unit and integration)
- [x] **If integration test modules are used:** I updated the workflow configuration (e.g., in `.github/workflows/integration_tests.yaml`, ensure the `modules` list is correct)
- [ ] **If this is a Grafana dashboard:** I added a screenshot of the dashboard
- [ ] **If this is Terraform:** `terraform fmt` passes and `tflint` reports no errors  
- [ ] **If this is Rockcraft:** I updated the version


tests/integration/conftest.py
```diff
@@ -199,64 +106,121 @@ async def get_unit_ips(application_name: str):
...
+    synapse_image: str,
+    model: Model,
+    synapse_charm: str,
+    postgresql_app: Application,
+    pytestconfig: Config,
+    get_unit_ips: typing.Callable[[str], typing.Awaitable[tuple[str, ...]]],
+):
+    """Build and deploy the Synapse charm."""
+    use_existing = pytestconfig.getoption("--use-existing", default=False)
+    if not use_existing and SYNAPSE_APP_NAME not in model.applications:
```
nrobinaubertin, line 121: `SYNAPSE_APP_NAME not in model.applications`: I remember wanting to do that when I started writing integ tests for discourse 2.5 years ago and the team response was that we should assume a clean state for the tests.  
`use_existing` is a QoL thing when testing locally but the latter check, is it used ?

tests/integration/test_charm.py
```diff
@@ -18,14 +17,20 @@
 from ops.model import ActiveStatus
 
 import synapse
-from auth.mas import MAS_CONFIGURATION_PATH, MAS_EXECUTABLE_PATH
+
+from .conftest import DUMP_MAS_CONFIG
+
+# Application name constants
+NGINX_INTEGRATOR_APP_NAME = "nginx-ingress-integrator"
+SYNAPSE_APP_NAME = "synapse"
```
nrobinaubertin: Why are those constants redefined here ?

tests/integration/test_s3.py
```diff
@@ -4,72 +4,369 @@
...
 
 # caused by pytest fixtures
-# pylint: disable=too-many-arguments, duplicate-code
+# pylint: disable=too-many-arguments, duplicate-code, unused-argument
 
 # mypy has trouble to inferred types for variables that are initialized in subclasses.
 ACTIVE_STATUS_NAME = typing.cast(str, ActiveStatus.name)  # type: ignore
+S3_MEDIA_INTEGRATOR_APP_NAME = "s3-integrator-media"
+S3_BACKUP_INTEGRATOR_APP_NAME = "s3-integrator-backup"
+SYNAPSE_S3_APP_NAME = "synapse-s3"
```
nrobinaubertin, line 36: It would maybe make sense to have a `tests/integration/constants.py` module at this point

srbouffard (approved):

tests/integration/test_charm.py
amandahla: Copilot's fault ;-)
I fixed now, good catch.

tests/integration/test_s3.py
amandahla: These are only used in this file so I prefer to keep it here.

tests/integration/conftest.py
amandahla: This fixture is used by multiple tests in different modules so this check only make sure that Synapse wont be re-deployed in case is already available.

tests/integration/conftest.py
```diff
@@ -198,64 +105,121 @@ async def get_unit_ips(application_name: str):
...
+        async with ops_test.fast_forward():
+            app = await model.deploy(
+                f"./{synapse_charm}",
+                resources=resources,
+                application_name=SYNAPSE_APP_NAME,
+                config={"server_name": SERVER_NAME},
+            )
+            await model.wait_for_idle(
+                apps=[SYNAPSE_APP_NAME],
+                status=typing.cast(str, BlockedStatus.name),
```
DeeKay3, line 134: Isn't `BlockedStatus.name` is a constant str already? Why did we need to cast?

In any case, when its string, I prefer to do just `str(BlockedStatus.name)`, seems more pythonic to me. Very small nitpick

tests/integration/conftest.py
```diff
@@ -198,64 +105,121 @@ async def get_unit_ips(application_name: str):
...
-
-
-@pytest.fixture(scope="module", name="postgresql_app_name")
-def postgresql_app_name_app_name_fixture() -> str:
-    """Return the name of the postgresql application deployed for tests."""
-    return "postgresql-k8s"
+        await model.add_relation(f"{app.name}:nginx-route", f"{synapse_app.name}:nginx-route")
+        await model.wait_for_idle(status=ACTIVE_STATUS_NAME, idle_period=10)
+    yield app
+    await model.remove_application(app.name)
```
DeeKay3: I like to wrap cleanup around "try-finally" block s.t.:

```
try:
    # do stuff
    yield app
finally:
    if app.name in model.applications:
        await model.remove_application(app.name)
```
So that if an error comes up, clean up still runs.

tests/integration/conftest.py
```diff
@@ -198,64 +105,121 @@ async def get_unit_ips(application_name: str):
...
+    async with ops_test.fast_forward():
+        app = await model.deploy(
+            OAUTH_EXTERNAL_IDP_INTEGRATOR.charm_name,
+            application_name=OIDC_APP_NAME,
+            channel=OAUTH_EXTERNAL_IDP_INTEGRATOR.channel,
+            revision=OAUTH_EXTERNAL_IDP_INTEGRATOR.revision,
+            config=mock_external_idp_config,
+        )
+        await model.wait_for_idle(apps=[app.name], idle_period=30)
+        await synapse_app.model.relate(synapse_app.name, app.name)
```
DeeKay3: Isn't just `model.relate(synapse_app.name, app.name)` sufficient?

DeeKay3 (approved):
small nitpicks

tests/integration/conftest.py
amandahla: Your comment makes sense. I didn’t mention this in the PR description, but a follow-up step for this work is converting the tests to Jubilant. With that in mind, I’d prefer to keep this PR focused on what the integration tests should cover and avoid expanding the scope for now.

tests/integration/conftest.py
amandahla: Interesting suggestion. I changed it now :)

tests/integration/conftest.py
amandahla: I dont know why only this line is set like that. I removed, let's see if there is no difference in the behavior.

PR #1101, opened 2025-12-17
chore: migrate architecture docs

erinecon:
Moves architecture documentation to Reference folder to align with Diataxis.


erinecon:
Closing because there's several hundred commits that I wasn't expecting.

PR #1162, opened 2026-04-24
Replace localstack with MicroCeph 

srbouffard:
### 1. Replace localstack with MicroCeph radosgw for S3 tests

S3 integration tests now use MicroCeph radosgw as the S3-compatible backend, matching the fix applied to discourse operator.

- Added `s3-installation.sh` (sets up MicroCeph radosgw with fixed credentials)
- Removed `localstack-installation.sh`
- Updated `tests/integration/conftest.py`: replaced `localstack_address` fixture with `s3_address` (auto-detects host IP, port 7480, fixed credentials)
- Updated `tests/conftest.py`, `tox.ini`, and `.github/workflows/integration_test.yaml` to use the new script and env var names

### 2. Fix `psycopg2-binary` install hook crash

charmcraft 4.x runs two pip passes: first installs `charm-binary-python-packages` as binary wheels, then installs `requirements.txt` with `--no-binary=:all:`. Without version pins in `charm-binary-python-packages`, the second pass detects a version mismatch and recompiles from source — producing a `.so` that links against system `libpq.so.5`, which is absent in Juju's charm container.

Fix: pin versions in `charmcraft.yaml` to match `requirements.txt` so the binary wheel is already satisfied and pip skips recompilation.

- `charmcraft.yaml`: `psycopg2-binary==2.9.10` and `cryptography==46.0.4` in `charm-binary-python-packages`
- `requirements.txt`: downgraded `psycopg2-binary` from `2.9.11` → `2.9.10` (2.9.11 also lacked a compatible binary wheel in this build context)


nrobinaubertin (approved):

danielvnguyen (approved):
LGTM

Small nitpick, I noticed there is still mentions of LocalStack in some other files

- `CONTRIBUTING.md`: there is still a "S3 integration tests (Localstack)" section
- `pyproject.toml`: one line mentioning LocalStack in the markers sub-section of the pytest configuration settings

Seems like these changes are more appropriate for a documentation task though?

PR #1207, opened 2026-07-20
fix(ci): grant permissions for auto_update_libs reusable workflow

srbouffard:
## Summary
- add explicit permissions to `.github/workflows/auto_update_libs.yaml`
- grant `contents: write`, `pull-requests: write`, and `id-token: write` to the reusable-workflow caller job

## Why
The auto-update libraries workflow has a long startup-failure streak and fails before jobs start without explicit write scopes when default workflow permissions are read-only.

## Resolution
Add explicit caller permissions so scheduled runs can start and create update PRs successfully.


danielvnguyen (approved):

srbouffard:
Superseded by #1208. #1207 was opened from the default track/1 branch into main, which pulled unrelated commits. #1208 is rebased from main and contains only the workflow-permission fix.

PR #1208, opened 2026-07-20
fix(ci): grant permissions for auto_update_libs reusable workflow

srbouffard:
## Summary
- add explicit permissions to `.github/workflows/auto_update_libs.yaml`
- grant `contents: write`, `pull-requests: write`, and `id-token: write` to the reusable-workflow caller job

## Why
The auto-update workflow can fail at startup without explicit caller permissions when default workflow token permissions are read-only.


danielvnguyen (approved):

srbouffard:
Superseded by #1209, which correctly targets track/1 (the repository default branch).

